PR #14, opened 2026-04-10
feat(server,client): S01 scaffold with ping/pong round-trip

### LINKED ISSUE #1: S01 — Go server + client scaffold
**Phase:** 0 · **Goal:** stand up both services and prove one round-trip.

## Tasks
- [ ] `server/` Go module (`go mod init`) with lint (`golangci-lint`), format (`gofmt`), test (`go test`), `make dev` target
- [ ] `client/` TypeScript + Vite app with ESLint, Prettier, Vitest
- [ ] Pick WebSocket library (gorilla recommended) — decision in `docs/decisions/0001-ws-library.md`
- [ ] Pick client framework (React or Svelte) — decision in `docs/decisions/0002-client-framework.md`
- [ ] WebSocket server on :8080 accepting JSON frames
- [ ] Client sends a `ping` action frame; server broadcasts a state delta; client logs it
- [ ] Draft `docs/protocol.md` — v0 schema for action and state-delta frames

## Exit criteria
`make dev` in `server/` plus the client dev server produce a working round-trip, and the v0 protocol schema is written down.

See [docs/sprints.md](../blob/main/docs/sprints.md) and [AGENTS.md](../blob/main/AGENTS.md) §4 for commit/PR discipline.

krakenhavoc:
## Summary
- **Stands up both services** for the first time and proves the WebSocket seam works end-to-end. This is the S01 exit criteria: `make server-dev` + `npm run dev` in `client/` produces a working ping → pong round-trip, and protocol v0 is written down in `docs/protocol.md`.
- **Go server**: `server/` module with a `gorilla/websocket`-based hub following the canonical chat-example pattern. Endpoints: `GET /healthz`, `GET /ws`. Graceful shutdown via `sync.WaitGroup`. 81.9% statement coverage on `internal/ws` with `-race`.
- **TypeScript client**: `client/` is a Svelte 5 (runes) + Vite + TS app. `src/lib/ws.ts` wraps the WebSocket in a `GameClient` class exposing Svelte stores for reactive UI. `App.svelte` is a minimal demo UI for S01 only.
- **Decisions locked in**: ADR 0001 (`gorilla/websocket` — boring, battle-tested, canonical patterns for everything we'll need in S02/S03). ADR 0002 (`Svelte 5` — less ceremony, stores map cleanly onto WebSocket state streams; flagged tentative with clear revisit criteria).
- **Review fixes included**: a parallel review agent flagged two bugs before I opened this PR and I fixed them in the same commit — see "Review feedback incorporated" below.

## Sprint
S01 — Go server + client scaffold

## Issues
Closes #1

## Based on
`docs/agents-and-sprint-plan` (PR #13). This branch inherits the Option B pivot and sprint plan; once PR #13 merges this PR's base auto-updates to `main`.

## What's in protocol v0
- `ping` (client → server), `pong` (server → client), `error` (server → client)
- Frame envelope: `{v, kind, id, payload}`, versioned, with additive-fields-within-a-version rule
- Error codes: `bad_version`, `bad_json`, `bad_request`, `internal`
- Explicitly out of scope for v0: any game state, any auth, any game actions beyond ping. Those arrive in S02 (`game state: zones + turns`) and S03 (`action protocol + state deltas`).

## Server architecture notes
The hub follows the gorilla chat example pattern because it is the pattern Gorilla exists to enable:
- `Hub` owns `map[*Client]struct{}` and mediates register/unregister under a mutex.
- Each connection gets **two goroutines** tracked by a `sync.WaitGroup`: a `readPump` that reads frames and dispatches by `Kind`, and a `writePump` that drains a per-client `send chan []byte`. **The hub never writes to a socket directly** — everything goes through the client's send channel, which makes concurrent-write bugs (the classic gorilla footgun) impossible by construction.
- Pong timer + read deadline for heartbeat health. 64 KB max frame size at v0; will grow once game states do.
- `Shutdown(ctx)` sends a `CloseGoingAway` to every live client, then waits on the `WaitGroup` with a `select` against `ctx.Done()`. Zero clients → returns immediately.

## Client architecture notes
- Svelte stores (`writable<ConnectionStatus>`, `writable<LogEntry[]>`) are the single source of truth for UI state. Components subscribe; components never touch the WebSocket directly.
- The client protocol types in `src/lib/protocol.ts` are a hand-maintained mirror of `server/internal/protocol/protocol.go`. This is a deliberate trade-off vs. code generation: at the scale of v0 the types are trivial, and code generation has its own complexity cost. If the mirror gets painful we'll introduce `oapi-codegen` or similar in S03.
- Vite dev server proxies `/ws` to `ws://localhost:8080`, so the client uses `ws://location.host/ws` in both dev and production without special-casing.
- Svelte 5 runes (`$state`) throughout — no legacy reactive `$:` blocks.

## Review feedback incorporated
A review agent looked at this before I opened the PR and caught two real bugs:

1. **Client reconnect race** in `client/src/lib/ws.ts` — before reconnect, a previous socket's async close event could fire *after* `connect()` had assigned a new socket, then null out `this.socket`, leaving the client silently unable to send. **Fix:** every event listener now captures its socket in a closure and guards with `if (this.socket !== socket) return`. Stale events from previous sockets are cleanly ignored.
2. **`Hub.Shutdown` always blocked the full context deadline** even when no clients were connected, because of an unconditional `<-ctx.Done()`. Made the dev loop feel sluggish. **Fix:** each registered client's `readPump`/`writePump` is now tracked by a `sync.WaitGroup`. `Shutdown` closes every connection, then `select`s between `wg.Wait()` and `ctx.Done()`. Covered by `TestShutdownEmptyHubIsImmediate` (regression, asserts <200ms on an empty hub) and `TestShutdownWithClientsWaitsForPumps` (three clients, asserts <1s).

A third note was about `AGENTS.md` §3/§5 allegedly still saying "not yet created" / "TBD" — the review agent was looking at an older snapshot. Both sections were updated in this branch to reflect the real scaffolded layout. The only remaining "not yet created" marker is for `scripts/`, which is accurate (it arrives in S04).

## Test plan
- [ ] CI (once configured) runs `make -C server test vet` and `cd client && npm run check && npm run lint && npm run build`
- [ ] `make server-dev` starts the server on :8080
- [ ] `curl http://localhost:8080/healthz` returns `ok`
- [ ] `cd client && npm install && npm run dev` starts Vite on :5173
- [ ] Opening http://localhost:5173 shows the connected status pill within a second
- [ ] Clicking "send ping" produces a matching pong in the activity log
- [ ] Stopping the server with Ctrl-C returns the prompt in well under 5 seconds (regression guard)

## Out of scope for S01 (don't review these, they land later)
- No game state, zones, players, turns — S02
- No game action types beyond ping — S03
- No lobby, no auth, no deck import — S04–S05
- No play area (PixiJS) — S05–S06
- CI configuration — deliberately deferred; `make lint` is enough for S01

## Notes for reviewer
Worth checking:
- The `sync.WaitGroup` pattern in `hub.go` — did I get the `Add(2)` / `defer Done()` pairing right? Running `go test -race ./...` locally passes, but another set of eyes is welcome.
- The protocol mirror approach in `client/src/lib/protocol.ts` — are we comfortable with the manual-sync trade-off, or do we want codegen from S02?
- ADR 0002 (Svelte 5) is tentative; if you or Ian has a strong React preference, now is the moment. Flipping costs a day in S01, a week in S05.


krakenhavoc:
## Code review pass — fixes applied in `a1bcb8e`

Two parallel review agents (one on the Go server, one on the TS client + build config + protocol mirror) reviewed this branch top to bottom. I also ran the full end-to-end happy path against the real binary (not just the httptest harness): start the Go server on :8080, start the Vite dev server on :5173, dial `ws://localhost:5173/ws` with a real gorilla/websocket client, verify a v0 pong with a valid RFC3339 `server_time` comes back.

### Blockers found and fixed
1. **`server/go.mod` pinned `go 1.26`** — a module released 6 weeks before today would lock out collaborators on older Go installs. Nothing here actually uses a 1.22+ feature, so I dropped the directive to `go 1.22`.
2. **Shutdown ordering in `server/cmd/server/main.go`** — `hub.Shutdown` ran before `srv.Shutdown`, so a client connecting in the race window could leak its pumps or block `wg.Wait()` for the full 5s context timeout. Swapped: `srv.Shutdown` first (closes the listener), `hub.Shutdown` second. **Measured real SIGTERM exit time after the fix: 12 ms.**
3. **`@eslint/js` imported in `client/eslint.config.js` but not declared in `package.json`** — only resolving via transitive hoist today; any `--install-strategy=nested` or eslint internal reshuffle would break `npm run lint`. Added to `devDependencies`.
4. **`npm test` exit 1 on a fresh clone** — `vitest run` with no test files exits non-zero; `--passWithNoTests` added.

### Worth-fixing items also landed in the same commit
- **Hub `closed` flag + `register` rejection** so late WebSocket upgrades during shutdown return 503 instead of joining a tearing-down hub. New test `TestServeWSRejectsAfterShutdown`.
- **`sendFrame` now disconnects on full buffer** instead of silently dropping the frame — the S01 exit criteria literally depends on the pong being delivered.
- **Makefile default target no longer rewrites source**; new `fmt-check` uses `gofmt -s -l`. `fmt` remains as the explicit write target.
- **`.golangci.yml` trimmed**: removed `govet: enable-all: true` which would enable `fieldalignment` and spam the intentionally-ordered Hub struct.
- **`protocol_test.go`**: exact-value assertions for all Kind strings, error codes, and the version constant against docs/protocol.md.
- **Client `disconnect()`** now explicitly sets status to `"disconnected"` before nulling the socket (the captured-socket close guard was short-circuiting the status update).
- **New `"info"` log direction** for connect/close events; previously mis-labelled as `"sent"` (green) and `"error"` (red), so the UI read every reconnect as a failure.
- **`engines: { node: ">=20.11" }`** in `package.json` so npm warns rather than failing obscurely on older Node.
- **`vite.config.ts` proxy target**: `ws://localhost:8080` → `http://localhost:8080` (canonical Vite form; `ws: true` handles the upgrade on top).
- **`App.svelte` refactored** to idiomatic Svelte 5: `$effect` for mount + cleanup, `{$status}` / `{$log}` auto-subscribed stores instead of manual `.subscribe` + `$state` copies.
- **Docs synced**: AGENTS.md §3 client subtree and client/README.md layout now list every file a first-day collaborator will touch.

### Protocol mirror
The client reviewer verified `client/src/lib/protocol.ts` line-by-line against `server/internal/protocol/protocol.go`: every Kind string, every error code, every JSON tag, `omitempty` → optional. **No divergence.** This is the most important cross-service invariant for S01, and it holds.

### Not fixed (deferred intentionally)
- **Mount/unmount cleanup edge case**: `mount(App, target)` returns an `unmount` handle that `main.ts` never calls, so `$effect` cleanup never runs. Works in practice because tab teardown closes the socket anyway. Will matter if App becomes a child of a router in a later sprint. Noted, not urgent.
- **No client unit tests exist yet** — vitest is in devDeps and runs cleanly with `--passWithNoTests`. Tests arrive with real logic in S02.
- **CI configuration** — deliberately deferred per AGENTS.md §7. `make lint` is enough at S01.

### Verification re-run after fixes
- `make fmt-check vet test build` — all green, 79.8% ws coverage, race detector clean, 6 ws tests + 3 protocol tests.
- `npm run check lint build test` — all green.
- Direct ping against `ws://127.0.0.1:8080/ws` — OK.
- Proxied ping through `ws://127.0.0.1:5173/ws` (Vite → Go) — OK.
- SIGTERM exit time: 12 ms.

Ready for human review.

Issue #9, opened 2026-04-10
S09 — Polish I — animations + VFX


krakenhavoc:
**Phase:** 5 · **Goal:** make the table feel alive.

## Tasks
- [ ] GSAP integration
- [ ] Card draw / play / tap / untap animations
- [ ] Damage number popups, combat arrows
- [ ] Sound pass (draw, play, tap, damage, turn change)
- [ ] Particle effects on ETB / death

## Exit criteria
Side-by-side video vs. Cockatrice shows a clear "this feels better" delta.

Issue #7, opened 2026-04-10
S07 — Turn/phase UI + chat + manual priority


krakenhavoc:
**Phase:** 3 · **Goal:** the fiddly but essential parts of a manually-driven game.

## Tasks
- [ ] Turn / phase / step indicator
- [ ] "Pass priority" button per player
- [ ] "Pass until end of turn" shortcut
- [ ] Any-player chat, with player colors and timestamps
- [ ] Quick-action buttons: draw, shuffle, mulligan, untap all, change life
- [ ] Priority indicator ("any responses?") visible across all 4 seats

## Exit criteria
Two players can play a complete turn cooperatively, passing priority manually, with chat, and it feels like a game — not a demo.

Issue #6, opened 2026-04-10
S06 — Hand, battlefield, zones UI


krakenhavoc:
**Phase:** 3 · **Goal:** render and manipulate the core game zones.

## Tasks
- [ ] Hand: fan layout, hover lift, drag to battlefield
- [ ] Battlefield: free card placement, tap/untap via click, stacking for tokens/copies
- [ ] Library / graveyard / exile / command zone — clickable, searchable, reorderable
- [ ] All zone state driven by server state deltas (no local truth)
- [ ] Card art rendered from the Scryfall cache

## Exit criteria
A player draws 7 cards, plays a land, casts a creature, taps a permanent — all visible, all driven by server-authoritative deltas.

krakenhavoc:
## Kickoff — S06 branch open

Branch: `feat/s06-zones-ui`. Project card moved to **In Progress**. `docs/sprints.md` synced (also flipped S03/S04/S05 rows to **done** — they had drifted).

### Proposed phase breakdown

This is mostly a client sprint: the server already has `draw_card`, `play_card`, `move_card`, `tap`, `untap`, `untap_all`, `shuffle_library`, `mulligan`, `change_life`, `add_counter` on the action bus. The two exceptions are flagged below.

1. **Phase 1 — Server: battlefield positions + "set position" action.** See architectural call-out below.
2. **Phase 2 — Client: `CardTile` primitive.** PixiJS `Sprite` per card, art loaded from `/cards/{id}/image?size=small`, tap state shown by 90° rotation. Replace the current "count inside a rectangle" zone drawing.
3. **Phase 3 — Hand: fan layout + hover lift.** Arc-based positioning, hover raises the card above its siblings; click / drag to cast.
4. **Phase 4 — Battlefield: free placement + click-tap.** Drag to drop at (x,y), click on a permanent to tap/untap. Stacks for token/copy groups.
5. **Phase 5 — Zone browsers.** Svelte overlay (outside Pixi) for library / graveyard / exile / command — list cards, click-to-move, drag-to-reorder (library only).
6. **Phase 6 — Docs.** Update `docs/protocol.md` with whatever new action + view fields we add; check exit-criteria boxes.

### 🚨 Architectural call-out: battlefield positions

[...]

Issue #3, opened 2026-04-10
S03 — Action protocol + state deltas


krakenhavoc:
**Phase:** 1 · **Goal:** clients drive the game state via the action protocol.

## Tasks
- [ ] Action types: `draw_card`, `play_card`, `move_card` (zone→zone), `tap`, `untap`, `untap_all`, `pass_priority`, `pass_turn`, `mulligan`, `shuffle_library`, `change_life`, `add_counter`, `set_commander_damage`
- [ ] Server applies action → computes state delta → broadcasts to all players in the game
- [ ] Delta format versioned in `docs/protocol.md`
- [ ] `cmd/bridge-cli/` — CLI test client that drives a game via the protocol
- [ ] Snapshot tests of a recorded game transcript (golden files)
- [ ] Crash recovery: server writes JSON snapshots after every action

## Exit criteria
The CLI plays a full scripted turn (draw, play land, cast creature, attack, pass turn) end-to-end against the server and the resulting state matches a golden file.

PR #27, opened 2026-04-13
S05 — Deck import + 4-player table layout

### LINKED ISSUE #5: S05 — Deck import + 4-player table layout
**Phase:** 3 · **Goal:** bring decks into games and render a 4-player table.

## Tasks
- [ ] Moxfield export parser
- [ ] Archidekt export parser
- [ ] `.txt` / `.cod` / `.dek` parsers
- [ ] Validation: 100-card singleton, commander legality (from Scryfall data), color identity
- [ ] PixiJS play area canvas mounted in the client shell
- [ ] 4-player table layout: you at the bottom, 3 opponents around the top
- [ ] Placeholder card backs rendered in each zone

## Exit criteria
Import a Moxfield deck, create a game, and see your seat at a 4-player table with the right zones laid out.

krakenhavoc:
## Summary

- New `internal/deck` package: Moxfield JSON + plain-text parsers, batched-unknowns resolve, Commander validation (100-card singleton, commander legality, color identity, format legality)
- `POST /games/{id}/decks`: RolePlayer may only set own seat; RoleAdmin may set any; 422 batches all violations; sideboard is a non-fatal warning
- `Game.ReplaceDeck` swaps a seated player's library + command zone while in lobby; `Start` now requires every seat uploaded (new `ErrDeckNotUploaded`)
- Banlist is snapshot-pinned to the Scryfall index loaded at server start; mid-game rotations don't invalidate existing decks
- Partner / companion surfaced as `ErrUnsupportedMechanic` — decks fail loudly rather than silently drop the second commander
- Client: per-seat deck-status badges + inline upload form + Start gate, PixiJS table canvas (viewer at bottom, opponents clockwise), placeholder card "backs" as solid rectangles

## Design decisions

- **Server-side parsing** — single source of truth for Scryfall data; avoids shipping the ~450MB index to every client
- **Separate `/decks` endpoint** (not overloading `/join`) — lets players swap decks in the lobby without re-joining
- **`Card` metadata extended** with `type_line`, `color_identity`, `legalities`, `oracle_text` + typed `CardFace` so validation and the future rules engine can share one schema
- **Placeholder card backs** as colored rectangles — Scryfall's card back is Wizards IP, deferred to custom asset work

## Deferred (flagged in docs/sprints.md)

- Archidekt JSON, `.cod`, `.dek`, text-file-upload parsers (each a small follow-up PR)
- Partner / companion support — rejected as `ErrUnsupportedMechanic` until the rules for alternative commanders are in
- Drag, hover preview, animations on the PixiJS table (S06+)

## Test plan

- [x] `go test -race ./...` in `server/` — all packages including new `deck` (20+ test cases: parser shapes, unknown-name batching, partner sniff, every validation code, sideboard warning, JSON wire)
- [x] `npm run check && npm run lint && npm run build` — clean (adds pixi.js; ~110KB gzipped)
- [ ] Manual smoke: admin creates a game, two tabs join, each uploads a minimal valid deck, admin starts, both tabs open the game view and see the 4-player table with their own hand visible and opponents face-down

Closes #5

Sprint: S05 — Deck import + 4-player table layout


krakenhavoc (commented):
# Code review — S05 deck import + 4-player table

Two-pass review (server + client). Prioritized below.

## 🔴 Bugs to fix before merge

1. **`ErrDeckNotUploaded` → HTTP 500, not 409.** `writeLobbyError` in `server/internal/lobby/http.go` (~L582-599) has no case for it. Tests pass only because `TestStartRequiresAllDecks` calls `l.Start` directly, not through HTTP. Contradicts `docs/lobby.md:168`.
2. **Unbounded request body on `POST /games/{id}/decks`.** `decodeJSON` (`server/internal/lobby/http.go:507-517`) has no `http.MaxBytesReader`, and the endpoint isn't in the rate-limit middleware list. An authed player can OOM the server with a multi-MB Moxfield blob.
3. **Batched-error claim broken for unknown cards.** `UnknownCardError` path (`server/internal/lobby/http.go:411`) returns `{"error": "..."}` as a flat string while other 422s return `{"violations":[…]}`. Client can't highlight unknowns the way docs advertise.
4. **`client/src/lib/api.ts:116-121` docstring lies.** Promises `LobbyApiError.violations`, but `authFetch` never attaches it. Either parse the body and attach, or delete the comment. Compounds with #3.
5. **Shared `deckError`/`deckSuccess` across all games** in `client/src/routes/Lobby.svelte:92-96`. Upload failing in game A shows the same banner inside every open `<details>`. Key by game ID like `deckSources`.
6. **Warnings silently dropped on 422.** `server/internal/lobby/http.go:425-444` discards sideboard warnings whenever `fatal` is non-empty. On the client, `uploadDeck` success path only surfaces a warning *count*, never the list (`Lobby.svelte:105-107`).

## 🟡 Real concerns

7. **Partner/companion sniff runs on every card, not just commanders** (`server/internal/deck/deck.go:116`). A mainboard card whose oracle text mentions "partner" (e.g. "target creature with partner") rejects the whole deck.
8. **`ErrPlayerNotInGame` missing from `writeLobbyError`** — falls through to 500. `SetDeck` can surface it.
9. **DFC / MDFC commanders mis-validated** (`server/internal/deck/validate.go:188-202`) — only checks top-level `TypeLine` / `OracleText`, not `CardFaces[*]`.
10. **Banlist snapshot pinning claimed but not enforced.** `List.ParsedAt` captured but unused; `Validate` reads live `Legalities` from the index. Fine today (no hot-reload), invariant not in code.
11. **Double-destroy on fast unmount** in `client/src/routes/Game.svelte:52-71`. If `init()` resolves between `destroyed=true` and its `.then`, both the then-callback and cleanup call `r.destroy()`. Pixi's `Application.destroy` isn't idempotent.
12. **`canStart` duplicates server logic** (`client/src/routes/Lobby.svelte:111`) — will drift when server rules tighten. Prefer surfacing the server error.
13. **First-frame layout against fallback (1280×720)** (`client/src/lib/table.ts:111`). Users briefly see wrong size before the ResizeObserver fires.

## 🟢 Nits

- Dead `void this.lastSize;` at `client/src/lib/table.ts:127`; `TextStyle` re-allocated per render.
- `pixi.js` brings in `@xmldom/xmldom`, `gifuct-js`, `earcut` — verify the ~110 KB gzipped claim with a build report.
- `TestUploadDeckPlayerCannotSetAnothersDeck` uses a deliberately-invalid source; auth check fires before parse, so it passes trivially.
- `TestValidationErrorMarshals` asserts only one code in JSON, not the full violation shape the handler emits.
- `background: 0x0b1220` set in both Pixi and CSS — pick one source of truth.
- 3-player seating leaves the right seat empty rather than symmetric left/top.

## Recommendation

Block on #1–#6. #7–#13 can be follow-ups, but #7 and #9 have a real chance of rejecting valid Commander decks in the wild.

krakenhavoc:
Pushed three commits addressing the review. Summary:

### 🔴 Bugs — all fixed (commit `c2bbe6b`)

1. `ErrDeckNotUploaded` now maps to 409 in `writeLobbyError`; `TestStartReturns409WhenDeckMissing` exercises it through the HTTP surface.
2. `POST /games/{id}/decks` gets a 2 MiB `MaxBytesReader` cap plus a dedicated 2rps/burst-10 rate limiter; oversized uploads return 413 (`TestUploadDeckBodySizeCap`).
3. Unified 422 shape: `UnknownCardError` and the new `UnsupportedMechanicError` both project to `[]Violation` via a `Violations()` method; `writeDeckViolations` is the sole 422 emitter. Covered by `TestUploadDeckUnknownCardReturnsViolations`.
4. `LobbyApiError.violations` / `.warnings` are populated by `authFetch` from the JSON body; the misleading `api.ts` docstring is replaced with an accurate one.
5. `deckError` / `deckSuccess` / `deckViolations` / `deckWarnings` in `Lobby.svelte` are now per-game `Record<string, …>`s; `clearDeckFeedback(id)` resets one game at a time.
6. Warnings survive 422: `writeDeckViolations` carries a warnings slice when non-empty; the client renders them in a dedicated `.warnings` list.

### 🟡 Concerns — fixed where worth it

7. Partner/companion sniff gated on `e.IsCommander` only (`TestResolvePartnerSniffCommanderOnly`).
8. `ErrPlayerNotInGame` added to `writeLobbyError` → 403.
9. `isLegalCommander` walks every `CardFace` — `TestValidateDFCLegendaryCommander` pins the Esika / Prismatic Bridge case.
10. Banlist-pinning docstring rewritten (`46f21d2`) to be honest about what the code does; explicit deferral for the hot-reload case.
11. `TableRenderer.destroy` double-dispatch guard via a `disposed` flag closed over the effect.

### 🟡 Concerns — deliberately deferred

12. `canStart` still mirrors server logic. Keeping it as a UX predicate with the server as the hard authority (the 409 now comes through correctly).
13. First-frame fallback layout is unchanged; the flash is one frame and the ResizeObserver fires synchronously on observe in modern browsers.

### 🟢 Nits — picked up opportunistically (commit `46dde54`)

- Dead `lastSize` / never-called `resize()` removed from `table.ts`.
- `TextStyle` instances memoized by `(size,color)` so redraws reuse them.
- Pixi background color is a single `BG_COLOR` constant; the redundant CSS `background` is gone.
- `TestValidationErrorMarshals` → `TestViolationJSONShape`, asserting the full `Violation` schema including `card` omission.
- `TestUploadDeckPlayerCannotSetAnothersDeck` now ships a valid 100-card source and also asserts that Carol can upload her *own* deck with the same bytes.

### Verification

- `go test -race ./...` — green, including 6 new tests.
- `npm run check && npm run lint && npm run build` — clean. Main chunk ~107 KB gzipped.

PR #28, opened 2026-04-17
S06 — Hand, battlefield, zones UI

### LINKED ISSUE #6: S06 — Hand, battlefield, zones UI

krakenhavoc:
## Summary

- Server: `game.Card` grows `BattleX, BattleY` (normalised floats in `[0, 1]`, cleared on battlefield exit alongside `Tapped`/`Counters`). New `SetBattlefieldPosition` mutation + `set_battlefield_position` action clamps out-of-range inputs and errors on non-battlefield targets. CardView / client protocol mirror the addition. `scryfall_id` drift picked up in `client/src/lib/protocol.ts`.
- Client: `CardTile` primitive (PixiJS Sprite + fallback rect + name overlay) with tap rotation, hover lift, and a module-level texture cache keyed on `(scryfall_id, size)`. Face-down variant skips art fetches entirely (opponent hand / library never emits network traffic for hidden bytes).
- Client: `table.ts` renders battlefield cards as real tiles positioned by their new `battle_x`/`battle_y`, and lays the viewer's hand out as a 60° arc at the bottom of the canvas with pointer-over hover lift.
- Interactivity: click hand card → `play_card`; click battlefield card → toggle `tap`/`untap`; drag battlefield card past a 4px threshold → `set_battlefield_position` on release. Click self library → `draw_card`. Spectator / admin views (no `sendAction`) stay read-only.
- Temporary dev-controls strip in `Game.svelte` exposes "draw 7 (mulligan)", "shuffle library", "pass turn" so the S06 exit criteria are drivable without waiting for S07's turn/phase bar. Flagged in-code as `superseded by S07`.

## Design decisions

- **Normalised [0, 1] positions.** Resolution-independent: changing the render surface doesn't invalidate saved snapshots, and clamping on the server keeps the wire trustworthy.
- **Position lifecycle mirrors Tapped.** Cleared on battlefield exit so a cycled permanent doesn't re-enter at its old spot. If we want "return to previous location" behaviour for specific rules effects later, we can add it then.
- **Commander-only partner-mechanic sniff from S05 stays as-is** — not reopened in this PR.
- **Dev controls strip.** S06's exit criteria ("draws 7 cards") need some trigger; dropping a shell of a turn bar now rather than synthesising a spec for it. S07 is the real sprint for turn / phase UI.

## Deferred (flagged in docs/sprints.md)

- Battlefield stacking for tokens / copies — tokens don't exist until the rules graft track (S13+).
- Graveyard / exile / command zone browsers — bundled with S07's turn/phase overlay work.
- Library searchability / reorder — S11 (hover preview / polish track).

## Sprint

S06 — Hand, battlefield, zones UI

## Issues

Closes #6

## Test plan

- [x] `go test -race ./...` — new coverage in `mutations_test.go` (position set, clamp, non-battlefield target, zone-exit clears), `actions_test.go` (set_battlefield_position dispatch), `view_test.go` (BattleX/Y projection + JSON wire).
- [x] `npm run check && npm run lint && npm run build` — clean; main bundle ~116 KB gzipped (up ~10 KB from S05 due to PixiJS BitmapFont chunk).
- [ ] Manual smoke: admin creates a 2-player game, both players join and upload decks, admin starts, each clicks "draw 7", drags a card from hand onto the battlefield, clicks it to tap. Verify both tabs see the same state.


krakenhavoc:
Pushed fixes in `6f71013` for the three 🔴 bugs from the review:

1. **Server NaN/Inf clamp.** `clampUnit` now uses `!(v >= 0)` so NaN and -Inf fall into the zero branch before `SetBattlefieldPosition` can write them onto a card. Without this, the stored NaN would bomb `encoding/json` and poison every subsequent snapshot. Covered by `TestSetBattlefieldPositionRejectsNaNAndInf` (NaN on x and y, -Inf, +Inf, upper-clamp symmetry).

2. **Client memory leak.** `TableRenderer.render()` now destroys each detached child subtree via `destroy({ children: true })`. Previous code called `removeChildren()` alone, which left Sprite / Graphics / Text / Container resources in GPU memory — every snapshot + every ResizeObserver tick orphaned the previous batch and the set grew without bound. Shared Textures (card art) are intentionally preserved by omitting `texture: true` so the Pixi Assets cache stays warm.

3. **Hover-ghost on re-render.** Module-scope `hoveredCardID` in `card-tile.ts` now persists the pointer target across tile teardown/recreate cycles; `isHovered(instance_id)` lets `drawHandFan` restore hover on a replacement tile immediately after `addChild`. A snapshot arriving while the user is hovering a card no longer shows a one-frame drop flash.

Verification:
- `go test -race -count=1 ./...` — green (game 81.8% cov, protocol 86.8%, actions 72.4%).
- `npm run check && npm run lint && npm run build` — clean.

The remaining 🟡/🟢 items from the review are logged as follow-up issues rather than included here, since they're deferrable:

- #29 — `BattleX/Y` omitempty conflates unpositioned with at-origin
- #30 — atomic `play_card` with optional (x, y) to eliminate play-to-position pop
- #31 — battlefield position lost on battlefield→shared→battlefield flicker (latent)
- #32 — unbounded texture cache — needs eviction
- #33 — throttle + prioritise card-image fetches on first render
- #34 — guard texture assignment against destroyed Sprite (dev-mode warning)
- #35 — shrink Card struct: float32 + field reordering (`good first issue`)
- #36 — tighten test coverage for `set_battlefield_position` (`good first issue`)

Commits on the branch: `d47c497` (server phase 1), `91b59a5` (CardTile + battlefield), `135c225` (hand fan + hover lift), `cc6d3f8` (interactivity), `64defed` (library click + dev controls), `f03bdcd` (sprints.md), `6f71013` (this fix bundle).

PR #195, opened 2026-04-27
feat(s19): 5 ETB-trigger cards via the auto-fire pipeline (sub-PR 3/8)

krakenhavoc:
## Summary

- First batch of catalog cards using the S19 `Triggered` slot — 3 new + 2 migrated from S14 `OnETB` direct-call
- All ETB triggers (no dies/cast/upkeep yet — those split across sub-PRs 4-7)
- Sub-PR 1's framework + sub-PR 2's optional-prompt path are both exercised end-to-end

## Cards

**New:**
- **Mulldrifter** — ETB draw 2 (mandatory) + flying. Evoke deferred to S29.
- **Reclamation Sage** — ETB destroy artifact/enchantment (optional `you may`)
- **Acidic Slime** — ETB destroy artifact/enchantment/land (mandatory) + deathtouch

**Migrated** (drop `OnETB`, declare `Triggered` + `OptionalPrompt`):
- **Eternal Witness** — ETB return top-of-graveyard (optional)
- **Solemn Simulacrum** — ETB search basic land tapped (optional). Dies-half stays for sub-PR 4.

## Sprint

S19 — Auto-fire triggered abilities

## Issues

Relates to #69 (sub-PR 3 of 8). **Stacks on [#194](../pull/194); merge order: #193 → #194 → this PR.**

## Sandbox patterns documented in code

- `Build` runs the effect inline (`DrawN`, `DestroyTarget`, `SearchLibrary` via `effect_api` ForEffect helpers) and returns `nil` — no `StackItem` queues. Effect events provide visibility. CR-correct stack-resolution timing for triggered abilities is deferred.
- `pickFirstOpponentNonland` is the shared auto-pick helper for ETB-destroy cards. Real target picker lands with S20.
- Eternal Witness keeps the S14 "top of graveyard" auto-pick simplification.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./...` — all packages green (cards/effects + game both green)
- [x] 4 new card tests + 1 helper (`answerLatestTriggerPrompt`)
- [x] 2 existing S14 tests updated to answer the new optional prompt
- [ ] **Manual testing — please verify in a browser:**
  - Cast Mulldrifter → on resolve, you draw 2 cards automatically (no prompt)
  - Cast Reclamation Sage with an opponent's artifact in play → prompt appears → "Yes" destroys it; "No" leaves it
  - Cast Acidic Slime with an opponent's land in play → land destroyed automatically (no prompt — mandatory)
[...]

Issue #69, opened 2026-04-19
S19 — Auto-fire triggered abilities


krakenhavoc:
Replaces S13.1's manual `announce_trigger` for catalog cards. After this sprint, ETB triggers, dies-triggers, "at the beginning of your upkeep" effects, and the rest of the Commander trigger surface fire automatically.

## Sub-PRs

- [ ] **1/8 — `s19-dispatcher`**: `TriggeredAbility` type on `Spec`; per-game `triggerHarvester` listener; LKI snapshot for LTB triggers
- [ ] **2/8 — `s19-prompt`**: `KindTriggerPrompt` frame + `TriggerPromptDialog.svelte` for optional / modal triggers
- [ ] **3/8 — `s19-etb-triggers`**: 5 ETB cards (Mulldrifter, Eternal Witness migration, Reclamation Sage, Acidic Slime, Solemn Simulacrum)
- [ ] **4/8 — `s19-dies-triggers`**: 4 dies-trigger cards (Solemn Simulacrum dies half, Skullclamp migration, plus 2)
- [ ] **5/8 — `s19-upkeep-triggers`**: 4 upkeep / step cards (Phyrexian Arena, Sylvan Library, Mana Crypt, plus 1)
- [ ] **6/8 — `s19-cast-triggers`**: 4 cast / opponent-draws cards (Smothering Tithe, Esper Sentinel, plus 2)
- [ ] **7/8 — `s19-combat-triggers`**: 4 combat-damage cards (Edric Spymaster of Trest, Bident of Thassa, plus 2)
- [ ] **8/8 — `s19-tests`**: APNAP-ordering, LKI-correctness, optional/modal coverage, fallback-to-manual coverage

## Reference

Full plan in `/home/node/.claude/plans/s19-auto-fire-triggers.md`.

## Manual fallback preserved

`AnnounceTrigger` from S13.1 stays for unimplemented cards and "hidden info" triggers (cards in hand with cast-replacement triggers).

## Depends on

S14 (event log + listener registry, pre-wired). S17 prompt frame for sub-PR 2. S13.1 `PendingTriggers` queue + APNAP drain (auto-fires populate the same queue).

Estimated 3-4 weeks.

krakenhavoc:
S18 carry-in (per [ADR 0014](../blob/main/docs/decisions/0014-combat-keywords.md#10-champion-of-lambholt-deferred-to-s19)):

[...]

PR #196, opened 2026-04-28
test(s19): Playwright e2e suite — 10 tests covering trigger UI + flow (sub-PR 3.5)

krakenhavoc:
## Summary

- Adds an e2e Playwright suite that exercises every S19 sub-PR 3 ETB card through the real browser → WS → engine path
- 10 tests covering mandatory + optional yes/no paths for all 5 cards, plus prompt-routing (only chooser sees the modal) and prompt-text wire plumbing
- Caught a real bug during development: 3 fabricated oracle IDs that turned the trigger harvester into a no-op when cards came in via Scryfall deck import. Fix shipped on PR #195.

## Sprint

S19 — Auto-fire triggered abilities

## Issues

Relates to #69 — supplements PR #195. **Stacks on [#195](../pull/195); merge order: #193 → #194 → #195 → this PR.**

## Why this PR exists separately

The Go integration tests in #195 assert on engine state (PendingChoices, hand counts, graveyard contents). They never touch the wire shape or the Svelte modal. This suite covers what they can't:

- the `kind: "trigger_prompt"` PendingChoiceView round-trips correctly
- only the chooser's browser pops the dialog; the opponent's stays clean
- the prompt's question copy actually reaches the modal header
- "Yes" → effect fires; "No" → effect drops

## Architecture (more detail in the source headers)

- Two admin-token WS connections, one bound to each seat. Merges their snapshots so tests can look up cards by name in either player's hand zone (CR 400.2 wholesale-hides hand+library from a non-bound admin spectator).
- `seedHandWithCard` pumps `draw_card` until the named card surfaces — library card identities are redacted on the wire; the only zone with name-visible cards is the owner's own hand (knower-stamped at deal time).
- `setupS19Game` 4-second pre-sleep throttles below the lobby's join rate limiter (5-burst, 1/sec — 6 joins in a 5-test run trips it without the sleep).
- Tests are serial — `test.describe.configure({ mode: "serial" })`. The lobby has global state; parallel tests would race on game IDs.

## Test plan

- [x] Run: `cd tests-e2e && ./node_modules/.bin/playwright test s19-triggers.spec.ts`
- [x] All 10 tests pass; ~60s wall clock for the suite (4s setup × 10 + ~2s body each)
- [x] Re-ran clean a second time to confirm stability
- [ ] CI integration is not in scope here; if/when wanted, lift the 4s sleep into a fixture and run with `process.env.CI=1` for deterministic retries


krakenhavoc:
Superseded by #199, which rebased this PR's commits cleanly onto main (squash-merging #193 broke the stacked bases with add/add conflicts). Content merged unchanged.

PR #197, opened 2026-05-29
fix(s19): warn on optional triggers with no legal target

krakenhavoc:
## Summary

Follow-up to the S19 trigger stack, found during live manual testing. The sandbox auto-targeter (`pickFirstOpponent*`) only ever picks **opponent**-controlled permanents, and there's no target-selection UI until S20. So an optional trigger whose only candidate is the caster's *own* permanent — or whose graveyard is empty — silently no-ops on "Yes". That reads as a bug:

- Reclamation Sage "Yes does nothing" (the only artifact was the caster's own Sol Ring)
- Acidic Slime destroying a land instead of the caster's own Sol Ring
- Eternal Witness "Yes does nothing" (empty graveyard)

Verified live in a running game that the engine itself is correct — these were all setup conditions. This PR surfaces the condition instead of hiding it.

## Changes

- **`HasLegalTarget` predicate** on `TriggeredAbility`, evaluated at prompt-queue time → `PendingChoice.NoLegalTarget` → wire `no_legal_target`.
- Reclamation Sage / Eternal Witness implement it.
- `ChoicePromptModal` shows an amber "answering Yes will pass without effect" warning and relabels the button **"Yes (no effect)"**.
- `Game.svelte`: auto-pass no longer fires while the viewer has an open pending choice (it raced the prompt modal).
- Regression tests for own-artifact-only, empty-graveyard, and the opponent-artifact positive control — paths neither the unit nor e2e suites covered.

## Test plan

- [x] `make test` — Go tests incl. 3 new + client check
- [x] Live-verified in browser against a running game

Sprint: S19 — Auto-fire triggered abilities
Issue: #69


krakenhavoc:
Superseded by #199, which rebased this PR's commits cleanly onto main (squash-merging #193 broke the stacked bases with add/add conflicts). Content merged unchanged.

PR #217, opened 2026-09-08
feat(s21): activated abilities in the catalog (sub-PR 2)

krakenhavoc:
## Summary

Stacked on #216. CR 602 activated abilities were the one thing the catalog couldn't express: `ActivateAbility` took a card ID and a free-text label and pushed a stack item with no cost and no effect. Players tapped the permanent themselves, dragged the sacrificed creature to the graveyard themselves, and resolved the effect in their heads. That can't carry S21 — an aristocrats deck *is* its sacrifice outlets, the cost is half the card, and the dies-triggers it causes are the whole strategy.

- **Engine**: `Spec.Activated` → `game.ActivatedAbilityShape`. An entry is a label, a struct cost (`Tap` / `SacrificeSelf` / `SacrificeOther` / `Mana` / `Life`), an optional target clause validated exactly as a spell's, a sorcery-speed flag, and the **same `Effect` closure S19 gave triggered abilities** — so resolution needed no new code at all; `resolveTopAbilityLocked` can't tell the two apart.
- **Catalog**: Goblin Bombardment (the S21 exit-criteria outlet), Carrion Feeder (grows; can legally and pointlessly eat itself), Krenko Mob Boss (tap cost, Goblins counted at resolution).
- **Client**: activated abilities join the existing right-click ability menu under the mana abilities, greyed with a reason when unpayable; a sacrifice cost opens a picker *before* targeting, matching the order costs are paid in.

## Two properties that are load-bearing

Both are tested directly rather than left to emerge:

1. **Every cost component is validated before any is paid.** A tapped Krenko fails without sacrificing; an illegal target fails without tapping. Payment order is mana → tap → life → sacrifice, because sacrifices move cards and invalidate the pointer the tap needs.
2. **Costs are paid at announce** (CR 601.2h), so a creature sacrificed to Goblin Bombardment puts its dies-trigger on the stack *above* the ability and it resolves first. That ordering is the archetype.

`SacrificeOther` reuses `TargetSpec` for its predicate because the shape is identical, but a sacrifice cost is **not** targeting — hexproof never applies, and the client picks from a list rather than the board. The controller restriction (CR 701.17b) lives in cost validation.

Tap costs now enforce summoning sickness, which the free-form path never did; `CardView.summoning_sick` rides the wire so the menu greys the entry instead of failing the click. The free-form `ActivateAbility` stays for non-catalog cards — the action layer picks by whether the payload names an `ability_index`.

## Sprint

Sprint: S21 — Tokens, sacrifice, aristocrats (sub-PR 2)

## Issues

n/a

## Test plan

- [x] `go test -race ./internal/...`; `go vet ./...`; gofmt
- [x] `effects/activated_test.go`: catalog wiring + cost shapes; Bombardment pays at announce and damages at resolution; rejects a missing / opponent's / non-creature / duplicated sacrifice **without paying anything**; the cost's dies-trigger resolves before the ability; Feeder grows and self-sacrifices harmlessly; Krenko counts at resolution, respects sickness then tap state; opponent's permanent and out-of-range index rejected
- [x] vitest (98) for the ability prompt; `svelte-check`; lint
- [ ] Manual: right-click Goblin Bombardment → pick a creature to sacrifice → click a target → damage lands after the dies-trigger

## Notes for review

**Base**: this and #216 branch from main *before* #215 merged (I can't fetch from my sandbox). GitHub reports both mergeable. I deliberately kept this PR independent of #215's `viewOfLegalTargets` signature change rather than write against code I couldn't compile — the cost is one small duplicated helper (`abilityLegalTargets` in view.go) that a follow-up should fold back in. Happy to rebase both branches properly if you'd rather.

**Deferred, in ADR 0020**: Ashnod's Altar fits neither surface — it's a mana ability (must skip the stack, CR 605.3a) whose cost sacrifices a *different* creature (needs the picker built here for stack-using abilities). It wants a third path where activation collects a cost choice then resolves immediately. Also out of scope: X in an activation cost, counters as a cost, "activate only once each turn".


krakenhavoc:
@copilot resolve the merge conflicts in this pull request

Issue #229, opened 2026-09-09
Develop branch + preview environment at dev.cmd.labxp.io


krakenhavoc:
Everything used to land on `main` and deploy straight to the table people play on. `develop` now exists as an integration branch with its own deployment, the flag machinery for tooling that is useful in preview and dangerous in production, four dev tools, and the host provisioning to stand it up.

Design: [ADR 0023](docs/decisions/0023-develop-environment.md). Runbook: [docs/environments.md](docs/environments.md).

## Merged into `develop`

| PR | Branch | What |
|---|---|---|
| #238 | `chore/develop-environment` | environment foundation |
| #239 | `feat/dev-card-spawner` | card spawner |
| #240 | `feat/dev-seat-swap` | seat swap |
| #241 | `chore/dev-host-provisioning` | provisioning script + Caddy |
| #242 | `feat/dev-replay-scrubber` | replay scrubber |

Merge commits rather than squashes, so the stack's ancestry survives and each of the 18 commits stays readable.

`develop` has **not** been merged to `main` — that is a production deploy and a separate decision.

## The design, in one rule

`CMDCTRL_ENV` is the only switch. `appenv.LoadFeatures` returns the zero value whenever the environment is not dev, *before* consulting any per-feature variable, so a `CMDCTRL_DEV_*` line in the production env file is inert and gets a WARN on boot. `GET /config` tells the client what to draw; `requireDev` independently refuses every dev route with **404, not 403**, wrapped *outside* `auth` so an anonymous probe cannot confirm the route exists.

## Dev tools

- **Frame inspector** — raw WS frames with seq, filters, pause, copy
- **Card spawner** — any Scryfall card into any zone; real OracleID so the effect catalog matches; ETB fires; spawns land in the replay
- **Seat swap** — view and act as any seat from one browser
- **Replay scrubber** — step the board through recorded snapshots; hides the live toolbar while scrubbing

Three of the four are client-only. Only the card spawner needed a server route — the action-registry gate ADR 0023 anticipated was never built, because nothing needed it.

Not built, deliberately: seeded shuffles (the spawner reaches a chosen board state directly) and life/counters/phase controls (admin sessions already have them through the ordinary action protocol).

## Remaining — on the host

The dev CD job will fail at the rsync step until this is done. It touches only `-dev` paths and never production.

1. `sudo deploy/provision-dev.sh --check`, then without `--check`
[...]

PR #245, opened 2026-09-09
chore(deploy): rework the develop preview as a second VM

krakenhavoc:
## Summary

- ADR 0023 §2 replaced: the preview is its own Proxmox VM, not a second service unit sharing production's box. The two hosts are now identical below the fqdn.
- `docs/environments.md` rewritten and `deploy/README.md` repointed at the HomeLab cloud-init template, which is now the single source of truth for both hosts.
- The four files that only made sense on a shared box are deleted: `provision-dev.sh`, `cmd-and-ctrl-dev.service`, and all of `deploy/caddy/`.

Infra half is [krakenhavoc/HomeLab#50](https://github.com/krakenhavoc/HomeLab/pull/50).

### Why §2 was wrong

The original reasoned from the cost of a second rented VPS. The host is not a rented VPS — it is a Proxmox node driven by Terraform, where a second VM is a map entry and its Cloudflare tunnel can be declared beside it. Once that was clear the comparison inverted, and the namespacing (`/opt/cmd_and_ctrl-dev`, `:8081`, `dev.env`, `MemoryMax=768M`, a symlink to production's Scryfall dump) all turned out to be machinery serving a constraint that did not exist.

The isolation argument stands on its own regardless: sharing a box meant a runaway preview could push production into swap, and `MemoryMax` bounded that rather than removing it.

The gain is not just tidiness. Because the two hosts are now identical below the fqdn, **the CD recipe that deploys the preview is the recipe that deploys production** — so a preview deploy rehearses the thing it is meant to rehearse. An environment reached by a different deploy path tests the deploy path as much as it tests the code.

---

## ⚠️ This PR is incomplete on its own — the CD job is now actively wrong

The merged `cmdctrl-cd` job deploys `develop` to `vars.CMDCTRL_HOST` — **production's host** — into `/opt/cmd_and_ctrl-dev`, and restarts `cmd-and-ctrl-dev`. Under the two-VM model none of those exist, so a push to `develop` after merging this will fail partway through, on production's box.

I could not push the fix: GitHub Apps cannot write under `.github/workflows/` without the `workflows` permission, which this one does not have. The patch is below and has to be applied by hand (or grant that permission and I'll push it directly next time).

**Apply this before merging, or immediately after — but before the next push to `develop`.**

### Edit 1 — replace the `cmdctrl-cd` `env:` block

Everything below the fqdn is a constant now, so this stops branching on the ref for five of the six values.

```yaml
    env:
      # Production and the develop preview are two VMs on the same
      # Proxmox node, identical below the fqdn (ADR 0023). So this job
      # branches on exactly two things — which host, and which
      # environment the service is told it is — and on nothing else.
      # Every path, the unit name and the port are the same constants
      # for both targets, which is what makes a preview deploy a
      # rehearsal of the production deploy rather than a lookalike.
      APP_ENV: ${{ github.ref == 'refs/heads/develop' && 'dev' || 'prod' }}
      # Gates the two Discord bot steps. Nothing else branches on it.
      IS_DEV: ${{ github.ref == 'refs/heads/develop' }}

      SSH_OPTS: -o StrictHostKeyChecking=accept-new -o ConnectTimeout=10

      # Identical on both VMs — written by the HomeLab cloud-init
      # template, which is the single source of truth for both hosts.
      # Named here rather than inlined so the steps below read the same
      # whichever target they are pointed at.
      DEPLOY_ROOT: /opt/cmd_and_ctrl
      WEB_ROOT: /var/www/cmdctrl-client
      ENV_FILE: /etc/cmd_and_ctrl/env
      SERVICE: cmd-and-ctrl
      # Loopback port the service binds (CMDCTRL_ADDR in its env file).
      # Used only by the post-deploy environment check below.
      LOCAL_PORT: "8080"
```

Note `CMDCTRL_HOST` is **gone** from this block — it is resolved in a step instead, for the reason in edit 2.

### Edit 2 — replace the whole `Verify CMDCTRL_HOST is set` step

```yaml
      # The deploy target is resolved in shell rather than with a
      # `${{ cond && vars.A || vars.B }}` ternary, on purpose. That
      # ternary is unsafe for exactly this pair: if CMDCTRL_DEV_HOST
      # were unset, `cond && vars.CMDCTRL_DEV_HOST` evaluates to the
      # empty string, which is falsy, so the expression falls through
      # to the `||` branch — and a develop push would deploy itself,
      # dev features and all, onto production. Fail loudly instead.
      - name: Resolve deploy target
        env:
          PROD_HOST: ${{ vars.CMDCTRL_HOST }}
          DEV_HOST: ${{ vars.CMDCTRL_DEV_HOST }}
        run: |
          if [ "${IS_DEV}" = "true" ]; then
            host="${DEV_HOST}"
            host_var="CMDCTRL_DEV_HOST"
          else
            host="${PROD_HOST}"
            host_var="CMDCTRL_HOST"
          fi

          if [ -z "${host}" ]; then
            echo "::error::Repo variable ${host_var} is not set. Configure under Settings → Secrets and variables → Actions → Variables."
            exit 1
          fi

          # Both variables pointing at the same box is the single
          # mistake that turns a preview deploy into a production
          # outage. Cheap to check, so check it.
          if [ "${IS_DEV}" = "true" ] && [ "${host}" = "${PROD_HOST}" ]; then
            echo "::error::CMDCTRL_DEV_HOST and CMDCTRL_HOST both point at ${host}. These are meant to be separate VMs; deploying develop here would overwrite production."
            exit 1
          fi

          echo "CMDCTRL_HOST=${host}" >> "${GITHUB_ENV}"

          echo "Deploy target: krkn@${host}  (from ${host_var})"
          echo "Environment:   ${APP_ENV}"
          echo "Service:       ${SERVICE}, 127.0.0.1:${LOCAL_PORT}"
          echo "Deploy root:   ${DEPLOY_ROOT}"
          echo "Web root:      ${WEB_ROOT}"
```

The ternary point is the one bit of this worth reading twice. `${{ x && vars.A || vars.B }}` is the idiom everyone reaches for, and for a host pair it fails in the worst possible direction: an unset `CMDCTRL_DEV_HOST` silently resolves to production, and the *first* symptom is `develop` live on `cmd.labxp.io` with card spawning enabled. Resolving in shell turns that into a failed run.

### Edit 3 — `Deploy bot binary`, two occurrences of the hardcoded path

```diff
-            krkn@${CMDCTRL_HOST}:/opt/cmd_and_ctrl/bot/bin/ || {
-              echo "::warning::Bot deploy skipped — /opt/cmd_and_ctrl/bot/bin/ may not exist yet. Create it on the host before the bot rolls out."
+            krkn@${CMDCTRL_HOST}:${DEPLOY_ROOT}/bot/bin/ || {
+              echo "::warning::Bot deploy skipped — ${DEPLOY_ROOT}/bot/bin/ may not exist yet. Create it on the host before the bot rolls out."
```

### Edit 4 — `Verify reported environment`, name the host in the output

```diff
-              echo "::error::${SERVICE} did not report env=${APP_ENV}. Check CMDCTRL_ENV in ${ENV_FILE}."
+              echo "::error::${SERVICE} on ${CMDCTRL_HOST} did not report env=${APP_ENV}. Check CMDCTRL_ENV in ${ENV_FILE} on that host."
               exit 1
               ;;
           esac
-          echo "Confirmed: ${SERVICE} is running as ${APP_ENV}."
+          echo "Confirmed: ${CMDCTRL_HOST} is running as ${APP_ENV}."
```

With two identical boxes, "the service is active" is no longer a useful line in a log that doesn't say *which* box.

### Drive-by while you're in there

The bug-report-token comment block currently sits above the **deployment identity** step rather than above the step it describes — a leftover from when the identity step was inserted. Worth moving down one step.

### What changed, in one table

| | before (one VM) | after (two VMs) |
|---|---|---|
| host | `CMDCTRL_HOST` for both branches | `CMDCTRL_DEV_HOST` / `CMDCTRL_HOST` |
| `DEPLOY_ROOT` | `/opt/cmd_and_ctrl-dev` vs `/opt/cmd_and_ctrl` | `/opt/cmd_and_ctrl` |
| `WEB_ROOT` | `…/cmdctrl-client-dev` vs `…/cmdctrl-client` | `/var/www/cmdctrl-client` |
| `ENV_FILE` | `/etc/cmd_and_ctrl/dev.env` vs `…/env` | `/etc/cmd_and_ctrl/env` |
| `SERVICE` | `cmd-and-ctrl-dev` vs `cmd-and-ctrl` | `cmd-and-ctrl` |
| `LOCAL_PORT` | `8081` vs `8080` | `8080` |
| bot steps | prod only | prod only (unchanged) |
| target check | "is `CMDCTRL_HOST` set" | resolves + rejects unset **and** same-host |

---

## Then: the setup that can only be done from the settings UI

In this order.

1. **Merge [HomeLab#50](https://github.com/krakenhavoc/HomeLab/pull/50) and `terraform apply`.** Add the four repo secrets there first — `CMD_AND_CTRL_DEV_ADMIN_TOKEN` (≥16 chars, and Terraform rejects it if equal to production's), `CLOUDFLARE_API_TOKEN` (Tunnel:Edit + DNS:Edit, nothing broader), `CLOUDFLARE_ACCOUNT_ID`, `CLOUDFLARE_ZONE_ID`. The apply migrates the existing prod VM into the `for_each` via a `moved` block — read that plan before confirming; it should show **no** destroy of `cmd_and_ctrl["prod"]`.
2. **This repo → Settings → Secrets and variables → Actions → Variables**, two new **variables** (not secrets):
   - `CMDCTRL_DEV_HOST` — the dev VM's address. Must differ from `CMDCTRL_HOST` or the new resolve step fails the run.
   - `CMDCTRL_DEV_PUBLIC_BASE_URL` — `https://dev.cmd.labxp.io`. Optional; the job defaults to that string, but setting it explicitly means the default is never load-bearing.
3. **Discord Developer Portal → OAuth2 → Redirects**: add `https://dev.cmd.labxp.io/auth/discord/callback`. Same application, same client ID and secret. Do **not** create a second bot.
4. **Settings → Branches → add a rule for `develop`**: require the `cmdctrl-ci` check, and require a PR. Without it, `develop` is a branch anyone can push straight to, which quietly removes the gate this whole ADR is about.

Do **not** copy production's `CMDCTRL_ADMIN_TOKEN` or `CMDCTRL_GITHUB_TOKEN` to the preview. The first would make a leak from the lower-trust box a compromise of the live table (ADR 0023 §8); the second would let preview bug reports file issues as production.

## Known gap, not fixed here

`AGENTS.md` still says "One VPS. One database. One deployable per service." in §2, and calls the bot's host "the prod VPS" in §5. That ground rule is, as far as I can tell, the original source of the reasoning that produced the wrong §2. Its §5 env-var list also never got `CMDCTRL_ENV` or the four `CMDCTRL_DEV_*` feature flags added when they shipped in #238–#242. Left for a focused follow-up rather than hand-retyping a 30 KB file through the API.

## Sprint

S21 — Tokens, sacrifice, aristocrats

## Issues

#229

## Test plan

- `actionlint` clean on the rewritten workflow (exit 0; the remaining shellcheck `SC2086`/`SC2029` infos are pre-existing and on unchanged steps — `${SSH_OPTS}` *must* stay unquoted to word-split, and client-side expansion in the `ssh` commands is intentional). The new `Resolve deploy target` step is shellcheck-clean.
- The resolve logic exercised against all five cases, outside Actions:

  | case | expected | got |
  |---|---|---|
  | develop, both set, distinct | ok → dev host | ✅ `CMDCTRL_HOST=dev.lan` |
  | develop, `CMDCTRL_DEV_HOST` unset | fail, names the var | ✅ exit 1 |
  | develop, dev host == prod host | fail, same-host guard | ✅ exit 1 |
  | main, prod set | ok → prod host | ✅ `CMDCTRL_HOST=prod.lan` |
  | main, prod unset | fail, names the var | ✅ exit 1 |

- `docs/environments.md` diff verified as 2 additions / 2 deletions — the intended typo fix and nothing else.
- **Not yet run against real hosts.** The end-to-end proof is the first `develop` push after the apply: it should land on the dev VM and pass "Verify reported environment" with `env=dev`, and `curl -s https://cmd.labxp.io/config` should still report `env=prod` with all four features `false`.


krakenhavoc:
**Correction to a commit message in this branch.**

`4e96d0d` ("deploy: drop the hand-maintained dev Caddyfile") justifies the deletion with:

> and this one already was: it lacked the `/auth`, `/avatars`, `/bugreport` and `/logout` routes the template's `@api` matcher now carries.

That is wrong, and the diff in this PR shows it — the deleted file's matcher reads:

```
@api path /ws /config /dev /dev/* /admin/* /games /games/* /cards/* /auth/* /avatars/* /bugreport /bugreport/* /me /logout /healthz
```

It had all four. **Production's** Caddyfile was the one missing them, which is the gap recorded in ADR 0023's consequences and fixed in HomeLab#50. I conflated the two while writing the commit message.

The deletion itself stands on other grounds: the file hardcodes `127.0.0.1:8081` and `/var/www/cmdctrl-client-dev`, neither of which exists under the two-VM model, and it duplicates a site block the cloud-init template now generates for both hosts. A stale duplicate of generated config is worse than no copy — it reads as authoritative.

Leaving the commit message as-is rather than rewriting the branch; this comment is the correction.

krakenhavoc:
@copilot resolve the merge conflicts in this pull request

Issue #94, opened 2026-04-19
S29 — Alt-cast paths from non-hand zones


krakenhavoc:
## Scope (~30 cards)

**Key decision:** introduce \`CastableZones []Zone\` per card (default = \`[Hand]\`). Flashback adds \`Graveyard\` with predicate "has flashback marker"; suspend adds \`Exile\` with predicate "time counters == 0". Cast dialog walks all legal zones and surfaces all legal cast paths as separate buttons with their costs.

**Mechanics:**
- [ ] Flashback (CR 702.34) — cast from graveyard for flashback cost; exile after
- [ ] Madness (CR 702.35) — replacement on discard: exile face-down with marker; may cast for madness cost
- [ ] Foretell (CR 702.143) — sorcery-speed `foretell_card` action: pay {2}, exile face-down; cast for foretell cost on a later turn
- [ ] Escape (CR 702.144) — cast from graveyard, exile N cards from graveyard as additional cost
- [ ] Suspend (CR 702.62) — sorcery-speed exile with N time counters; auto-fire removes one each upkeep; cast for free with haste-until-EOT when last removed
- [ ] Cycling (CR 702.32) — activated ability of cards in hand: pay cost, discard, draw a card; cycling triggers fire from hand (CR 702.32c)
- [ ] Splice (CR 702.47) — addon to instants/sorceries

**Cards:**
- [ ] 8 flashback (Faithless Looting, Past in Flames, Snapcaster Mage, Lingering Souls, Cackling Counterpart, Increasing Vengeance, Memory's Journey, Deep Analysis)
- [ ] 4 madness (Asylum Visitor, Fiery Temper, Big Game Hunter, Avacyn's Judgment)
- [ ] 4 foretell (Saw It Coming, Behold the Multiverse, Augury Adept, Reflections of Littjara)
- [ ] 4 escape (Underworld Breach, Uro, Kroxa, Klothys)
- [ ] 4 suspend (Greater Gargadon, Living End, Hypergenesis, Lotus Bloom)
- [ ] 4 cycling (Decree of Pain, Decree of Justice, Eternal Dragon, New Perspectives)
- [ ] 2 splice (Glacial Ray + Goryo's Vengeance pattern)

## Builds on
[...]

PR #411, opened 2026-09-11
feat(cards): flashback — cast from the graveyard, then exile (refs #94)

krakenhavoc:
## Summary
- Flashback (CR 702.34), the first mechanic on top of #409's `CastableZones` — and the one the foundation was shaped around. Two declarations compose: `CastableZones` opens the graveyard, an `AlternativeCost` bound to it carries the price.
- The `Flashback()` constructor bundles the price, the zone binding, and `ExileOnLeavingStack` — the half a hand-rolled cost would forget, and the half that keeps flashback from being a one-card infinite.
- Three cards: **Faithless Looting** `{R}` / flashback `{2}{R}`, **Lingering Souls** `{2}{W}` / flashback `{1}{B}`, **Cackling Counterpart** `{1}{U}{U}` / flashback `{5}{U}{U}`.

**Stacked on #409.** Review that one first; this branch contains its commit until it merges.

## Notes

**"Any time it would leave the stack" is the whole mechanic.** CR 702.34a's clause is a *replacement*, not an exile appended to the resolution, and the difference is observable on every exit from the stack that isn't a resolution:

- A flashed-back spell that **fizzles** is exiled.
- A flashed-back spell answered by **Hinder** is exiled rather than shuffled into its owner's library — the replacement beats the counter's chosen destination.

Both are wrong *in the player's favour* under the naive implementation, because the card comes back to the graveyard to be flashed back again. `routeStackCardToGraveyardLocked` now takes the cost that was paid, and `CounterSpell` overrides its destination.

**Why these three cards.** Faithless Looting is the simplest possible case. Lingering Souls is the card that explains why flashback is a cast *path* rather than a second mana cost — `{2}{W}` and `{1}{B}` are different colours. Cackling Counterpart is the first flashback card with a target clause, which is what #409's routing of the graveyard cast through the Board's ordinary prompt chain exists for.

**Five of #94's eight flashback cards are not here**, and the reasons are worth recording so the next person doesn't re-derive them:

| Card | Blocked on |
|---|---|
| Deep Analysis | Flashback is "{1}{U}, **Pay 3 life**". `AlternativeCost` has no life component; `AbilityCost.Life` exists and is the obvious place to borrow from. |
| Past in Flames | *Grants* flashback to every instant and sorcery in your graveyard. Needs a dynamic `CastableZones`, not a catalog-declared one. |
| Snapcaster Mage | Same — grants flashback to one target card until end of turn. |
| Memory's Journey | Three targets across a graveyard and a library, plus a shuffle. |
| Increasing Vengeance | Spell copying, which has no primitive. |

All five stay out under ADR 0037's whole-text rule.

## Sprint
S29 — Alt-cast paths from non-hand zones

## Issues
Relates to #94. Depends on #409.

## Test plan
- [x] `gofmt -l .`, `go vet ./...`, `go test ./...` — full server suite green
- [x] New: `server/internal/game/flashback_test.go` — 5 cases covering all three exits from the stack (resolve, fizzle, counter-with-a-destination), that a **printed-cost** cast of the same card still goes to the graveyard (the replacement belongs to the cost, not the card), and that the exiled card cannot be cast again.
- [x] New: `server/internal/cards/effects/flashback_test.go` — the three cards end to end against the **real** catalog: each declares both halves with the right cost, Faithless Looting draws through its graveyard cast, Lingering Souls makes two flying Spirits, Cackling Counterpart copies its target, and the two wrong cast paths (flashback from hand, graveyard cast at the printed cost) are refused.


krakenhavoc:
Rebased-by-merge now that #409 has landed, so the "stacked on #409" note in the description is stale — the diff here is just flashback.

`origin/main` was briefly unbuildable while this was in flight (`GoldToken` declared twice, from #378 and #379 merging back-to-back), so this branch also carries the merge that picks up #417's fix. Merge commits rather than a rebase because force-pushing is off the table.

Full server suite, `svelte-check` and vitest are all green locally against the merged tree.

PR #350, opened 2026-09-11
fix(cards): Thought Vessel, Stasis, and nine other stale simplifications — closes #338

krakenhavoc:
Closes #338. The report was *"I was forced to discard at my end step"* with a Thought Vessel out — **a card that is in the catalog** and whose file says the static half is a no-op. That comment was written against an engine we no longer have: hand-size enforcement landed in S13.4.

**The sweep found something worse than the reported bug.**

## ⚠️ Stasis had no off switch
Its card file claimed the upkeep-sacrifice half "lands with S19". S19 triggers and `PayUnless` shipped long ago. So **Stasis froze the table's untap steps forever, with its own release clause dead** — a permanent lock with no way out. Nobody had hit it because nobody had cast it.

That is the severity ceiling of this bug class: a stale simplification means a card in the catalog is *quietly wrong*, and unlike an absent card nobody is expecting it to be.

## Thought Vessel — a derived getter, not a stored value
`Spec.NoMaxHandSize` declares the clause; `Game.EffectiveMaxHandSizeLocked(p)` walks the battlefield for a permanent the player controls with that bit; `populateDiscardPendingLocked` asks it instead of reading `Player.MaxHandSize`.

Deliberately **not** a `Static`: `StaticAbility.Apply` takes a `*Characteristic` and a target `*Card`, so there is **no seat in the signature**. A player-scoped effect has no characteristic to modify and no CR 613 layer to sit in, and growing a parallel player-layer pipeline for one clause wasn't worth it.

**Deriving makes leave/overlap correct for free.** "Set on enter, restore on leave" has to answer *restore to what?*, and gets two real cases wrong: two Vessels, where the first to leave restores the cap while the second is still out; and a player whose max was changed by something else, whose value the restore clobbers. There is no stored value to strand — and no new game state, so `clone.go` and undo are untouched, and `set_max_hand_size` still composes.

`PlayerView.MaxHandSize` now reports the **effective** cap, so the client badge stops saying "10 / 7" for a player cleanup will never prompt.

Failing test first: `TestThoughtVesselNoMaximumHandSize` failed with `map[2e9181e5…:3]` — the controller told to discard 3, exactly the report.

## Fixed — blocker had already shipped
| Card | Stale claim | What unblocked it |
|---|---|---|
| **Stasis** | "upkeep-sacrifice half lands with S19" | S19 triggers + `PayUnless` — **no off switch** |
[...]

Issue #338, opened 2026-09-11
[in-app] Thought Vessel Maximum Hand Size not recognized


krakenhavoc:
I was forced to discard at my end step

---

<details>
<summary>Client log — last 200 entries before the report</summary>

```text
16:01:08.257  received  snapshot seq=189 turn=5 seat=0 step=upkeep
16:01:08.258  sent      action pass_priority id=e746e80d
16:01:08.336  received  snapshot seq=190 turn=5 seat=0 step=upkeep
16:01:08.392  received  snapshot seq=191 turn=5 seat=0 step=draw
16:01:08.395  sent      action pass_priority id=32776fd0
16:01:08.469  received  snapshot seq=192 turn=5 seat=0 step=draw
16:01:08.511  received  snapshot seq=193 turn=5 seat=0 step=precombat_main
16:01:24.986  sent      action pass_priority id=65713625
16:01:25.060  received  snapshot seq=194 turn=5 seat=0 step=precombat_main
16:01:34.770  received  snapshot seq=195 turn=5 seat=0 step=begin_combat
16:01:34.771  sent      action pass_priority id=814b6447
16:01:34.845  received  snapshot seq=196 turn=5 seat=0 step=begin_combat
16:01:34.894  received  snapshot seq=197 turn=5 seat=0 step=declare_attackers
16:01:39.272  sent      action pass_priority id=58f895a5
16:01:39.345  received  snapshot seq=198 turn=5 seat=0 step=declare_attackers
16:01:54.380  received  snapshot seq=199 turn=5 seat=0 step=declare_blockers
16:01:55.764  sent      action pass_priority id=20454746
16:01:55.836  received  snapshot seq=200 turn=5 seat=0 step=declare_blockers
16:02:06.989  sent      action cast_spell id=354bafc8
16:02:07.049  error     server error code=bad_request message=you do not hold priority
16:02:13.996  sent      action cast_spell id=48a3a888
16:02:14.058  error     server error code=bad_request message=you do not hold priority
16:02:16.895  received  snapshot seq=201 turn=5 seat=0 step=declare_blockers
16:02:17.028  sent      action cast_spell id=657e9c3a
16:02:17.097  error     server error code=bad_request message=you do not hold priority
16:02:27.268  received  snapshot seq=202 turn=5 seat=0 step=declare_blockers
16:02:29.900  sent      action cast_spell id=3e30461b
[...]

Issue #293, opened 2026-09-11
Card catalog coverage: the next 2000 Commander cards by play rate


krakenhavoc:
Card coverage is the single largest lever on "is this playable yet",
and until now it has been chosen by hand — a deck triage here, a staples
list there. This issue replaces the guessing with a ranking: **the next
2000 Commander cards the catalog does not have, ordered by how often
Commander decks actually play them**, split into 20 batches of 100 and
triaged by the machinery each card needs.

It is the whole-format sequel to
[`docs/decklists/top-100-commander-staples.md`](../blob/main/docs/decklists/top-100-commander-staples.md),
which did the same thing for the first 100. Summary doc for this one:
`docs/decklists/card-coverage-roadmap.md` (branch `docs/card-coverage-roadmap`).

## The audited catalog count

**288 cards are registered** on `main` at `f26c961`.

Not 289, and emphatically not "however many `Register(Spec{` sites grep
finds". Two things bite here, and both have produced a wrong number in
the last week:

1. **Cards register from tables inside `for` loops.** `temples.go`,
   `check_lands.go`, `battle_lands.go`, `bond_lands.go`, `shocklands.go`,
   `painlands.go`, `talismans.go` and the fetchland cycle each register a
   whole cycle from one literal `Register` site. Grepping the literal
   undercounts by dozens. The honest number is `len(effects.All())`.
2. **`len(effects.All())` measured from a test reports 289**, because
   `flicker_test.go`'s `init()` registers a non-card probe spec keyed
   `test-flicker-etb-probe`. It is test-binary-only and correct to exist;
   it just is not a card. Subtract it. Every other registered spec
   resolves to a real, Commander-legal Scryfall card — 288 for 288, no
   tokens, no helpers, no strays.

### Method (reproducible)

```bash
# 1. the true registered count + the oracle-id set
#    (throwaway test, deleted before commit)
cat > server/internal/cards/effects/zz_probe_test.go <<'EOF'
package effects
import ("os";"sort";"testing")
func TestZZProbeCatalog(t *testing.T) {
    all := All()
    lines := make([]string, 0, len(all))
    for _, s := range all { lines = append(lines, s.OracleID+"\t"+s.Name) }
    sort.Strings(lines)
    f, _ := os.Create("/w/catalog_probe.txt"); defer f.Close()
    for _, l := range lines { f.WriteString(l + "\n") }
    t.Logf("REGISTERED_COUNT=%d", len(all))
}
EOF
[...]

PR #412, opened 2026-09-11
feat(server): publish the card catalogue, with completeness declared per card

### LINKED ISSUE #410: Publish the card catalogue, and make "is this card fully implemented?" a field rather than a comment
Two things that turn out to be the same problem.

## 1. We want a public page listing every card the engine plays

A browsable showcase of the catalog — searchable, filterable, with card images — so a player can answer "does it play my deck?" without importing one.

## 2. That page cannot be built on the signal we have

The page has to say which cards work **in their entirety**, and not every registered card does. Since S14 the catalog has carried its gaps in prose (AGENTS.md §7 "declared simplifications"). Surveying all 381 card files, that convention cannot be read by a program:

- **Free-form wording.** "No simplification." / "No simplifications." / "No other simplification." / "No simplification remains.", and six-plus ways of announcing a gap.
- **Silence is ambiguous, and it is the majority.** 220 of 505 registry entries sit in files that never mention the subject. Mind Stone and Blaze are complete implementations whose authors saw nothing worth saying; nothing distinguishes them from a card nobody checked.
- **Notes go stale in both directions.** Of the 87 files that *do* mention a simplification, **24 describe a gap that has since been closed**. A parser keyed on "this file says simplification" marks all 24 as broken.

A page built on that would be wrong in both directions — and listing a partially-implemented card as "working" is worse than having no page, because it makes a promise the engine doesn't keep.

## This is the follow-up #350 asked for

#350's closing section, verbatim:

> Every one of these is **a comment written in the present tense about a future sprint, with nothing that fails when that sprint lands.** A registry-level assertion tying "declares a simplification" to a test would catch the next batch automatically. Worth its own issue.

This is that issue. #338 was the same bug class reaching a player (Thought Vessel), and #350 found Stasis shipping a table-lock with its own release clause dead.

## Proposed

- `Spec.Completeness` (`Unreviewed` / `Full` / `Caveats`) + `Spec.Caveats []string`, with **the zero value meaning Unreviewed**. A spec that forgets the field, or predates it, publishes as unaudited rather than as working.
- `Register` panics on `Caveats` without text and on text without `Caveats` — the two mistakes a machine can catch, caught at boot.
- Deliberately **no** panic on an undeclared spec: cards land thirty at a time from several branches, and a hard gate just teaches reviewers to stamp `Full` for a green build.
- Backfill from what humans actually decided, not from inference. Leave the rest Unreviewed and let the page be the worklist.
- `GET /catalog` (public, no session) + a client route, with the three buckets visible and filterable.

Relates to #293 (catalog coverage). Closes the loop on #338 / #350.

krakenhavoc:
## Summary

- **A public card catalogue at `#/catalog`** — every card the engine automates, searchable and filterable by colour identity, type and completeness, with card images and a live count. No session, no dev gate; it ships in production.
- **`Spec.Completeness` + `Spec.Caveats`** replace comment-parsing as the answer to "does this card work in its entirety?". The zero value means **Unreviewed**, not "working".
- **Backfilled the 504 registered cards** from what a human actually decided: **141 Full, 144 Caveats, 219 Unreviewed**. Nothing inferred.

## Sprint

## Issues

Closes #410. Follow-up to #338 / #350 — this is the "registry-level assertion" #350's closing section asked for. Relates to #293.

## The judgement call, stated plainly

The task was "cards that should be working in their entirety". **The catalog could not distinguish that from "registered" without new metadata.** Parsing the existing prose convention (AGENTS.md §7) was surveyed across all 381 card files and rejected on evidence:

| Problem | Evidence |
| --- | --- |
| Free-form wording | "No simplification." / "No simplifications." / "No other simplification." / "No simplification remains.", plus 6+ ways of announcing a gap |
| **Silence is ambiguous, and is the majority** | 219 of 504 entries sit in files that never mention the subject. Mind Stone and Blaze are *complete* implementations whose authors saw nothing worth saying — indistinguishable from a card nobody checked |
| **Notes go stale in both directions** | Of the 87 files that *do* announce a simplification, **24 describe a gap that has since been closed** (Awakening Zone, Birds of Paradise, Wash Away …). A parser marks all 24 as broken |

A page built on that would be wrong in both directions — worse than no page, because it promises things the engine does not do.

**So: a field, with three states.** "Nobody has checked this" is a real answer and gets its own bucket on the wire and on the page. `Register` panics on `CompletenessCaveats` with no `Caveats`, and on `Caveats` under any other value — the two mistakes a machine can catch, at boot.

**`Register` deliberately does *not* reject an undeclared spec.** Seven branches are adding cards right now; a hard gate makes every batch a merge-blocking chore and the predictable outcome is a reviewer stamping `CompletenessFull` for a green build — the one result the field exists to prevent. The lazy path is safe and visible rather than fast and wrong. This is argued at length in [ADR 0042](docs/decisions/0042-card-catalog-page.md).

### Where the 504 came from

| Bucket | Entries | Source |
| --- | ---: | --- |
| `Full` | 141 | 84 files plainly declaring "No simplification", plus the 24 reviewed files found to have no current gap |
| `Caveats` | 144 | 63 reviewed files, each read **against its code** and given one player-facing sentence, plus 3 cycle/helper files |
| `Unreviewed` | 219 | everything else — left at the zero value |

The 219 are not a defect introduced here; they are the catalog's pre-existing state, now visible instead of implicit. The page is the worklist.

## Found while building it

- **60 modal double-faced cards register only their land back** (`<oracle_id>#1`); nothing automates the creature front. No card file can declare that gap because no card file *owns* that face, so the catalog derives it — a card with an unautomated printed face gets an automatic caveat naming it. Invisible before this page existed.
- **`isPlayablePrint` was missing the `"Card"` / `"Card // Card"` placeholder type lines** that AGENTS.md §7 warns about; the layout and set_type switches did not catch all of them. Fixed, which also improves `byName` deck-import resolution.

## Shape

- `server/internal/catalog` — its own package: the join needs both `internal/cards` and `internal/cards/effects`, and effects is otherwise blank-imported exactly once so nothing depends on it.
- `cards.Index` gains `byOracle` + `FindByOracleID`. The catalog is keyed by `oracle_id` and there was no such lookup.
- **`GET /catalog/image/{id}` is public; `/cards/{id}/image` stays session-gated.** That gate is worth keeping — unauthenticated it proxies any of ~35k Scryfall UUIDs. The catalogue route accepts only the representative printing of a registered card: a few hundred IDs, fixed at build time, already published by `GET /catalog`. Same `ImageCache`, so one disk cache and one SSRF-guarded fetch between them. Scope check runs *before* the availability check so an out-of-scope ID learns nothing.

## The three-list trap

`/catalog` added to all of them in this PR, as `deploy/Caddyfile`'s comment demands:

- [x] `@api path` in `deploy/Caddyfile`
- [x] `server.proxy` in `client/vite.config.ts`
- [x] `API_PATH` in `client/src/sw/service-worker.js` (+ its test)

**PWA not regressed:** catalogue art joins the existing `card-image` cache (same immutability), the JSON stays uncached, precache list unchanged, `npm run build` emits the same `sw.js` shape. Routing is hash-only so no `navigateFallback` concerns.

## Test plan

- [x] `gofmt -l .` clean, `go vet ./...` clean, `go test ./...` green (19 packages, 0 FAIL)
- [x] `npm run check` — 0 errors, 0 warnings; `npm run lint` clean; `npx vitest run` — 341 tests, 24 files
- [x] `npm run build` — SW plugin emits, precache list unchanged
- [x] Response size measured, not guessed: **228 KB raw, 53 KB after Caddy gzip** for all 504 cards, with a 5-minute browser cache
- [x] **13 new Go tests** in `internal/catalog`: an undeclared spec publishes as unreviewed and is never counted Full; buckets partition the catalog; face-only registration collapses to one card *with* a caveat; completeness merges pessimistically; missing printings published and counted; type-line parsing; `#N` key splitting; **the route answers with no credential**; the image route 404s a non-catalog card
- [x] **7 new Go tests** in `effects`: zero value is Unreviewed; wire strings; `Register` rejects caveats-without-text, text-without-caveats, empty caveat; accepts an undeclared spec; **every shipped caveat is a player-facing sentence with no engine jargon**
- [x] **14 new client tests** in `catalog.test.ts`: colour filtering is on *identity* (a Signet is not colourless); OR within a facet, AND across; unreviewed stays its own bucket; face-name search; image URL omits `face=0` so it shares existing cache entries
- [x] **End-to-end against the real 629MB Scryfall dump**: 504 cards, **0 missing printings**; `/catalog/image/…` → 200 image/jpeg with no cookie; `/cards/…/image` → **401** (unchanged); `/catalog/image/<Black Lotus>` → **404 not a catalog card**; `/games` → **401**; and the same JSON through the Vite dev proxy (proving the proxy entry, not index.html)
- [ ] Visual check in a browser — no browser automation in this environment, so the page's layout is unverified visually. Worth a look before merge.

## Notes for review

1. **Worth a second opinion: is a *public* page of card art consistent with AGENTS.md §1/§8's "private, personal use" posture?** The page serves WotC art from our own origin to anonymous visitors. I scoped it as tightly as I could (catalogue cards only, reusing the existing cache) and it exposes no game state, but the public-ness is a product call, not an engineering one. Easy to gate behind a session later — one line in `main.go` and one in `App.svelte`.
2. The 144 caveat sentences were written by reading each file against its code. They are the main thing to spot-check; a few are judgement calls about how much detail a player wants.
3. The card-file diff is 173 single-hunk edits (one `Completeness:` line each, plus gofmt realignment inside the same literal). No card behaviour changes. Additive field appended to the end of `Spec`, so branches adding cards merge cleanly and their new cards default to `Unreviewed`.
4. **ADR number:** 0042, checked against `docs/decisions/` on **all 46 remote branches** (`git ls-tree` per branch, re-run immediately before opening this PR). Highest in use was 0041; 0005/0024/0029/0030 remain permanently unused. At least one sibling agent is picking a number concurrently — if 0042 lands elsewhere first, I'll renumber this file rather than ask them to move.


krakenhavoc:
CI is red here on `Server — vet`, and it is **not** this branch.

`origin/main` currently does not compile — `GoldToken` is declared twice in `server/internal/cards/effects/tokens.go` (lines 165 and 364), from two concurrent card PRs that each added the same helper to different regions of the same file. GitHub builds every PR merged with main, so all 11 open PRs are failing on it. Filed as #416 with the one-line fix.

This branch does not touch `tokens.go`. Verified green in isolation against its own base:

- `gofmt -l .` clean, `go vet ./...` clean, `go test ./...` — 0 FAIL
- `npm run check` 0 errors / 0 warnings, `npm run lint` clean, `npx vitest run` — 341 passed
- `npm run build` clean, service-worker precache list unchanged

Happy to rebase once #416 lands.

Issue #324, opened 2026-09-11
[in-app] Warp is not working


krakenhavoc:
when casting anticasual vestige I should be prompted for warp or full cost and then if warped pay the lower amount and then be able to cast from exile later.

---

**Screenshots**

![att-1.png](https://cmd.labxp.io/bugreport/att/2f1fe42e-e298-4ee2-bb10-03a3dc5e3848/att-1.png)

<sub>Uploaded by the reporter. Served from the game server; expires after 90 days.</sub>

---

<details>
<summary>Client log — last 45 entries before the report</summary>

```text
15:11:54.351  info      connected to wss://cmd.labxp.io/ws?game=422b603e-131f-447b-86bf-6ffba7e22e81&token=REDACTED&player=de5f7f80-eab0-4e6e-a0ee-1765de62fc7b
15:11:54.358  received  snapshot seq=225 turn=5 seat=0 step=precombat_main
15:11:55.792  sent      action pass_priority id=a6af1fdb
15:11:55.867  received  snapshot seq=226 turn=5 seat=0 step=begin_combat
15:11:56.712  received  snapshot seq=227 turn=5 seat=0 step=begin_combat
15:11:56.712  sent      action pass_priority id=58a6a42a
15:11:56.784  received  snapshot seq=228 turn=5 seat=0 step=declare_attackers
15:11:59.845  received  snapshot seq=229 turn=5 seat=0 step=declare_attackers
15:12:00.881  sent      action pass_priority id=6ce23883
15:12:00.953  received  snapshot seq=230 turn=5 seat=0 step=declare_blockers
15:12:00.997  received  snapshot seq=231 turn=5 seat=0 step=declare_blockers
15:12:00.998  sent      action pass_priority id=2245a975
15:12:01.071  received  snapshot seq=232 turn=5 seat=0 step=combat_damage
15:12:01.112  received  snapshot seq=233 turn=5 seat=0 step=combat_damage
15:12:01.112  sent      action pass_priority id=54d7b6c8
15:12:01.189  received  snapshot seq=234 turn=5 seat=0 step=end_combat
15:12:01.229  received  snapshot seq=235 turn=5 seat=0 step=end_combat
15:12:01.229  sent      action pass_priority id=0b3385cd
15:12:01.305  received  snapshot seq=236 turn=5 seat=0 step=postcombat_main
15:12:01.345  received  snapshot seq=237 turn=5 seat=0 step=postcombat_main
[...]

PR #413, opened 2026-09-11
feat(cards): warp — a discount now, the real card later (refs #94, #324)

krakenhavoc:
## Summary
- Warp (CR 702.183) — the assessment in #94's brief was right: it is ~90% composable from machinery that already shipped, and the missing 10% is one field.
- `Warp("{R}")` bundles the discount with `WarpExile`, which schedules a CR 603.7 delayed trigger to exile the permanent at the next end step and leaves an `ExilePlayPermission` behind for the later cast.
- `ExilePlayPermission.NotBeforeTurn` is the piece ADR 0037 had already identified as missing: the grant had a ceiling (`UntilTurn`) but no floor.
- **Weftstalker Ardent** carries it. That card was the single exception ADR 0037 §5 named — a Spec that shipped knowingly omitting a printed clause — and the exception list is empty again.

**Stacked on #409 and #411.** Nothing here depends on `CastableZones`; warp is paid from hand and opens no new cast source. It is stacked only to keep the S29 branches linear and avoid a conflict in `game/alternative_cost.go`.

## Notes

**Three shipped mechanisms in a trench coat.** The discount is an `AlternativeCost` (#257). The take-back is a delayed trigger (#255), scheduled through the same entry seam evoke's sacrifice-on-entry uses. The later cast is an ordinary cast from exile under the `ExilePlayPermission` impulse exile introduced (#269), unbounded the way airbend's is — so the client's existing exile button renders it with **no new client code** beyond the turn floor.

**Why `NotBeforeTurn` exists, and why it is checked first.** An unbounded grant is live the instant it is stamped, and it is stamped *during the very end step that exiled the creature*. Without a floor, a warped creature with flash could be recast the moment it left. The floor is checked ahead of `WhileExiled` because "for as long as it remains exiled" widens the *ceiling* and must not be read as removing the floor. It is computed at **schedule** time rather than at fire time — both readings agree for every printed warp card (creatures cast at sorcery speed on their controller's turn), but the schedule-time reading is the one that survives a warped creature with flash.

**#324 is not closed by this.** The bug reports **Anticausal Vestige**, whose other half — "put a permanent card with mana value less than or equal to the number of lands you control from your hand onto the battlefield tapped" — needs a hand-selection `PendingChoice` kind that does not exist (there is no "put a card from your hand onto the battlefield" prompt, server or client). Under ADR 0037's whole-text rule the card still does not join the catalog. The ADR's row is updated to record which half is now done, so the next person doesn't re-derive it.

## Sprint
S29 — Alt-cast paths from non-hand zones

## Issues
Relates to #94, relates to #324. Depends on #409 and #411.

## Test plan
- [x] `gofmt -l .`, `go vet ./...`, `go test ./...` — full server suite green
- [x] `npm run check` — 0 errors, 0 warnings
- [x] `npx vitest run` — 333 tests green
- [x] New: `server/internal/cards/effects/warp_test.go` — 6 cases against the real catalog entry. The warp cost is charged and the printed cost is not; the creature really enters and is **exiled** (not sacrificed, not destroyed) at the next end step; the grant names the warping player, is unbounded, and is dark on the warp turn and live on the next; the creature is recast out of exile for its **printed** `{2}{R}` on a later turn; and a printed-cost cast of the same card schedules nothing and stays on the battlefield.
- [x] Extended: `client/src/lib/impulseExile.test.ts` — the button is withheld before `not_before_turn` and offered from it onwards, and floorless grants (impulse exile, airbend) are unaffected.


krakenhavoc:
#409 has landed, so the dependency list in the description is down to #411 — merge that first and this becomes a warp-only diff.

Also carries the merge that picks up #417's `GoldToken` fix, since `origin/main` was briefly unbuildable while this was in flight. Merge commits rather than a rebase because force-pushing is off the table.

Full server suite, `svelte-check` and vitest all green locally against the merged tree.

Issue #89, opened 2026-04-19
S31 — AI bot seat (heuristic policy)


krakenhavoc:
**Phase:** 8 (first AI sprint) · **Goal:** fill an empty Commander seat with a bot good enough for solo practice and 1–3-friend games. After S27–S30 close the engine gaps (card-type completeness, cost modification, alt-cast paths, damage/cloning/face-down/protection), the engine is Arena-parity with ~600-card catalog and structured `Effect` descriptors on every card — for the first time in the project, a competent AI is actually buildable. S31 ships that competence as a tiered, swappable policy framework.

**Not in scope: tournament-strength AI.** Forge has 15+ years of rule-based heuristic work and still plays "dumb but playable"; XMage's MCTS takes minutes per turn; academic MCTS + RL work (Cowling-Ward-Powley 2012; MageZero AlphaZero-style) is multi-year research. This sprint's bar is Forge's: makes legal moves, makes locally-sensible decisions, doesn't deadlock, uses removal on threats.

## Design decisions

1. **Heuristic-first.** Rule-based evaluation + greedy action selection informed by the structured `Effect` descriptors the S14+ catalog already provides. No per-card scripts — the bot reasons over effect types.
2. **Bot = virtual seat.** A bot is a `Player` with no WS connection; a goroutine subscribes to the same game stream, runs through the same visibility filter as a human client, and emits actions via the existing action protocol. Zero protocol forking.
3. **Swappable `Policy` interface.** `Policy interface { DecideAction(view *ViewOfGame, legal []Action) Action }`. S31 ships `HeuristicPolicy`; future sprints plug in MCTS or LLM-backed policies — contract stays the same.
4. **4-player-first targeting.** Multi-opponent Commander shapes every targeting decision: threat ranking across 3 opponents, aggression rotation, concede heuristic.
5. **Safety rails over cleverness.** Loop detection, illegal-action filter, unknown-card graceful pass.

### Bot infrastructure (`server/internal/bot/`)
- [ ] New package with `Policy interface`, `RandomPolicy` (baseline), `HeuristicPolicy` (deliverable)
- [ ] `bot.Seat{PlayerID, Policy, Difficulty, MinThinkMs}` — virtual-seat abstraction
- [ ] Goroutine-per-bot: subscribes to `ws.Room` state deltas using `FilterViewFor(playerID)`; emits `Action`s via `actions.Dispatch`
[...]

PR #431, opened 2026-09-11
feat(server): S31 sub-PR 0 — the public game log (relates to #89)

krakenhavoc:
## Summary

- **`GameView.log`** — the public game log ([ADR 0033](https://github.com/krakenhavoc/cmd_and_ctrl/blob/main/docs/decisions/0033-ai-bot-seat.md) §4): the last 200 table-visible events, oldest first, through the same visibility filter as every other view field. A **projection** of `game.Game.Events`, not a stored buffer — so nothing new lands on `game.Game`.
- **Emits on** step boundaries, casts, resolutions, fizzles, counters, zone changes, sacrifices, draws, life changes, damage, attack and block declarations, tokens, eliminations. Two additive engine event kinds carry the two that had no event at all: `step_began` and `block`.
- **Client panel** — a right-side drawer grouping entries under their step, tail-following, with a "mine" seat filter.
- **bugstore** — `gamelog.txt` alongside the screenshots and the pinned replay, served at `GET /bugreport/{id}/gamelog` behind admin auth.

## Sprint

S31 — AI bot seat (heuristic policy), sub-PR 0

## Issues

Relates to #89

## Why this is the first thing in S31

A policy handed only a snapshot cannot see that the seat to its left wiped the board last turn, that this creature has attacked it twice, or that an opponent held up two blue mana and did nothing. Those are not refinements — they are most of what a Commander player reasons about. The log is what makes a view *temporal* rather than instantaneous, and because `Input.View` is byte-identical to a human's, it is worth exactly as much to the human.

## Visibility — the part that matters

The log is a cheating vector if it is wrong once, so it is enforced twice, because the two leaks are different shapes.

**At build time.** An event whose card sits in a hidden zone at *both* ends — library → hand, hand → library — is emitted with **no card reference at all, not even the instance ID**. The ID alone would be the leak: it lets a client follow a tutored card and recognise it when it surfaces on the battlefield three turns later. A draw likewise never names the card, for anyone, including the drawer. `TestPublicLogCollapsesDrawsAndNeverNamesThem` asserts the strong form — the hidden card's UUID appears nowhere in the serialised log, for any of the three viewer kinds.

**At filter time.** Every surviving card reference goes through the **same `isKnower` closure** `FilterViewFor` already runs over every `CardView` — not a parallel model, literally the same predicate — and the rendered `text` is regenerated from the redacted inputs. A face-down permanent's entry reads "a card entered the battlefield" for everyone but its controller, and "Sea Gate Restoration entered the battlefield" for the controller. `knowers` are cleared on the filtered output exactly as `redactCardForViewer` clears them, so re-filtering a stored replay frame redacts rather than re-reveals (`TestPublicLogRedactionIsIdempotent`).

Names and knower sets are resolved **out of the view that was just assembled**, not out of `game.Game`. Resolving them anywhere else would have been a second visibility model to keep in sync with `FilterViewFor`'s, and the two would eventually disagree.

## Wire cost

`TestPublicLogWireCost`, saturated four-player board (four seats, 48 permanents, 32 graveyard cards, 40 turns of history):

```
four-player frame: 35555 B without the log, 63911 B with a full 200-entry log
                   (+28356 B, +79.8%; 141 B/entry)
deflated:           5174 B without,        7565 B with  (+2391 B)
```

**+28.4 KiB uncompressed, +2.3 KiB deflated, 141 B/entry.** Call it +28 KiB per frame per viewer, or ~113 KiB per broadcast on a four-player table.

The schema is shaped by that number, not the other way round:

- **players are seat indices, not UUIDs** — the convention `TurnView.ActiveSeat` already uses; a uuid string is 36 bytes and this struct repeats 200 times per frame;
- **`step` rides only on `step` entries**, everything after one inheriting it;
- nothing is present that the entry's kind already implies (a cast's destination is always the stack; a resolved spell's trip to the graveyard is not a second fact).

The test **gates on bytes-per-entry (≤160 B) and total log bytes (≤32 KiB)**, and deliberately *not* on a share of the frame: the synthetic cards in the fixture carry no oracle text, faces or ability lists, so the denominator is smaller than a real board's and a ratio would flatter the log. The two absolutes are what this package controls.

Projection cost, isolated (`BenchmarkPublicLogProjection`, 1080 events): **229 µs/frame**, once per broadcast rather than once per viewer — `ws.Room` builds the view once and filters it per client. That is ~15% of what `ViewOfGame` already costs on the same board.

Two levers are deliberately left on the table for sub-PR 2's budget conversation: the hub does not enable `permessage-deflate` (the numbers above say what it would be worth), and the real fix — shipping only the entries a client has not seen — needs a delta frame the v0 protocol does not have.

## Snapshot drift classification

**No new `Card` or `Game` field, so nothing new to classify — and that is the argument, not a dodge.**

The log is a pure projection of `game.Game.Events`, which `snapshot_drift_test.go` already classifies **`carried`**. So the history is carried across an undo, a snapshot restore, and a deploy for free, and there is exactly one copy of the facts. A `Game.Log` ring buffer would have been a second copy of data the engine already keeps — a thing to classify, a thing to clone, and a thing to drift out of sync with `Events` the first time someone emitted an event without going through it.

The persistence claim is executable rather than asserted: `TestPublicLogSurvivesSnapshotRestore` captures a snapshot, restores it, and requires the projected log to come back entry-for-entry with the same seqs and the same text.

The cost of the choice is honest: the projection walks the whole event log on every broadcast rather than reading a ready-made tail. The benchmark above is what that costs.

## Engine changes (both additive)

- **`EventStepBegan`** — emitted from `runStepEntryHooksLocked` *after* the S17 skip-step replacement window, so a step Stasis cancelled never announces. Distinct from `EventBeginUpkeep` / `EventBeginEndStep`, which exist for the trigger harvester and cover the two steps cards actually name. Nothing in the rules engine listens for it.
- **`EventBlock`** — the other half of `EventAttack`, emitted from `DeclareBlocker` only on a *new* pairing, so re-pointing a blocker in the sandbox does not announce twice (the same line `EventAttack` draws).

## bugstore

`Report.PinGameLog` writes `gamelog.txt` next to `replay.jsonl`, recorded in the manifest and served at `GET /bugreport/{id}/gamelog` behind admin auth. The issue body **references it and never inlines it**, holding ADR 0017 §4's line that game state stays out of issue bodies. Entitlement to *render* it mirrors `bugReplaySource` exactly: admins anywhere, everyone else only for the game their session is bound to.

The pin is the **unfiltered** projection. An operator triaging "the game let them cast that twice" needs what the server believed, not what one seat was shown — which is also why it is admin-gated even though the contents are public by construction.

Where the replay says what the state *was* at every frame, the log says what *happened*. That is usually the question the report is asking.

## Test plan

- [x] `gofmt -l .` clean, `go vet ./...` clean, `go test -race ./...` green (the whole suite, in the devcontainer image)
- [x] `npm run check` — 0 errors, 0 warnings; `npx vitest run` — 25 files, 347 tests; `npm run lint` clean; `vite build` succeeds
- [x] `TestPublicLogCollapsesDrawsAndNeverNamesThem` — a hidden card's instance ID appears nowhere in the serialised log for spectator, owner, or opponent; consecutive draws collapse to one counted line
- [x] `TestPublicLogRedactsCardsTheViewerDoesNotKnow` — the same entry, three viewers: named for the knower and the admin, "a card" for the opponent, with the name absent from the opponent's serialised log and the instance ID still present (it is already on their wire from the battlefield)
- [x] `TestPublicLogRedactionIsIdempotent` — re-filtering an already-filtered view for a different seat does not re-reveal
- [x] `TestPublicLogIsBounded` — 800 events in, 200 out, tail kept, seq-ordered
- [x] `TestPublicLogSuppressesRedundantEntries` — a sacrifice is not also a zone move; a resolved spell is not also a trip to the graveyard
- [x] `TestPublicLogSurvivesSnapshotRestore` — the drift classification, executable
- [x] `TestPublicLogWireCost` — measures and gates; numbers above
- [x] `TestBugReportPinsGameLogAndServesItToAdminsOnly` / `TestBugReportDoesNotPinAnotherGamesLog` — 200 for admin, 403 for the seated reporter, 401 unauthenticated, and a session in game A cannot make the server render game B's log
- [x] `bugstore`: round-trip, manifest entry, empty and oversized refusals leave no file, path-traversal ids rejected
- [x] `client/src/lib/gameLog.test.ts` — 10 tests over the grouping, the seat filter, and the tone map

## Not done

- **No live screenshot of the panel.** `svelte-check`, the unit tests and `vite build` all pass, but I did not stand up a server with a seeded four-player game to look at the drawer. Worth a glance before merge.
- The `mine` filter and the drawer's open state are tab-local, not in S11.5 settings. Deliberate — whether you have the log open is not table state — but it does mean it resets on reload.


krakenhavoc:
Frame-budget coordination from sub-PR 2 (#429) — we're both adding to the same `GameView`, and the sprint note warns against each of us assuming the other's headroom, so here are my numbers rather than a guess.

**Measured on a full four-player board**, active seat holding priority in its main phase, seven-card hands, developed boards (`TestLegalMovesFrameBudget` in `server/internal/protocol/legal_moves_test.go`, logged on every `-v` run):

| board | frame before | frame after | `legal_moves` |
|---|---|---|---|
| 4 seats × 3 creatures | 25,367 B | 39,326 B | **13,959 B (+55.0%)** |
| 4 seats × 6 creatures | 30,143 B | 44,102 B | **13,959 B (+46.3%)** |
| 4 seats × 10 creatures | 36,511 B | 50,463 B | **13,952 B (+38.2%)** |
| idle seat (owes nothing) | 30,143 B | 30,143 B | **0 B** |

So: **a four-player frame is ~25–37 KB today, and `legal_moves` adds ~14 KB — but only on frames where the viewer actually owes a decision.** Zero on every other frame, including every frame an opponent receives during your turn.

Two things that might matter to you:

1. **I gated at 24 KiB, not at "whatever fits".** That's roughly a quarter of a four-player frame, deliberately leaving the log the same order of room. `TestLegalMovesFrameBudget` fails rather than letting the field creep, and the constant's comment says explicitly that raising it is the wrong fix.

2. **If the log needs more room, take it from me — there's slack that costs nothing.** `legalMovesWireCap` (`server/internal/protocol/view.go`) caps the list at 48 moves and degrades past that by keeping one move per `(source, kind)`. Dropping it to ~24 roughly halves the field, and **costs the client nothing it reads today**: `timing.ts` only ever asks "does this card have a move?", and targeting still runs off `CardView.legal_targets`. The full target expansion on the wire is speculative headroom for a future target picker, not a live consumer. Say the word and I'll lower it, or lower it yourself in your PR — I'd rather that than either of us discovering the ceiling in a playtest.

One incidental note in case it saves you time: the unfiltered `GameView` is what `captureLocked` marshals for the crash dump **and the replay log**, so anything you put on it exported is paid twice per `Apply` and lands in every replay file. I kept the per-seat enumeration in an unexported field for the hidden-information reason, and the free side effect is that the replay log doesn't carry it. A bounded log is public by nature so that may not apply to you, but the replay-size arithmetic is worth doing once.

(No action needed on your side; this is just the number, on the record, so neither of us assumes the other's headroom.)

Issue #499, opened 2026-09-14
internal/legal enumerates nothing for four live PendingChoice kinds — a seat owing one gets an empty move list


krakenhavoc:
`internal/legal`'s `choiceMoves` handles 16 of the 20 live `PendingChoiceKind` values. Four have no `case` and the `switch` has no `default`:

- `PendingChoiceMayCast` (`may_cast`) — `game/pending_choice.go`, raised by the S28 cascade path
- `PendingChoiceLegendRule` (`legend_rule`) — `game/legend_rule.go`
- `PendingChoiceCreatureType` (`choose_creature_type`) — `game/creature_type_choice.go` (S26)
- `PendingChoiceChooseProtector` (`choose_protector`) — `game/battle.go`

This is not a cosmetic gap, because of how the enumerator short-circuits. `choiceMoves` sets `owed = true` for **any** pending choice owned by the seat, and `enumerateLocked` returns immediately when it does:

```go
// Pending choices come first and, when one is owed by this seat,
// they are the ONLY moves ...
if e.choiceMoves() {
	return e.out
}
```

So a seat owing one of those four kinds is enumerated an **empty move list** — not the choice's answers, and not `pass_priority` either.

Consequences:

- **Bot seats.** `Runner.step` treats `len(moves) == 0` as "nothing to do, go back to sleep" (`aiseat/runner.go`). A bot owed a legend-rule or cascade choice sleeps until something else wakes the room, and nothing else will, because the engine refuses `pass_priority` while a choice is open. That is a stalled table, and the existing rejection-based guard (`MaxConsecutiveRejects`) never fires because nothing is ever dispatched.
- **Human seats.** `GameView.legal_moves` is empty for the same window, so anything the client derives from it (post-#429 `timing.ts`) has no entry to look up.

Every deck ships a commander, so `legend_rule` is reachable in ordinary play the moment a token copy or a second printing hits the battlefield.

Suggested fix: add the four cases, and give the `switch` a `default` that logs rather than silently emitting nothing, so the next kind added to `game/` fails loudly instead of wedging a seat.

Found while reconciling `docs/sprints.md` § S31 sub-PR 1 against the code (docs-only PR; no code touched).

Relates to #89.

krakenhavoc:
Reproduced unattended, with a seed, by the catalog soak being added in #601 — four random bots on decks dealt out of the catalog itself:

```
[...]

Issue #478, opened 2026-09-13
A fetched permanent is stranded in the library when its entry queues a replacement prompt


irobinson010:
Found by the batch 16 card agent (#309, PR #477) with a probe test; reported rather than fixed because card batches stay out of `server/internal/game/`.

## What happens
The search path (`SearchLibraryThenForEffect`) — and, by the same shape, exile-return and reanimation — runs the CR 614 entry pipeline but is not `entryResumable`. When two entry replacements apply to the fetched permanent and the CR 616 ordering prompt is queued, the move is abandoned: **the card stays in the library both before and after the prompt is answered.** The library is never shuffled either.

## Repro
1. An opponent controls Kismet and Thalia, Heretic Cathar (two "enters tapped" replacements that apply to your lands).
2. Cast Circuitous Route (or activate any fetch) and find a Guildgate.
3. A `replacement_order` prompt appears. Answer it. The Guildgate is still in the library.

This is the declared gap in `AGENTS.md` §7 ("neither entry site is `entryResumable`") reaching a real game: with one replacement the fetched land enters tapped with no prompt; with two, it is lost.

## Related, same root
`CreateTokenForEffect` skips the zone-move replacement pipeline entirely, so no enters-tapped replacement ever sees a token; `thalia_heretic_cathar.go`'s comment claims tokens arrive tapped, which is not true on `main`.

## Where
`server/internal/game/effect_api.go` search / reanimation / exile-return entry sites, `entry_choice.go`, `replacements.go` (`applyReplacementsLocked` → `errReplacementPending`). The land-play path already has the resume; the fix is a resume frame that carries the search continuation (and the shuffle) — `TestFetchedShocklandEntersTappedWithNoPaymentOffered` pins the narrower declared gap and should flip with it.

Refs #263, #293, #309.

krakenhavoc:
**Verdict: still reproducible. Confirming for the 2026-09-14 triage sweep — and flagging that this, #359 and #529 are one family that should be fixed together.**

`origin/main` @ `f1bdc37`. Nothing has touched this since the report.

The behaviour is not merely current, it is **declared in the code**, at length. `server/internal/game/effect_api.go:1365-1387`, above `searchEnterBattlefieldLocked`:

> *"DECLARED SIMPLIFICATION — this event is NOT `entryResumable`, so a fetched permanent never gets an entry prompt."*

and, on why not:

[...]

PR #543, opened 2026-09-14
repro(aiseat): Myriad Landscape's fetch wedges a bot seat and the whole table (FAILING test, do not merge)

krakenhavoc:
**This PR is a reproduction. It is expected to FAIL CI, and it must not be merged as-is.** It adds one test and changes no production code. The failing subtest is `basics=6`; the other four subtests pass and bracket the condition.

Filed from a forensic review of a live two-bot table that hung on Myriad Landscape's fetch and hung again, on the same move, after a server restart. Root cause and recovery are written up in the companion issue.

## What fails

```
--- FAIL: TestMyriadLandscapeSearchWedgesABotSeat/basics=6
    step 1: engine rejected "Myriad Landscape — up to two basic lands that share a land type: take Island Forest": game: invalid parameter
    step 2: engine rejected ... (same move)
    step 3: engine rejected ... (same move)
    WEDGE at step 3: ... and no pass is on offer — runner.step's forced-pass
    rescue has nothing to force.
    choices=[kind=search_library reason="Myriad Landscape — up to two basic
    lands that share a land type" max=2 cands=8]
```

## The mechanism

Three pieces, each defensible on its own:

1. `effects/myriad_landscape.go` enforces "two basic lands that **share a land type**" through `SearchLibrarySpec.Validate`, because the constraint is a property of the pair and no per-card predicate can express it.
2. `legal/choices.go`'s `PendingChoiceSearchLibrary` branch enumerates `combinations(SearchCards, 1, SearchMax, MaxExpansionPerSource)`. It never consults `Validate` — the hook lives on the unexported `searchResume` frame — so it offers mismatched pairs as if they were legal moves.
3. `game/search_choice.go`'s `ResolveSearchLibrary` rejects such a pair with `ErrInvalidParam` and deliberately does **not** dequeue the choice, so a client "can try again rather than losing the prompt".

A human retries differently. `heuristic.decideChoice` is a deterministic argmax, and `valueOfChoice` sums `cardValue` over the picked cards, so a two-land pick always beats a one-land pick and ties go to the lowest index — the first pair in lexicographic order, every time.

There is no escape. `legal.EnumerateFor` returns **only** a choice's answers when a seat owes one (`legal.go:173-178`), so `aiseat.PassIndex` is `-1`, and `Runner.step`'s guard

```go
if rejects >= r.cfg.MaxConsecutiveRejects {
        if pi := PassIndex(moves); pi >= 0 { ... } else { return true }
}
```

falls into `return true` and sleeps. Rejected dispatches never broadcast, so the replay's frame count goes static and the table looks idle rather than looping.

It survives a restart because `GameSnapshot.Restorable()` is `Continuations.Empty()`: the prompt carries a `searchResume` continuation, so no restore point is written while it is open and the last good one has the activated ability still on the stack. Restore, re-resolve, re-wedge.

## Why it needs a thinned library

`MaxExpansionPerSource` is 12, and `combinations()` fills that budget with one-card picks before it ever reaches two-card picks. With a full Commander mana base the bot is only ever offered single basics, every one of which is legal — which is also a quiet play-quality bug: Myriad Landscape fetches one land, not two. Once the library is down to roughly a dozen matching basics — a ramp deck on turn 8+ — pairs start being offered and the wedge becomes reachable.

## Related

- Sibling of #499: same trap (a seat owing a choice with no way out), different cause. There the kind has no `case` and the list is empty; here the kind is handled and the list is non-empty but every answer the policy will pick is rejected. A `default`-less switch and a rescue that assumes a pass exists are the shared structure.
- **Not** #478 — that needs two entry replacements and a CR 616 ordering prompt. Here a single `SelfEntersTapped` applies and the fetched cards are plain basics; the entry is not where this stalls.
- Not the missing S31 loop guard either: the shipped reject guard *does* fire, it just has nothing to force.


krakenhavoc:
Companion issue with the full write-up, blast radius, candidate fixes, and the recovery path for a table that is already wedged: #544

Related: #499 (same trap, different cause — a seat owing a choice with no way out). Not #478.

Field context: #545, from the same replay review.

krakenhavoc:
Retuned to the field evidence from replay `aca910f0-63c3-4a0f-bef4-5a03aef44704` (the actual wedged table). Full confirmation write-up: #544 (comment).

Decks are now 99 cards with real basic counts, and the activation happens on turn 1-2 with a full library — matching the field state rather than a hypothetical thinned one:

```
simic-ramp_7F_5I   candidates=10 answers=13 two-card-picks=2   FAIL
6F_6I              candidates=10 answers=13 two-card-picks=2   FAIL
20F_20I            candidates=35 answers=13 two-card-picks=0   PASS
```

The header comment blaming a thinned late-game library was wrong and has been rewritten. The trigger is the deck's **basic count**, not the game state: `simic-ramp` ships 12 basics in 99 cards, so the candidate pool is under `MaxExpansionPerSource` from the opening hand and the wedge lands on the first activation of the game.

One thing the field data adds that changes the shape of the fix: with 11 candidates the enumerator offers **exactly one** two-card pick, `(pool[0], pool[1])`. Filtering invalid pairs out is necessary but leaves zero pairs — correct and unwedged, but Myriad Landscape then fetches one land instead of two. The budget walk needs to reach *valid* pairs, not merely skip invalid ones.

krakenhavoc:
Superseded by #546, which takes this reproduction over as its acceptance test.

`myriad_wedge_test.go` is carried across from `bb8bf02` unchanged except for one added assertion: `pairs == 0` now fails. Without it the `20F_20I` subtest passed while the bot quietly fetched a single land off a card that prints two — exactly the silent half of the bug the retune identified. The file header is rewritten to say what closed, not what is broken.

Confirmed red before the fix and green after, all three subtests:

```
simic-ramp_7F_5I   candidates=10 answers=13 two-card-picks=6   PASS
6F_6I              candidates=10 answers=13 two-card-picks=6   PASS
20F_20I            candidates=35 answers=13 two-card-picks=6   PASS
```

Two-card picks at every shape, including the one that used to offer none.

Please close this rather than merging it — #546 carries the test and the fix together. The retune was load-bearing: without the field evidence I would have shipped the `Validate` filter alone, which unwedges the table and leaves Myriad Landscape fetching one land forever.

krakenhavoc:
Superseded by #546, which took this reproduction over (red before / green after) and added the `pairs == 0` assertion this version was missing — without it the 20F_20I subtest passed while the bot quietly fetched one land.

Closing rather than merging, as this PR was always marked do-not-merge.

PR #553, opened 2026-09-14
feat(lobby): a player can pick a pre-built deck, and see what the engine really does with it

krakenhavoc:
A player who just wants to play a working game had no path to one: the only way to fill a seat was to upload a decklist and take your chances on catalog coverage. Four curated decks already existed — already build-tested to contain no card the engine fails to implement — and only a **bot** could reach them. This opens them to humans, and tells the truth about what they are.

## What the player now sees and does

In the deck panel (both the lobby's `choose your deck` section and the in-game deck-import modal), above the paste box:

- Four decks, each with its archetype, commander, colour pips and one-line plan.
- Under each, a **coverage line**: *"Every card is implemented · 55 of 88 nonbasic cards as printed"* — or *"All N nonbasic cards play exactly as printed"* for a deck that earns it.
- For the selected deck, the breakdown (*"Raid and Ransack: 22 simplified, 11 not yet reviewed. Basic lands are always played in full."*) and a collapsed **"which 33 cards, and what changes?"** disclosure listing every imperfect card with the clause the engine skips, in player-facing language.
- One button — **play this deck**. The seat then reads *"Raid and Ransack"* instead of *"deck pending"*.

The paste box is untouched and sits below, introduced as *"Or bring your own list."* The picker hides itself entirely when the server offers no pre-built decks, so the upload path is never blocked by it.

Verified by driving the real app (server + Vite, real Scryfall dump): pick → install → seat labelled, plus the 422 on an unknown deck id.

## Showing the promise honestly

This is the part that can go subtly wrong. **"Every card is registered" is not "every card works fully."** `decks_test.go` enforces the first — it fails the build the moment a card in a pre-built deck stops resolving to a registered `effects.Spec`. The second is what a player hears when a picker says "all cards implemented", and on these four decks it is not true.

So `internal/decks/coverage.go` computes each deck's profile from `internal/catalog` — **the same computation the /catalog page runs**, not a second one. That matters beyond DRY: completeness is merged across faces, and an MDFC whose front face has no spec earns a caveat no card file declares. A reimplementation would eventually disagree with `/catalog` about the same card, on the same site.

The client then makes only the claim that is true: the phrase *"exactly as printed"* is applied to the `full` count and nothing else, and *"every card is implemented"* is said only when `unregistered` is zero. `prebuiltDecks.test.ts` pins both rules.

`unreviewed` is kept as its own bucket rather than rounded into either column — it means *nobody has graded this card*, which is neither a known gap nor a clean bill of health.

## The four decks' real completeness profile, measured today

Counted per distinct card. Each deck is 100 physical cards; the basic-land rows are exempt (the engine synthesises a basic's mana ability from its type line) and are the reason these do not sum to 100.

| Deck | Non-basic cards | Full | Caveats | Unreviewed | Basic rows | Unregistered |
|---|---|---|---|---|---|---|
| `izzet-aggro` — Raid and Ransack | 88 | **55** | 22 | 11 | 2 | 0 |
| `simic-ramp` — Deep Roots | 88 | **61** | 18 | 9 | 2 | 0 |
| `esper-control` — The Long Answer | 92 | **70** | 16 | 6 | 3 | 0 |
| `mono-black-aristocrats` — Body Count | 82 | **60** | 9 | 13 | 1 | 0 |

Every card in every deck is still registered — the coverage test passes, and the `CMDCTRL_SCRYFALL_DUMP`-gated test still validates all four against the real dump (117,738 printings). So 62–76% of each deck plays exactly as printed, and the rest is implemented-with-a-declared-simplification or ungraded. The caveats are mostly small and honestly stated by the card files (Farseek fetching only basics, shocklands entering tapped when put onto the battlefield by a spell, Krosan Grip without split second). **That is the number the picker shows.** It is lower than "all their cards are implemented" sounds, and hiding it would have been the wrong trade.

A new gated test asserts the profile is **identical with and without a Scryfall dump loaded** — `GET /decks` is served either way, and a player must not see one number in staging and another in production.

## The package move, and why

**Moved:** `server/internal/aiseat/decks` → `server/internal/decks`.

These decks were never bot-specific; what was bot-specific was that only bots could reach them. A human-facing feature importing a package called `aiseat` is a lie about what depends on what, and the next person pays for it. The bot picker still reads the same registry through `cmd/server`'s `botDeckCatalog` adapter (unchanged behaviour, `cmd/server/botdecks_test.go` still green).

The one real cost: `profile.go` projected a deck onto `model.DeckProfile` for the LLM tiers' prompt, so a neutral `internal/decks` would have had to import `internal/aiseat/model`. Rather than accept that inverted edge, the **projection** moved instead — to `server/internal/aiseat/deckprofile`, which imports both and which nothing in `internal/decks` knows about. It is not in `cmd/server` with the other boot wiring because it is a tested projection with rules of its own; it is not in `internal/aiseat/tiers` because `DeckProfileFunc`'s comment is explicit about that package not depending on the deck catalog.

Net dependency direction: `lobby → decks`, `cmd/server → decks`, `aiseat/deckprofile → decks`. `decks` imports `cards`, `deck` and `catalog`, and nothing under `aiseat`.

## One install path, not two

`deck` is a **field on `POST /games/{id}/decks`**, not a route of its own. The server expands the deck id to that deck's plain-text decklist and runs the existing `ParseText → Resolve → Validate → SetDeck` pipeline. `POST /games/{id}/seats/bot` has had exactly this shape since S31, for exactly this reason: a pre-built deck cannot be legal by a rule an uploaded one is not held to. Sending both `deck` and `source` is a 400 rather than a precedence rule; an unknown id is a 422 naming the ids this build has.

`TestEveryPrebuiltDeckInstalls` runs all four through the real HTTP route in CI (against a synthesised index, since CI has no dump), so a deck a player can pick but not receive fails here rather than in their face.

### New route

`GET /decks` — the catalog plus each deck's coverage profile. Session-gated and game-independent, like `/bot/options`, and deliberately not folded into it: a deployment with no bot host still wants a deck picker, and `aiseat.DeckInfo` has no room for a coverage profile (a bot does not read disclosures). Documented in `docs/lobby.md`, including how to read the coverage block. `/decks` is a new top-level prefix, so it is added to `deploy/Caddyfile`'s `@api` matcher, `client/vite.config.ts`'s proxy and the service worker's `API_PATH` (with a test) — the three that must agree or it 404s in production only.

## Scope

**No new decks.** Four is enough to ship the surface and deckbuilding is a different task; I did check the existing four rather than assume, which is where the table above comes from.

Is four enough to ship? Yes for shipping, no for staying. Four decks is one archetype each and three of the four are two-colour — a player who wants mono-green ramp or an Equipment deck still has to bring a list. I have filed an issue proposing more rather than spending this PR on card lists.

## Gates

- `gofmt -l .` clean, `go vet ./...` clean, `go test -race -timeout 30m ./...` all packages pass.
- `npm run check` 0 errors 0 warnings, `npm run lint` (eslint + prettier) clean, `vitest` 505 passing.
- `CMDCTRL_SCRYFALL_DUMP=… go test ./internal/decks/ -run RealDump -v` — all four decks legal against the real dump, and the coverage-invariant test passes. The nightly workflow's path was updated for the move.


krakenhavoc:
Follow-up on deck breadth filed as #554 — the picker needs more than four decks to stay useful, but that is card-list work and does not belong in this PR.

Issue #544, opened 2026-09-14
A bot seat wedges the table forever on Myriad Landscape's search, and the wedge survives a restart


krakenhavoc:
**A bot seat can trap itself on a search prompt and hold the whole table forever. It survives a restart.** Seen live on a two-bot game; reproduced as a failing test in #543.

## Symptom

A bot activates Myriad Landscape. The table stops. Nothing is dispatched, the replay's frame count goes static, and the seat never acts again. Restarting the server does not help: the activated ability comes back on the stack, resolves, and the table wedges again on the same move.

## Mechanism

Three pieces, each defensible on its own.

1. `server/internal/cards/effects/myriad_landscape.go` enforces "up to two basic land cards that **share a land type**" through `SearchLibrarySpec.Validate` (`basicsShareALandType`). That is the right call — the constraint is a property of the *pair*, so no per-card `Predicate` can express it.
2. `server/internal/legal/choices.go`, `case game.PendingChoiceSearchLibrary`, enumerates `combinations(c.SearchCards, 1, c.SearchMax, e.opts.MaxExpansionPerSource)`. It never consults `Validate` — the hook rides on the unexported `searchResume` frame and the enumerator has no access to it — so it offers mismatched pairs as if they were legal moves.
3. `server/internal/game/search_choice.go`'s `ResolveSearchLibrary` rejects such a pair with `ErrInvalidParam` and **deliberately does not dequeue the choice**, so that a client "can try again rather than losing the prompt".

A human retries differently. `heuristic.decideChoice` is a deterministic argmax and `valueOfChoice`'s `choiceSearchLibrary` branch sums `st.cardValue` over the picked cards, so a two-land pick always outscores a one-land pick and ties go to the lowest index. Same move list in, same rejected move out, forever.

There is no escape hatch. `legal.EnumerateFor` returns **only** a choice's answers while a seat owes one (`legal.go:173-178`), so `aiseat.PassIndex(moves)` is `-1`, and the rescue in `Runner.step`

```go
if rejects >= r.cfg.MaxConsecutiveRejects {
        if pi := PassIndex(moves); pi >= 0 {
                idx, reason = pi, "forced pass after repeated rejections"
        } else {
                return true   // <- sleeps, holding the table
        }
}
```

falls through to `return true`. The guard fires correctly; it just has nothing to force. Rejected dispatches never broadcast, so from the outside the table looks idle rather than looping — a static frame count is the signature, not a churning one.

[...]

Issue #352, opened 2026-09-11
Mana pipeline: restricted / derived / scaled / gated mana, and mana inside a mana ability's cost


krakenhavoc:
The [card-coverage roadmap](https://github.com/krakenhavoc/cmd_and_ctrl/blob/main/docs/decklists/card-coverage-roadmap.md)
ranks the mana pipeline **#2 of 24 missing mechanics** across the next
2000 Commander cards — **98 cards it is the sole blocker for, 183 it
appears in, 115 it is the dominant blocker for** — and it is the only
entry in the top ten with no tracking issue. This is that issue.

Umbrella: #293. Sibling mechanic issues: #93 (cost modification, rank 1),
#95 / #176 (protection, rank 3), #279 (until EOT, rank 4), #280
(attachments, rank 6).

It is the dominant blocker for batch 01 (#294, 16 cards) and a
runner-up or dominant blocker in batches 02, 03, 06 and 07.

## The five sub-gaps

The roadmap groups this as "restricted / derived mana, mana from a
spell, gated or scaled mana abilities". Reading the actual cards it is
five distinct seams, in ascending cost:

### 1. A mana component inside a *mana* ability's cost

`ManaAbilityCost` carries `Tap`, `Sacrifice`, `SacrificeOther` and —
since #267 — `Life`. It has no mana component, so `{1}, {T}: Add {W}{U}`
is unwritable. That is **the entire Signet cycle** (10 cards, ranks
159–1075), plus Cabal Coffers' `{2}, {T}` and the filter-land cycle's
`{W/U}, {T}`.

The roadmap calls this the cheapest sub-gap and it is: one field, one
validate-and-pay step in `ActivateManaAbility`, no new game state.

### 2. Restricted mana — "spend this only on X"

Ancient Ziggurat, Eldrazi Temple, Shrine of the Forsaken Gods,
Delighted Halfling, Cavern of Souls, Secluded Courtyard, Unclaimed
Territory, Pillar of the Paruns.

`ManaToken.Restrictions []string` has existed since S15 with the
comment *"a list of opaque tags the cost validator can honour"* — and
nothing has ever written one or read one. The production half is
trivial; the whole difficulty is the **spend half**.

> ⚠️ **This is a #259 trap.** If the restriction is not enforced at
> spend time, "add {C}, spend only on Eldrazi" is plain `{C}` and every
> card in this group ships **strictly better than printed**. Either
> both halves land or neither does.

[...]

Issue #76, opened 2026-04-19
S24 — Equipment, auras, attachments


krakenhavoc:
## Scope (~30 cards)

- [ ] Attached-state on Card (aura ↔ enchanted-permanent link, equipment ↔ equipped-creature link)
- [ ] Attach, Detach, EquipPay, EnchantTarget primitives
- [ ] **Aura attach infrastructure + control-change replacement for Mind Control** — deferred from S17 per [ADR 0013](../blob/main/docs/decisions/0013-replacement-effects.md) §out-of-scope. Aura's controller grants control over the enchanted creature via a layer-2 continuous control-change effect registered at attach time. Uses the S17 `ReplacementEffect` pipeline for the attach/detach boundary.
- [ ] Sword of Feast and Famine, Sword of Fire and Ice, Lightning Greaves, Swiftfoot Boots, Skullclamp (extended), Rancor, Mind Control, Curse-style auras, …

## From S17 planning (ADR 0013)

S16 bundled aura attach into S17 originally; re-homed here because attachment + combat-state coupling fits S24's equipment/aura scope more naturally than the S17 event-pipeline work. S17's `ReplacementEffect` struct + pipeline are the substrate aura-attached-state transitions plug into.

## Depends on
S16 (layer system — control-change is layer 2), S17 (replacement pipeline for attach/detach events), S18 (equipment interacts with combat keywords).

Detailed plan TBD; lands just-in-time after S23.

Estimated 2-3 weeks.

krakenhavoc:
S18 carry-in (per [ADR 0014](../blob/main/docs/decisions/0014-combat-keywords.md#11-baneslayer-angel-ships-without-protection-clauses)):

- **Protection** (CR 702.16) deferred from S18 to S24 alongside Mind Control's control-change replacement. The concrete forcing function from S18 is **Baneslayer Angel** — the card ships here with flying + first strike + lifelink but without its "protection from Demons / Dragons" clauses. When S24 lands protection, update Baneslayer's `server/internal/cards/effects/baneslayer_angel.go` to include the clauses (predicate-based guards at damage, block, target, and attach/equip SBA hooks).

krakenhavoc:
## S24 implementation status

Three PRs, all CI-green, built on [ADR 0036](https://github.com/krakenhavoc/cmd_and_ctrl/blob/main/docs/decisions/0036-attachments.md).

| PR | What | State |
|---|---|---|
[...]

PR #562, opened 2026-09-14
feat(rules): a restriction vocabulary — "can't attack", "can't block", "can't be blocked"

krakenhavoc:
## Summary

- The engine can say **"can't"** for the first time: five bits on `Characteristic.Restrictions`, written by static abilities, read when attackers and blockers are **declared** (CR 508.1c / 509.1b) and when abilities are **activated** (CR 602.5a).
- The engine and `internal/legal` agree through **one predicate per declaration**, not two copies of a rule — the #544 lesson, applied before it could happen again in combat.
- Ships Pacifism, Arrest, Faith's Fetters and Rogue's Passage, and completes **Whispersilk Cloak** and **Carrion Feeder** — two caveats retired, both cards promoted to `full`.

## Sprint

S24 — Equipment, auras, attachments

## Issues

Relates to #76 (the "**Can't attack or block**" row in its 2026-09-14 status comment — *"the cleanest next sub-PR in this seam"*). Applies the lesson of #544 / #546. New: [ADR 0045](https://github.com/krakenhavoc/cmd_and_ctrl/blob/feat/combat-restrictions/docs/decisions/0045-combat-restrictions.md).

---

## The taxonomy, and why a field rather than a keyword

Every continuous effect the catalog could express before this either changed a characteristic or granted a keyword. There was no way to say "can't", and that one sentence is why three cards were missing and two catalogued ones were wrong — Carrion Feeder was shipping **stronger than printed**, which inverts the standing convention.

`Characteristic.Restrictions` is a `game.Restriction` bitmask, OR-accumulated:

| Printed text | Bit | Who is restricted |
|---|---|---|
| "~ can't attack" | `CantAttack` | the creature |
| "~ can't block" | `CantBlock` | the creature |
| "~ can't be blocked" | `CantBeBlocked` | the **defender's** options |
| "activated abilities can't be activated" | `CantActivate` | non-mana activations |
| "…unless they're mana abilities" | `CantActivateMana` | mana activations |

Three reasons it is a field and not a string in `Characteristic.Abilities`, which would have needed no new field and no new wire key:

1. **It is not an ability the creature has.** Pacifism's restriction belongs to the *Aura*. CR 613 gives restrictions no layer at all for exactly this reason.
2. **It would render as a keyword badge** — the failure mode ADR 0038 exists to prevent, in reverse.
3. **"Loses all abilities" would switch Pacifism off.** Darksteel Mutation and Song of the Dryads are the next cards in this seam and they clear `Characteristic.Abilities` in layer 6. A Mutated *and* Pacified creature still can't attack. `TestLosingAllAbilitiesDoesNotClearRestrictions` pins that ahead of the card landing — deliberately, because another agent is on layer 6 right now.

Because the field is only ever OR'd into and nothing clears it, the layer a restriction is written in is unobservable and so is its timestamp.

**`CantBeBlocked` is on the attacker and read on the defender's side.** It is carried by the permanent the effect is attached to and consumed inside `CanBlock`, the one predicate that sees both cards. Keeping that straight is most of the reason the taxonomy is written down.

**The activation pair is two bits because Faith's Fetters prints the difference** — Arrest stops mana abilities, Fetters explicitly does not ("unless they're mana abilities"). Two bits also map one-to-one onto the engine's two activation entry points, so neither gate has to know the other's card text. A single "restricted" flag would have made Fetters read as Arrest and switched off half the mana base it was pointed at.

## How the engine and the enumerator are kept in agreement

**Named predicates, one copy each, called by both sides.** This is the part #544 is about: an enumerated move the engine refuses is not a bad move, it is a hung table, because a seat that owes a decision is enumerated that decision's answers and nothing else.

| Predicate | Engine callers | Enumerator callers |
|---|---|---|
| **`game.AttackerEligible(card, seat)`** | `DeclareAttackers` (bulk) | `legal.combatMoves` |
| **`game.CanBlock(attacker, blocker)`** | `DeclareBlocker` | `legal.combatMoves`, `seatOwesBlockDecisionLocked` (#328's auto-pass guard) |
| **`game.CanActivateAbilities`** | `ActivateCatalogAbility`, `ActivateLoyalty` | `legal.activatedMoves` |
| **`game.CanActivateManaAbilities`** | `ActivateManaAbility`, `gatherTapSources` | `legal.manaMoves` |

`AttackerEligible` is new and is modelled on `BlockerEligible` (#328). The attacker-side eligibility list — creature, controlled by seat, untapped, not already declared, no defender, not summoning sick — **used to be spelled out in `mutations.go` and spelled out again in `legal/combat.go`**. It is now one function, and "can't attack" joined it in one place.

`internal/legal/restrictions_test.go` asserts the invariant with `dispatchAll` (every enumerated move is accepted by the dispatcher) **and** asserts the engine really does refuse the withheld move — without that second half a future enumerator could pass by going silent on both creatures.

The auto-tapper is in the table on purpose: it *plans* a payment that `ActivateManaAbility` executes, so a source the executor will refuse must never be planned. The failure mode is a half-paid cost with earlier permanents already tapped — the same argument #352 made for condition-gated abilities.

## Did the client need changing?

**Two lines of rendering, no rules.** `CardView.restrictions` is a new additive `[]string` of snake_case tokens, separate from `abilities` because the two render differently: a keyword gets a badge, a restriction disables a control and supplies the reason.

- `attackAll.ts` gains a fourth `AttackBlocker` reason, `"restricted"` → *"can't attack"*. The count on "Attack Bob with all 5 creatures" was previously off by one per pacified creature, silently, because the bulk verb skips what it disagrees with.
- The right-click ability menu greys an Arrested permanent's rows with a reason instead of opening a rejection toast — and leaves a Fettered permanent's **mana** row alone, which is a visible difference between the two cards.

Nothing re-derives who may attack. The field is read, not computed; #429 spent a PR deleting client-side rules duplication and this does not start a new one. The `CombatArrows.svelte` / `attackTargets.ts` path already reads the server's sets and needed nothing.

## Cards

| Card | State |
|---|---|
| **Pacifism** | new, `full` |
| **Arrest** | new, `full` — including "activated abilities can't be activated" |
| **Faith's Fetters** | new, `full` — "enchant permanent", the 4 life, and the mana carve-out |
| **Rogue's Passage** | new, `full` — EDHREC rank 18, and the turn-scoped half of the vocabulary |
| **Whispersilk Cloak** | caveats → **full**, `RestrictAttached(CantBeBlocked)` |
| **Carrion Feeder** | caveats → **full**, `RestrictSelf(CantBlock)` |

Two caveats retired, both of which named this gap as the cause:

- *"'Can't be blocked' isn't granted — the equipped creature blocks and is blocked normally."*
- *"'This creature can't block' isn't enforced — the Feeder can block."*

**Rogue's Passage earns its place beyond being a top-20 card.** It proves the half the Auras cannot: the same bit with a *duration* instead of an attachment, registered into the S32 until-end-of-turn registry and swept at cleanup, with no permanent on the battlefield saying it is in effect. Both paths write one field and the same two consumers read it — which is the argument for a field rather than a per-card hook, made in running code.

Three card-facing primitives, one per duration: `RestrictAttached`, `RestrictSelf`, `RestrictUntilEOT`. `EnchantPermanent()` is new beside `EnchantCreature()`, and it matters beyond Fetters' target clause: the CR 704.5n legality re-check runs the Aura's own spec every turn, so an Aura declared "enchant creature" falls off a host that stops being one and this one does not.

## Are Silent Arbiter and Crawlspace reachable now?

**Yes, and deliberately not in this PR.** They are not bits — a count is a property of the whole declaration, not of one permanent — so shipping them as bits would have been the expensive wrong abstraction this seam was warned about.

The seam is already there: `BlockerCountValid` validates a block *count* at the declare-blockers close-out for menace, and `DeclareAttackers` receives the entire attacking set in one call. A count limit belongs beside `BlockerCountValid` as a set-shaped predicate, with the per-permanent bits untouched. The open question to answer **before** writing them is enumerator agreement, which is harder for a count than for these: a per-move enumerator has to reason about a set. That is stated in ADR 0045 rather than discovered later.

Also out, with the reason recorded:

- **Propaganda / Ghostly Prison / Norn's Annex** — "can't attack unless its controller pays {2}" is a *cost*, not a prohibition. A bit cannot carry a cost. It wants an attack-cost pipeline keyed on the **defending player**. Nothing here blocks it.
- **Goad** and **landwalk** are conditional on the other side of the pairing. Goad's home is `canAttackTargetLocked` (already shared with `AttackTargetsForEffect`); landwalk's is `CanBlock`, once it can see the game. Lord of Atlantis's islandwalk caveat stays declared for that reason.

## One rules consequence worth stating

Restrictions are checked **at declaration and only there**. CR 506.4 lists what removes a permanent from combat — leaving the battlefield, changing control, ceasing to be a creature — and "acquired a restriction" is not on it. A creature pacified after attackers are declared keeps attacking and deals its damage. `TestRestrictionIsCheckedOnlyAtDeclaration` pins it, because the intuitive-but-wrong implementation (re-checking during the damage step) would be a silent rules bug in the *player's* favour.

## Test plan

- [x] `server/internal/game/restrictions_test.go` — every engine gate, driven through the real layer pipeline via a stubbed `CatalogStaticAbilities` hook rather than a hand-stamped `Card.effective`: both attack verbs, both block directions, the `#328` auto-pass signal, the CR 506.4 mid-combat case, the layer-6 ability-strip interop, the Arrest/Fetters activation split, a Fettered planeswalker's loyalty abilities, and the auto-tapper.
- [x] `server/internal/legal/restrictions_test.go` — the #544 invariant: `dispatchAll` over every enumerated move, plus the assertion that the engine refuses what the enumerator withheld.
- [x] `server/internal/cards/effects/restrictions_test.go` — the six cards through a real cast or activation, including Pacifism releasing the creature when the Aura is destroyed and Rogue's Passage expiring at cleanup.
- [x] `client` — `attackAll.test.ts` (the new reason, its precedence, and that the *block*-side bits do not exclude a creature from an attack) and `contextMenu.test.ts` (Arrest greys both rows, Fetters greys one).
- [x] `gofmt -l .` clean, `go vet ./...` clean.
- [x] **`AISEAT_GAME_TESTS=1 go test -race -timeout 40m ./...`** — all packages ok, `internal/aiseat` at **579s** with the gated whole-game bot tests on. Combat is where bots spend most of their decisions and that gating is exactly why #544 shipped.
- [x] `npm run check` 0 errors / 0 warnings, `npm run lint` clean, `npx vitest run` **497/497**.
- [x] Coverage census regenerated after rebasing on `origin/main`: 1542 → **1546** registry keys, `full` 1070 → 1076, `caveats` 389 → 387.

ADR number checked across every remote branch per AGENTS.md §"Picking an ADR number" (`0001`–`0044` present, `0005`/`0024`/`0029`/`0030` permanently unused); **0045** was free at the time of writing.


krakenhavoc:
Brought up to date with `origin/main` after **#555** (23 more Equipment and Auras) landed. Merged rather than rebased so the push stays a fast-forward and nothing already pushed is rewritten.

The only conflict was the catalog census, resolved the documented way — take `origin/main`'s side, regenerate with the coverage package's `-update`. **The numbers in the PR body above are from the pre-#555 baseline and are superseded by these:**

| Measured | before | after |
|---|---:|---:|
| Registry keys | 1565 | **1569** |
| Declared `full` | 1090 | **1096** |
| Declared `caveats` | 392 | **390** |

Checked #555's 23 new cards for caveats this PR would retire: none. Beamtown Beatstick is the only one of them whose text mentions blocking, and its caveat is about combat damage to a *battle*, which is a different gap.

Re-ran everything on the merged tree: `gofmt` / `go vet` clean, `go test -race ./...` green, `npm run check` 0/0, `npm run lint` clean, `vitest` **510/510** (up from 497 — #555's client tests came in with the merge). CI green on `21d029c`.

Issue #59, opened 2026-04-19
S12.5 — Discord identity for players


krakenhavoc:
**Phase:** 6 (follow-up to S12 deploy) · **Goal:** the playgroup uses Discord for coordination already — lean all the way in. Invite links opened from Discord land the player in the game with their Discord display name and avatar; a bot posts/DMs invites from Discord itself; the player's presence reflects what they're doing in the game.

Every "out of scope" deferral from initial planning is pulled into this single sprint.

### Discord OAuth (core)
- [ ] Register the Discord application; `CMDCTRL_DISCORD_CLIENT_ID` / `CMDCTRL_DISCORD_CLIENT_SECRET` env vars (dev + prod). Callback URLs for both environments registered with Discord.
- [ ] Server routes: `GET /auth/discord/start?game=<id>&t=<invite>` builds the Discord authorize URL with PKCE + state; `GET /auth/discord/callback` exchanges the code, calls `/users/@me`, completes `Lobby.Join` on the bound invite, mints the session.
- [ ] Server-side state store (`game_id`, `invite_token`, `pkce_verifier`, 5-min TTL) — reuse the pattern from `auth.MemoryAuthenticator`; no schema migration.
- [ ] Extend `auth.Principal` with optional `DiscordID`, `DiscordUsername`, `DiscordAvatarHash`, `DisplayName`. `MemoryAuthenticator` preserves them across `Validate`.
- [ ] Extend `lobby.SeatInfo` with `display_name` (Discord `global_name`, fallback username, final fallback manual name) and `avatar_url`.
- [ ] Client `Join.svelte` gets a primary "Sign in with Discord" button; manual name entry kept as fallback.

### Avatar rendering
- [ ] `client/src/lib/protocol.ts` — `PlayerView` / `SeatInfo` grow the two optional fields.
- [ ] `PlayerHeader.svelte` renders a 24px circular avatar before `.seat-dot` when present.
- [ ] Lobby seat list + any S10 politics / commander-damage UI that names seats pick up `displayName` / `avatarUrl`.
- [ ] Server-side avatar cache `$CMDCTRL_DATA_DIR/avatars/<discord_id>/<hash>.png`: first fetch hits `cdn.discordapp.com`, cache with immutable headers, re-fetch when hash changes. Client requests `/avatars/<discord_id>`; server serves from cache or proxies on miss. Keeps Discord CDN URLs out of the state stream.

[...]

PR #597, opened 2026-09-16
feat(auth): Discord sign-in before the invite — identity-only sessions

krakenhavoc:
## Summary

- **Discord sign-in moves to the login page.** `/auth/discord/start` now takes either `?game=&t=` (invite-link flow, unchanged) or no parameters at all. The unbound flow mints an identity-only session and returns to the SPA, which collects an invite code. Half the pair is a 400 — guessing would silently discard a seat claim the user asked for.
- **A fourth role, `auth.RoleIdentified`:** the Discord fields, no `GameID`, no `PlayerID`. It can do exactly one thing — `POST /join` with a code, which trades it for a player session. `WSAuthorizer` refuses it *by name* rather than via the `default` arm, whose "unrecognised principal role" would be a lie about a role the server mints itself.
- **`Lobby.FindByInvite` + `POST /join`** resolve a bare code to its table, so the login box takes a code *or* a link. The lookup does not short-circuit on match (response time shouldn't be a function of map position), and skips spectator invites and archived tables.

Full rationale, including what this deliberately softens in ADR 0003, is in [ADR 0050](docs/decisions/0050-discord-login-identity.md).

## Sprint

S12.5 — Discord identity for players.

**This is not the active sprint (S33).** It is a follow-up to the S12.5 exit criteria: sign-in was built bound to an invite link, so the button could only ever appear on a page you reach by clicking a link that already names a table. Landing on the site and signing in was not possible. The change is self-contained (one new role, one new route, one lookup) and touches no S33 surface.

## Issues

Relates to #59. Closes nothing — the bot, Rich Presence and re-link tasks on that issue are untouched.

## Notable trade-off

Invite tokens are now globally searchable by anyone who can reach `POST /join`. Previously you needed the game id **and** the token; now the token alone suffices. Entropy (16 random bytes) plus the shared rate-limit bucket make guessing impractical, but it is a real widening of what a token is, so it is recorded in the ADR rather than left implicit.

## Drive-by fix

`docs/lobby.md` claimed `409` covers an empty name on `POST /games/{id}/join`. The error switch in `http.go` maps `ErrEmptyName` to **400**. A test asserting the documented behaviour is what caught it; both rows are corrected.

## ADR number

`0050`. Checked across **all 130 remote branches** with the §4 `git ls-tree` scan — `0028`–`0049` are in use, `0050` was free.

## Test plan

Run locally, all green:

- [x] `go build ./...`, `go vet`, `gofmt -l` clean
- [x] `go test -race` on `internal/lobby`, `internal/auth`, `internal/discord`
- [x] `internal/docsguard` (ADR number uniqueness + H1 agreement)
- [x] `npm run check` + `npm run lint`
- [x] New tests: unbound start; half-an-invite rejected; callback mints an identity session with no seat in the fragment; join-by-code seats the Discord user; a body `name` cannot relabel a verified identity; anonymous join still needs a name; unknown code, archived table, and already-seated session all refused; `FindByInvite` ignores spectator invites; WS authorizer refuses an identity session

Still to do by hand, on `cmd-dev.labxp.io` after merge — the part no test covers:

- [ ] Sign in from the login page, paste a bare code, land seated with the Discord name + avatar
- [ ] The same page with a pasted full link (should go via the Join page as before)
- [ ] Manual name entry still works with Discord configured


krakenhavoc:
Follow-up commit `28306b1` — two bugs in the first commit, found while fixing a red CI run.

- **`joinByCode` landed on `#/games/<id>`, skipping the lobby.** The invite-link flow goes through it deliberately (`Join.svelte`: "lobby first so they can import a deck and see other seats' status", s085/#43), so a code-joiner would have arrived at a table with no deck. Both paths now land in the same place.
- **The page never collected a name.** ADR 0050 calls the anonymous bare-code path first-class, but the UI posted an empty name and took a 400 the user could do nothing about. A name field now appears only when the input is a bare code *and* no Discord identity is signed in; a pasted link still hands off to the Join page, which asks for itself.

**e2e selectors updated**, since the invite box is now labelled `invite code or link` and its button says `join` (`smoke.spec.ts`, `auth.spec.ts`), plus a new test for the bare-code path in `join.spec.ts`.

⚠️ **The e2e specs could not be run locally.** Playwright 1.59.1 publishes no browser build for this host: `ERROR: Playwright does not support chromium on ubuntu26.04-x64`. These specs are nightly-only in CI, so the next nightly is where they first execute — the selector changes were checked against the component source by hand, not by running them. Worth a look on the next nightly.

Also reformatted `Login.svelte` per Prettier — that was the original CI failure, masked locally by a `npm run lint | tail` that reported `tail`'s exit code instead of npm's.

Issue #602, opened 2026-09-16
E2E: the #266 freeze walk times out because most priority windows now auto-pass, not because a board froze


krakenhavoc:
`state-freeze-266.spec.ts` has failed on the 09-14 and 09-15 nightlies, both attempts, with `Test timeout of 90000ms exceeded` inside the walk loop.

The spec walks twelve priority windows by clicking `next`, and each iteration waits up to 20s for some seat to hold priority and then up to 15s for both boards to agree on the step. Since #429 the client's smart auto-pass asks the server's move list instead of walking cards, and far more windows now pass on their own — so iterations spend their time waiting for a cursor that does not stop, and twelve of them no longer fit in the 90s `test.slow()` budget.

Neither captured page snapshot shows a frozen board: one has `next` enabled at seq 64, the other has it disabled at seq 85, which is two attempts of the same page at different points, not two clients disagreeing.

## Proposed fix

Keep what the spec guards — both clients converge, the game advances, no client error — but stop depending on how many windows happen to stop:

- give the walk an explicit budget rather than inheriting `test.slow()`;
- treat "no seat held priority" as the cursor having moved on rather than a failure, and drive the next window when nobody holds it;
- compare the two boards on a settled snapshot rather than on step text sampled mid-advance.

This is a test-robustness change. #266 itself stays open: the freeze it was written for has never been reproduced here.

krakenhavoc:
**Closing 2026-09-16:** resolved by **#606**, merged to `develop` (live on cmd-dev). It reaches production with promotion PR **#700** (develop → main).

Issue #601, opened 2026-09-16
Catalog soak: fail a bot game on an effect error, play random catalog decks, and report which cards bots actually exercised


krakenhavoc:
Bots already play complete four-seat games every night. They exercise the engine, but nothing checks whether the **cards** did anything, and the decks they play are a handful of hand-written vanilla cards plus four curated lists — a few hundred of the ~1,500 registered specs.

Three cheap additions turn the existing soak into a catalog exercise. None of them claims to verify a card is *correct*; that stays the job of the per-card unit tests.

## 1. Fail a game on an effect error

The engine already emits `EventEffectError` whenever a catalog primitive fails, and deliberately continues so a partial failure cannot wedge the stack. Nothing in `internal/aiseat` reads it, so a card that throws on every resolution is invisible for as long as the game still finishes.

Assert zero `EventEffectError` events at the end of a bot game, reporting the card and message.

## 2. Play random decks drawn from the whole catalog

`newRandomBattleRoom` seats four copies of `battleDeck` — Mountains, Bears, Ogres and a few Bolts. Drawing decks from `effects.All()` instead, seeded so a failure reproduces, puts real catalog cards on the battlefield.

Needs card data (type line, mana cost, colour identity), so the deck builder reads the Scryfall dump and the nightly `bot-games` job gains the dump steps the `playwright` job already has. Without the dump the test skips, exactly like the other dump-gated tests.

## 3. Publish a per-card report

Walk the event log at the end of each game and count, per catalog card: cast, resolved, entered the battlefield, triggered, errored. Two things fall out:

- the cards bot games never reach, which is where new unit tests are worth writing;
- the cards that error, which are bugs.

## Not in scope

Per-card behavioural assertions. A bot game cannot say a card did the *right* thing without a per-card expectation, and writing 1,500 of those is the unit-test suite that already exists.

Refs #89.

krakenhavoc:
**Closing 2026-09-16:** resolved by **#606**, merged to `develop` (live on cmd-dev). It reaches production with promotion PR **#700** (develop → main).

Issue #600, opened 2026-09-16
Nightly bot-games: TestManagerPlaysALobbySeatedTable's 60s budget is smaller than the test's own -race runtime


krakenhavoc:
The nightly `bot-games` job has failed on 09-14 and 09-15 with:

```
--- FAIL: TestManagerPlaysALobbySeatedTable (60.02s)
    manager_test.go:60: table did not finish: turn 16 step declare_blockers
```

It is a budget, not a hang. The test waits a hardcoded 60s for four random bots to finish a table, while the job runs the whole package under `-race` on a self-hosted runner shared with CI.

## Measurements

Same commit (`62adbb5`), containerised Go 1.22, idle machine, three consecutive runs:

| mode | per-run wall clock |
|---|---|
| no race detector | 2.6s, 2.9s, 3.6s |
| `-race` | 22.7s, 33.2s, 30.6s |

So an idle machine already spends half the budget, and the nightly's own log shows the package taking 1037s wall for that step. Contention does the rest.

## Proposed fix

Take the budget from the same `AISEAT_WALLCLOCK` knob the other whole-game tests use (default 300s) instead of a literal 60s, so a loaded runner is a slow pass rather than a failure, and the knob documented in AGENTS.md covers this test too.

krakenhavoc:
**Status 2026-09-16:** the fix (#606, `AISEAT_WALLCLOCK` budget) is merged to `develop`. A manual run of the workflow on #606's branch passed (run 35096986069). The scheduled nightly runs against `main`, so it stays red until promotion **#700** merges. **Keeping this open until the first scheduled nightly on `main` after #700 passes**; close it then.

krakenhavoc:
The first scheduled nightly on `main` after promotion #700 passed: run 35228656578 (E2E, `schedule`, 2026-09-17 13:41 UTC, head 5400eb5), which includes the #606 `AISEAT_WALLCLOCK` fix. Closing as fixed.

Issue #599, opened 2026-09-16
Smart autopass closes the declare-attackers window straight after a bulk declaration, so #318's undo is unreachable


krakenhavoc:
The nightly E2E has failed every run since 2026-09-12. `attack-all-318.spec.ts` fails at the same assertion on both attempts, and the cause is a real product bug rather than a flaky test.

## What happens

`Game.svelte` renders the #318 attack-with-all cluster while `attackAllReady || canUndoDeclaration`, and the Undo button inside it while `canUndoDeclaration` — which requires `canDeclareAttackers`, i.e. the cursor still in `declare_attackers`.

With default settings (`declare_attackers` is a default stop, `smartAutoPass` on) the window is held by `hasAnyLegalResponse`, which since #429 is `legal_moves.some(m => m.kind !== "pass")`. The enumerator offers an attack move per **eligible** creature (`legal/combat.go`, `game.AttackerEligible`). A declared attacker is tapped and no longer eligible, so **after a bulk declaration there is no non-pass move left**, smart auto-pass fires, the cursor leaves `declare_attackers`, and the cluster unmounts.

The undo affordance is therefore unreachable in normal play. `attackAll.ts` states the design it breaks: the bulk verb exists so that one undo restores tap state and declarations together, "which is why this button is here rather than only in the ⋯ menu".

## Evidence

- `tests-e2e/tests/attack-all-318.spec.ts:163` — `expect(undo).toBeVisible()` fails with `element(s) not found`; the preceding `waitFor` confirming three declared attackers passes, so the declaration itself works.
- Nightly runs 34693258621 (09-12), 34759717262 (09-13), 34860357579 (09-14), 34977159355 (09-15): same spec, same assertion, both attempts.
- Failure window brackets #429 (S31 sub-PR 2), which moved the smart-autopass predicate onto the server's move list. No client file changed in that window.

## Proposed fix

Hold the `declare_attackers` window while the viewer has declared attackers this combat and can still undo. This belongs next to the `owesBlockDecision` guard in `client/src/lib/priority.ts` (#328), which exists for the same reason: an auto-pass that closes an unrecoverable window.

Scope is the smart-autopass path only. An explicit autopass toggle still passes.

krakenhavoc:
**Closing 2026-09-16:** resolved by **#606**, merged to `develop` (live on cmd-dev). It reaches production with promotion PR **#700** (develop → main).

PR #606, opened 2026-09-16
fix(nightly): the E2E goes green again, and bot games start exercising the catalog

krakenhavoc:
## Summary

- **The nightly has been red since 2026-09-12** — three failures, two causes, neither of them flakiness. `attack-all-318` fails because smart auto-pass closes the declare-attackers window the instant a bulk declaration lands, taking #318's Undo button with it; `state-freeze-266` times out because most windows now auto-pass and the walk waits for a cursor that never stops; `TestManagerPlaysALobbySeatedTable` fails because its 60s budget is smaller than its own `-race` runtime.
- **Bot games start exercising the catalog.** `TestCatalogSoak` deals decks out of the registry itself (1,521 of 1,584 specs are dealable), fails on any `EventEffectError`, and writes a per-card report of what the bots actually reached. The nightly uploads it on every run.
- **It found two real defects on its first two runs**, both now filed: #499 reproduced on an ordinary catalog card, and #605 — a commander-zone prompt that outlives its card, leaving a seat with 52 legal moves the engine refuses and a wedged table.

## Sprint

S31 — AI bot seat, for the two `internal/aiseat` commits.

**Out of sprint, and stated per AGENTS.md §4:** the client fix (#599) and the e2e walk (#602) belong to no active sprint. They are here because they are the rest of "the nightly is red", and splitting them out would leave a PR that fixes a third of it. #599 is a one-predicate client change with unit tests; #602 touches one spec.

## Issues

Closes #599, closes #600, closes #601, closes #602.
Relates to #89, #266, #318, #429, #499, #605.

## Test plan

Everything below was run against this branch in a Go 1.22 container and with the repo's own Node toolchain.

- `go build ./...`, `go vet ./internal/aiseat/...`, `gofmt -l .` — clean.
- `make lint` (golangci-lint v1.64.8, Go 1.24 toolchain) — clean.
- `go test ./internal/aiseat/ -run TestManagerPlaysALobbySeatedTable -count=3` — passes; 2.6s/2.9s/3.6s.
- Same test with `-race`, `-count=3` — passes; 22.7s/33.2s/30.6s, which is the measurement behind #600.
- `AISEAT_GAME_TESTS=1 CMDCTRL_SCRYFALL_DUMP=… AISEAT_CATALOG_GAMES=2 go test -run TestCatalogSoak` — passes: `83 of 1521 dealable cards were reached; 0 errored`, one stall reported and attributed to #499.
- `npx vitest run src/lib/priority.test.ts` — 22 passed, including 5 new ones for the declare-attackers guard.
- `tsc --noEmit` in `tests-e2e`, `prettier --check` and `eslint` on every file touched — clean.
- **Playwright itself has not been run locally** (no Go toolchain on this machine outside a container, and the suite starts the server with `go run`). The E2E workflow has been dispatched against this branch; the two specs it fixes are the verification, and this PR should not merge until that run is green.

## Notes for review

- `TestCatalogSoak` reports stalls rather than failing on them, deliberately. The reasoning is in the commit body and in a comment on the test: stall-hunting on fixed decks under pinned seeds is already `TestRandomBotSoak`'s job, whereas this test's input changes with every card PR.
- A seed fixes the four decks, not the play — the bot seats are concurrent goroutines. Both issue write-ups carry that caveat.
- The nightly's `bot-games` job now downloads the Scryfall dump, sharing the cache key the `playwright` job already uses.


krakenhavoc:
The E2E workflow dispatched against this branch is green: run 35096986069, both jobs.

- `attack-all-318` passed in 23.5s on the first attempt. It has failed both attempts on every nightly since 09-12.
- `state-freeze-266` passed in 21.1s. It had been timing out at 90s.
- Playwright: 47 passed, none failed or retried.
- `bot-games` passed, including `TestManagerPlaysALobbySeatedTable` under `-race`.
- `TestCatalogSoak` ran with the dump rather than skipping: 49 of 1,521 dealable cards reached, 0 errored, and one stall on Cavern of Souls attributed to #499.

CI/CD run 35096985277 is also green. The merge condition in the description is met.

PR #618, opened 2026-09-16
fix(legal): may_cast and choose_creature_type have answers — no seat is handed an empty list

krakenhavoc:
## Summary

- **The last two choice kinds from #499 now have answers.** A seat owing `may_cast` or `choose_creature_type` used to get an empty move list. The engine refuses `pass_priority` while a choice is open, so that meant a sleeping bot and a frozen table.
- **may_cast** offers "cast it" and "don't cast it". Declining is marked always-legal.
- **choose_creature_type** offers the creature types of the seat's own creatures, most common first, up to five. The candidates come only from the public battlefield. A creatureless board still gets "Human", so the list is never empty.
- The other two kinds #499 listed, `choose_protector` and `legend_rule`, were already handled on develop.

## Sprint

S31 — AI bot seat

## Issues

Closes #499. Relates to #601, #544.

## Test plan

- [x] Four new tests in `internal/legal/pending_kinds_test.go`. Each replays every offered answer through `actions.Dispatch` on a clone. The may-cast test also proves each answer runs its own branch.
- [x] `go test ./...` on the whole server: 28 packages ok, plus `gofmt` and `go vet ./...`.
- [x] `make lint` with golangci-lint v1.64.8 is clean.
- [x] Catalog soak on the real dump, six games on seeds 601–606, including the seeds that previously froze on Door of Destinies and Cavern of Souls. Results: six winners, **no stalls**, no "no moves enumerated" warnings, 253 of 1,521 cards reached, 0 effect errors.

## Notes for review

- Bot games aren't deterministic, so the clean soak is strong evidence rather than proof. The unit tests are the proof.
- The same soak surfaced one unrelated rejected move: the enumerator offered casting Crackle with Power at a commander, and the engine refused it with `invalid parameter`. It's filed separately, not fixed here.


krakenhavoc:
The unrelated rejected move mentioned in the description is now #619: the enumerator offers Crackle with Power with X=0 and one target, and the engine rejects that target count.

Issue #793, opened 2026-09-17
A life change can now pause on a CR 616 prompt: callers that read Life synchronously (Exsanguinate-style, pay-life costs) need a continuation


irobinson010:
Noticed while fixing #482 (PR #790).

Every life change now runs the CR 614 window (#790). That means a change can PAUSE on the CR 616 order prompt when two different life replacements apply. Two kinds of caller assume it cannot:

1. **Effects that read `Life` before and after**, e.g. "each opponent loses X life; you gain life equal to the life lost this way" (Exsanguinate, Debt to the Deathless, Torment of Hailfire's drain option, Gray Merchant): if an opponent's loss pauses, the read-back sees 0 and the gain is wrong. The right shape is the `Then` continuation the search and scry primitives use: the tail carries the continuation and runs it with the actual delta once the change lands.
2. **Pay-life costs during a cast or activation** (`ManaAbilityCost.Life`, `AbilityCost.Life`, Phyrexian mana, Snuff Out): a pause mid-cast leaves the cast half-paid. Costs should probably not be replaceable life changes at all — CR 118.3 makes paying life a cost, not a life-change *effect* — so route costs through a non-replaceable write and say so in the ADR.

**Tests.** Exsanguinate for 3 with two different life replacements on an opponent: the gain equals what was actually lost after the prompt is answered. Snuff Out with a life replacement in play: the cast completes and the life is paid once, no prompt.

irobinson010:
Claimed (2026-09-17): an agent has this, branch `fix/793-life-change-continuation` off `develop`. Please don't pick it up in parallel.

irobinson010:
PR up: #806 — the life tail now carries a continuation (`ChangePlayerLifeThenForEffect` / `LoseLifeEachThenForEffect`), and paying life goes through `PayLifeForEffect`, which runs the CR 614 window (CR 119.4 makes a payment a life loss) but settles without a CR 616 prompt. ADR 0013 §5b has the rule.

irobinson010:
[...]

Issue #792, opened 2026-09-17
CR 616 order prompt appears for identical replacement effects, whose order cannot matter (two Doubling Seasons, two Rhox Faithmenders)


irobinson010:
Noticed while fixing #482 (PR #790); pipeline-wide, so filed separately.

When two or more applicable replacement effects are the SAME effect from two copies of one card (two Doubling Seasons on a counter; two Rhox Faithmenders on a life gain; two Hardened Scales), the CR 616.1 order prompt is still queued, and the answer cannot change the outcome. It is a click the player has to make for nothing, and under #730's gate it now also blocks the table until answered.

**Fix shape.** `applyReplacementsLocked` (`server/internal/game/replacements.go`): before queueing the order prompt, partition the applicable set by identity of the *effect* (same catalog key and same index into the card's `Replacements`, or the same declared `Label`); if every applicable effect is one identity, apply in any order without prompting. #725 did the special case for pure cancels (`ReplacementEffect.PureCancel`); this is the general one. Keep the prompt whenever two different effects apply, even if they happen to commute (the rules say the affected player chooses; only unobservable choices are safe to skip).

**Tests.** Two Doubling Seasons: 4 counters, no prompt. Doubling Season + Hardened Scales: prompt as today. Two Faithmenders: quadruple, no prompt (the #790 test currently answers a prompt for this case and can be simplified).

irobinson010:
Claimed (2026-09-17): an agent has this, branch `fix/792-identical-replacements-no-prompt` off `develop`. Please don't pick it up in parallel.

irobinson010:
PR up: #800 — https://github.com/krakenhavoc/cmd_and_ctrl/pull/800

irobinson010:
Done in #800, merged to `develop`: a window whose applicable replacements all share one `replacementIdentity` (card, slot, controller) applies without the CR 616.1 order prompt; mixed windows still prompt because interleaving can matter (Season → Scales → Season). ADR 0013 §5a and AGENTS.md §7 record it.

irobinson010:
Follow-ups filed: #801 (one encode/decode helper for replacement IDs with a slot bound) and #802 (re-check `AppliesTo` between applications in a CR 616.1 chain).

Issue #482, opened 2026-09-13
Effect-side life changes skip the RepEventLife replacement pipeline (Rhox Faithmender is wrong)


irobinson010:
Reported by the batch 07 agent (#300) as a latent gap and confirmed by the batch 19 agent (#312, PR #481) as a live wrong card. Filed rather than fixed because card batches stay out of `server/internal/game/`.

## What happens
`RepEventLife` replacement effects run only on the public `ChangePlayerLife` mutator. Two other paths change life without consulting them:

- `ChangePlayerLifeForEffect` — every catalog `GainLife` / drain.
- `applyLifelinkLocked` — lifelink gains.

`DealDamageToPlayerForEffect` also emits no life-loss event, so a "whenever you lose life" watcher misses spell damage.

## Repro
1. Push Rhox Faithmender (registered `CompletenessFull` in batch 14; "if you would gain life, you gain twice that much instead").
2. Cast Lightning Helix at an opponent.
3. Caster gains 3. Expected 6. Batch 14's test only exercised the public mutator, which is why it passed.

## Cards affected
Rhox Faithmender is wrong today. Angel of Vitality (batch 19), Alhammarret's Archive (batch 08), Bloodletter of Aclazotz (batch 07) and every other life-replacement card were skipped for this. Until fixed, Faithmender should be declared `caveats` rather than `full`, or its test extended to the effect path so the drift guard sees it.

## Where
`server/internal/game/effect_api.go` (`ChangePlayerLifeForEffect`), the lifelink helper in combat, and `DealDamageToPlayerForEffect` — route them through the same `applyReplacementsLocked` call `ChangePlayerLife` makes, the way `DealDamageToPlayerForEffect` already runs the damage pipeline.

Refs #293, #300, #312.

krakenhavoc:
**Verdict: still reproducible, verified by reading the code. Confirming for the 2026-09-14 triage sweep.**

`origin/main` @ `f1bdc37`. Nothing has touched this since the report.

The primary claim holds exactly. `ChangePlayerLifeForEffect` (`server/internal/game/effect_api.go:142-155`) is:

```go
p.ChangeLife(delta)
g.EmitEvent(Event{Kind: EventChangeLife, …})
```

[...]

PR #806, opened 2026-09-17
fix(engine): a life change carries its continuation, and a cost never pauses (CR 616)

irobinson010:
## Summary

Since [#482](https://github.com/krakenhavoc/cmd_and_ctrl/issues/482) (PR #790) every life change runs the CR 614 window, so every life change can **pause** on a CR 616 ordering prompt. Two kinds of caller assumed it could not, and #793 is both of them.

**1. "You gain life equal to the life lost this way" now rides a continuation.** `ReplacementEvent` carries a `lifeTail` — the caller-facing sibling of `damageTail` — and `runLifeTailLocked` is the one place it runs, reached from every *terminal* outcome of a life event: landed (`applyResolvedLifeChangeLocked`, which is the same function the CR 616 resume calls), replaced away under CR 614.10, or its player gone. `applied` is the post-replacement delta, signed the way the event is, and `0` for the two non-outcomes — so a caller adding up several players' losses is always told, and never stalls on a leg that moved nothing.

The public surface in `effect_api.go`:

```go
// one player
g.ChangePlayerLifeThenForEffect(src, opp, -x, func(g *game.Game, applied int) error { ... })
// each opponent — the batch, built on the single form
g.LoseLifeEachThenForEffect(src, ctx.Opponents(), x, func(g *game.Game, lost int) error { ... })
```

`ChangePlayerLifeForEffect` is unchanged for the fire-and-forget majority (it now delegates with a nil `then`). The batch drains players **in sequence**, each from the previous one's continuation, so the running total is a value carried forward rather than a shared accumulator an undo could not rewind — one place knows how to wait, not N.

**2. Paying life is a life loss, but a cost never pauses.** CR 119.4: *"paying an amount of life is the same as losing that much life"*, so the window **does** run on a payment and a life-loss replacement sees it — the issue's guess that costs should skip the pipeline (CR 118.3) would have been a rules change. What a payment is not is life *gain*, so Rhox Faithmender and Alhammarret's Archive never touch it; that half needs no special case at all. What a cost may not do is stop to ask a question: CR 601.2h pays a spell's costs as one indivisible step, CR 601.2 rewinds the announcement if they cannot all be paid, CR 602.2b says the same for an activation and CR 605.3b resolves a mana ability without the stack. So `PayLifeForEffect` sets `ReplacementEvent.mustSettleNow` and the apply-loop settles without prompting: an ordering window applies in the order it was gathered (the escape §5a already uses for pure cancels, identical effects and an eliminated chooser), and anything that would ask its own question is skipped un-applied (`optionalReplacementResumableLocked`'s weaker-never-stronger posture). The declared cost is the affected player's CR 616 ordering choice on a payment.

**One undo fix fell out.** `Clone` shallow-copied the `PendingChoice`, so an undo snapshot shared the in-flight `ReplacementEvent` with the live game — and answering a prompt mutates that event in place. Undoing the answer and answering again doubled an already-doubled delta and found the continuation consumed. `cloneReplacementResume` gives the snapshot its own copy of the event (the gathered `applicable` list stays shared; a resume only reads it). That was latent for counters and damage too, not just life.

**A lint, since `go vet` can't see this.** `life_continuation_guard_test.go` parses every catalog file and fails on a `.Life` read that follows a life change in the same function, with an allowlist for the false positives that will eventually turn up. It has its own fixture test so it cannot quietly rot into a no-op.

No new ADR — this is [ADR 0013 §5b](docs/decisions/0013-replacement-effects.md), dated 2026-09-17, next to §5a from #792.

## Sprint

## Issues

Closes #793

## Callers

Every site that changes a player's life was enumerated (`ChangePlayerLifeForEffect`, `ChangePlayerLife`, `AbilityCost.Life`, `ManaAbilityCost.LifeCost`, `AlternativeCost.Life`, `AdditionalCost.PayLifeX`, `ReplacementEffect.EntryLifeCost`, ward's pay-life, the CR 702.15b lifelink credit): **74 call sites** outside the two entry-point files, of which **eleven changed**.

### Needed the amount, or needed the change to have finished — converted (4 sites, 5 cards)

| Site | What it needed |
|---|---|
| `effects/batch21_helpers.go` `b21DrainEachOpponentAndGainTheTotal` | Gained `n × len(opponents)`, which is the real total only while nothing replaces anybody's loss. Now `LoseLifeEachThenForEffect` + gain the reported total. Serves **Kokusho, the Evening Star** and **Debt to the Deathless**. |
| `effects/exsanguinate.go` | Read `p.Life` before and after each opponent's loss — zero for a paused one. Now the shared drain body. |
| `effects/gray_merchant_of_asphodel.go` | Same `x × len(opponents)` arithmetic, with the approximation admitted in its own card comment. Now the shared drain body. |
| `effects/batch34_helpers.go` `b34PayLifeToDraw` (**Crossway Troublemakers**) | "Pay N life. If you do, draw a card" — a cost whose *"if you do"* has to know the payment finished. Moved to `PayLifeForEffect`. |

### Pay-life costs — moved to the non-pausing cost path (7 sites)

| Site | Cost |
|---|---|
| `game/additional_cost.go` `payAdditionalCostLocked` | `AdditionalCost.PayLifeX` ("as an additional cost, pay X life") |
| `game/alternative_cost.go` `payAlternativeCostLocked` | `AlternativeCost.Life` (Snuff Out's "pay 4 life") |
| `game/activated.go` `ActivateCatalogAbility` | `AbilityCost.Life` |
| `game/mutations.go` (mana-ability payment) | `ManaAbilityCost.LifeCost` (Mana Confluence — CR 605.3b, no stack to pause on) |
| `game/entry_choice.go` `ResolveEntryPayLife` | a shockland's "as this enters, you may pay 2 life" — previously nested a paused pipeline inside another one |
| `effects/ward.go` `wardPayLifeOrCounter` | "unless that player pays N life" |
| `effects/sylvan_library.go` | "pay 4 life to keep it", whose chain continues per card |

Phyrexian mana is not a caller: `{W/P}`'s life self-pay is parsed but not implemented (`game/mana.go`, "life self-pay is S17" — still open).

### Read `.Life` *before* a life change — correct as written, no change

`effects/batch30_helpers.go` (Betor's "each opponent loses half their life"), `effects/batch29_helpers.go`, `effects/peer_into_the_abyss.go`, `effects/quietus_spike.go`, `effects/batch31_helpers.go` `b31LifeBecomes` (CR 119.5 "life total becomes N" — the delta is computed from the total before the change), `effects/castle_locthwain.go`. Each computes its amount from a total read *before* the change; a pause cannot make that read wrong.

### Fire-and-forget — no change needed (the remaining 63 sites)

Every "you lose N life" / "each opponent loses 1" / `GainLife` that does not use the amount afterwards: the `GainLife` primitive, `eachOpponentLosesLife`, `drainEachOpponent` (Ayara, Nadier's Nightblade, Cauldron Familiar — "and you gain 1 life" is a separate printed clause, not "the life lost this way"), Blood Artist, Zulaport Cutthroat, Sanguine Bond, Exquisite Blood, Phyrexian Arena, Thoughtseize, Night's Whisper, Reanimate, Feed the Swarm, Deadly Tempest, Wound Reflection, Dark Confidant, the tutors, `damage_tail.go`'s CR 702.15b lifelink credit, and the `ChangePlayerLife` sandbox verb. They start a life change and are done with it, which is exactly what the unchanged entry point is for.

### Out of scope, noticed

`effects/batch14_helpers.go:252` `b14DamageEachOpponentGainThatMuch` (**Creeping Bloodsucker**) has the identical shape for **damage**: it reads each opponent's life before and after `DealDamageToPlayerForEffect` and gains the difference, so a damage event that pauses on a CR 616 prompt makes the gain short. Damage has no caller-facing `Then` today and adding one is a second engine change, so it is filed on its own as #807 rather than smuggled in here.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...`
- [x] `AISEAT_GAME_TESTS=1 go test ./internal/...`
- [x] `make lint` (golangci-lint v1.64.8, as in CI)
- [x] `go test ./internal/cards/coverage/` — no completeness or clone-group change, so no `-update` and no census edit in `docs/decklists/card-coverage-roadmap.md`
- [x] `go test ./internal/game/ -run "Drift|Snapshot|Clone"` — nothing new is snapshotted (`lifeTail` and `mustSettleNow` ride the `ReplacementEvent` on the existing `replacementResume` frame, already counted by `ContinuationCensus.ChoiceResumeFrames` and already `dropped` in `snapshot_drift_test.go`)

New tests:

- `server/internal/game/life_continuation_test.go` — the continuation runs with the **replaced** amount; it waits for a two-effect CR 616 prompt and then runs with the post-replacement amount; it runs with `0` on a CR 614.10 cancel; the batch sums the *applied* amounts with one opponent doubled; the batch **with one paused leg and one unpaused** totals both; the batch skips seats that have left; a payment runs the life-**loss** window (CR 119.4) and is invisible to a life-**gain** replacement; a payment never queues a prompt and never fires a replacement that would ask a question; a payment that cannot be made is refused and pays nothing (CR 119.4); **undo across the pause** — rewinding to before the drain drops prompt and continuation together, and rewinding *into the open prompt* and answering again produces the same total and the same life totals.
- `server/internal/cards/effects/life_continuation_cards_test.go` — Exsanguinate for 3 with two different life-loss replacements on one opponent **and Rhox Faithmender** on the caster: the drain pauses, nothing moves until the prompt is answered, and the caster then gains the true 14 doubled to 28. Gray Merchant against an opponent whose life total can't change gains 4, not 2 × 3 opponents. (Alhammarret's Archive is not in the catalog — the Faithmender is the only card in it with a `RepEventLife` replacement and two Faithmenders collapse under #792 — so the second effect of the mixed window is registered for the test, and the file says so.)
- `server/internal/cards/effects/life_continuation_guard_test.go` — the catalog lint, plus a fixture test proving the lint still sees the pre-#793 shape and passes the continuation.


krakenhavoc:
Post-merge review of this PR found two bugs in the continuation contract, plus a few smaller items. Neither bug can happen with the current catalog, but both will once a second life replacement lands. Details, repros and fixes are in #808.

Issue #742, opened 2026-09-17
Choose a color (CR 105.4): a choose_color prompt with a stored colour, a resolution-time form, and one pick for "N mana of any one color"


krakenhavoc:
## Why

Nothing lets a player choose a color. Evidence on develop (`ad9413f`):

- `git grep` for `ChooseColor`, `ChosenColor`, `choose_color` or "chosen color" in `server/` finds nothing.
- The only colour picker is the mana pick, `PendingChoiceMana` (`game/pending_choice.go:303-309`, `ColorOptions`). `ResolveManaChoice` (`pending_choice.go:793`) adds **one** `ManaToken` and has no continuation and nothing it stores.
- `resolve_choice` routes by payload **presence**: any `color` field goes straight to `ResolveManaChoice` (`actions/actions.go:1123-1124`). A new kind can't reuse that field without routing by kind.
- "N mana of any one color" can't be expressed. Each `{W|U|B|R|G}` slot queues its own pick. ADR 0040 (`docs/decisions/0040-mana-pipeline.md:213`) and ADR 0037 (`0037-unimplemented-card-signal.md:180`, Lotus Field #332) record the gap.

The pattern to copy is S26's creature-type choice, which shipped without an ADR:
- `game/creature_type_choice.go`: `PendingChoiceCreatureType` (`:42`), `QueueCreatureTypeChoiceForEffect` (`:54`), `ResolveCreatureTypeChoice` (`:81`) writing `Card.NamedTribe`, and `EventCreatureTypeChosen` (`:47`).
- `NamedTribe` is saved in `snapshot.go:299`, and cleared on zone change (`zone.go:197`) and on exile return (`effect_api.go:2031`).
- Its answers are in `legal/choices.go:34-38`, and the client branch is in `ChoicePromptModal.svelte:204-227`.
- Effects read it through `cards/effects/tribal.go`.

The resolution-time form needs a continuation, like `confirmFrame` / `chooseCardsFrame` (`game/chained_choice.go:120`, `:141`), which `ContinuationCensus` counts (`snapshot.go:443`).

## Cards

26 cards where this is the only core blocker, 31 with it as any core blocker. The verifier found 3 of 25 sampled cards (12%) also blocked by something else (≈23 real). Another 3 need only the "N mana of one color" amount, so leaving the amount out makes it 24% (≈20).

[...]

PR #836, opened 2026-09-17
feat(engine): "any color" mana offers all five colours, commander identity first

krakenhavoc:
## Summary

Owner decision (2026-09-17): an "any color" mana source offers **all five colours**, with the commander's colour identity listed first. Nothing is narrowed away. Before this, the engine narrowed every multi-option ("pipe") slot to the commander's identity unless the card opted out, so Birds of Paradise, Treasure, Lotus Petal, Coalition Relic, Ornithopter of Paradise, Cultivator's Caravan, Dragonstorm Globe, Decanter of Endless Water and Phyrexian Altar all showed a mono-green deck a single `{G}` button. That is narrower than the printed cards.

**Engine**
- `game.manaPickOptionsFor` is now the one option list that `ActivateManaAbility`, `AddManaForEffect` and the auto-tapper (planner and executor) all read. By default it keeps the printed set and moves identity colours to the front, keeping printed order within each group. Under a Golgari commander, Birds offers `B G W U R`. With no identity, the printed order is left alone.
- `NarrowToCommanderIdentity` replaces `IgnoreCommanderIdentity` as its inverse, on `ManaAbilityShape`, `effects.ManaAbility`, `AddManaOptions` and `effects.AddMana`. Every `IgnoreCommanderIdentity: true` line in the catalog is gone. Only **Command Tower, Arcane Signet, Commander's Sphere and Path of Ancestry** set the new flag. The intersection keeps its fallback: if nothing overlaps, the full set is offered.
- **The identity is read wherever the commander is (review fix).** `commanderIdentityFor(g, p)` used to read only the command zone. After the commander was cast, the identity came back empty: Birds went back to WUBRG order, and Command Tower, Arcane Signet, Commander's Sphere and Path of Ancestry offered all five colours. It now collects every card the player owns with `IsCommander`, in the command zone, on the stack, on the battlefield, in the graveyard, in exile, in hand or in the library, and returns the union of their identities, so partners combine (CR 903.4a). A stolen commander still counts for its owner. `filterPipeByCommanderIdentity` is deleted, and its tests call `manaPickOptionsFor(g, ..., true)` directly.
- **Auto-tap:** a slot that only pays generic mana still picks an identity colour, because `pickColorForSlot` falls back to `options[0]`. A coloured cost outside the identity can now be paid from Birds, as printed.
- **Auto-tap source order under an identity:** Birds, Treasure and the Signet-style rocks used to narrow to one option under a mono-colour commander. They scored like a basic (`restrictivenessScore` 1, `tierForGeneric` 2). They now keep five options (score 5, tier 1), the same as in a game with no commander. For generic costs the solver taps them before basics and saves the basics for coloured pips.
- **Bots:** the legal enumerator keeps the server's order, and the heuristic breaks ties on the lowest index. With no colour need from its hand, the bot takes the identity colour.

**Scope note:** two- and three-colour lands (guildgates, shocks, temples, tri-lands, original duals) used to narrow too. They now offer both or all three printed colours, identity first. This follows the same rule: print width, identity first. The printed cards say "Add {B} or {R}", so this matches the printed text. In a legal deck nothing changes: CR 903.5c keeps an off-identity land out of the deck, so the old intersection was already a no-op. Only a land whose colours fall outside its controller's identity (a stolen land, for example) now offers its full printed set. Neither auto-tap nor the bots have a new decision to make there.

**Catalog audit.** A temporary test listed every spec with a pipe mana ability and no opt-out. A Python scan of the Scryfall dump found that only four catalog cards print "commander's color identity". Stale comments on about 30 cards now describe the new behaviour, including Treasure/Gold, Dragonstorm Globe, Cultivator's Caravan, Decanter, Phyrexian Altar, Ornithopter, the tri-lands and original duals, Mana Confluence, Lotus Cobra, Deathrite Shaman and Sanctum of Fruitful Harvest. No completeness declaration or caveat string mentioned the narrowing, so none changed.

**Client:** a new `colorButtons` helper (`client/src/lib/manaPick.ts`) builds the picker's buttons in the server's order, without re-sorting. `ChoicePromptModal` uses it for `mana_pick` and `choose_color`. The wire shape is unchanged.

**Docs:** ADR 0040 has a dated addendum and the superseded bullet is struck through. AGENTS.md mana-ability notes are updated. protocol.md documents the `color_options` order. protocol.ts has an updated comment.

## Tests
- `effects`: `TestAnyColorManaOffersAllFiveIdentityFirst` checks that a mono-green commander gets `G W U B R` from Birds, Lotus Petal, Coalition Relic, Ornithopter, Caravan, Dragonstorm Globe, Decanter, Treasure and Phyrexian Altar, and that an off-identity answer is legal. `TestCommanderIdentityTextStillNarrows` covers the four identity cards. `TestDualLandOffersBothColoursIdentityFirst` covers a two-colour land. `TestOnlyCommanderIdentityCardsNarrow` pins the list of narrowing cards. The dump-gated `TestNarrowToCommanderIdentityMatchesOracleText` checks that a card narrows exactly when its oracle text has the clause. Existing tests were updated for the new order.
- `game`: `TestManaPickOptions` is a table test. `TestAutoTapAnyColorPrefersCommanderIdentity` checks that a generic slot mints `{G}` and that `{W}` can be paid from Birds.
- `game` (review fix): `TestCommanderIdentityHoldsOnTheBattlefield` casts a `{1}{G}` commander from the command zone and checks the identity on the stack. With the commander resolved onto the battlefield, it checks that Birds offers `G W U B R`, Command Tower offers `[G]` (both on activation and in the planner), effect-added mana orders `G` first, and auto-tap for `{1}` from Birds mints `{G}`. `TestCommanderIdentityEveryZone` walks the commander through the battlefield (including under an opponent's control), graveyard, exile, hand and library, and checks that the opponent's identity is unaffected. `TestCommanderIdentityUnionsPartners` covers partners. `TestAutoTapBirdsPaysOffIdentityPipBesideABasic` pays `{W}{B}` and `{B}{W}` from Birds plus a Plains under a mono-W commander, which exercises the greedy executor. Before the fix, with the lookup limited to the command zone, all three identity tests fail.
- `legal`: the mana_pick answers list the identity colour first. `heuristic`: with an empty hand, the bot picks the identity colour.
- Client: `manaPick.test.ts` checks that the picker renders five buttons with G first.
- gofmt, `go vet ./...`, and `go test ./...` with the dump, `AISEAT_GAME_TESTS=1` and `AISEAT_CATALOG_GAMES=3` (catalog soak passes). `make lint` (golangci-lint 1.64.8). Client: `npm run check`, `npx vitest run` (845 passing), `npm run lint`.
- After the review fix and the merge of `origin/develop`: gofmt and `go vet ./...` are clean, and `go test ./...` passes with the dump, `AISEAT_GAME_TESTS=1` and `AISEAT_CATALOG_GAMES=4` (`TestCatalogSoak` and the dump-gated `TestNarrowToCommanderIdentityMatchesOracleText` pass). `make lint` (golangci-lint 1.64.8) is clean. The fix touches no client code.

Refs #843 (tracking issue for the owner decision; #742 is closed). Follow-up: #844. CR 903.4f says a "commander's color identity" source adds nothing when there is no identity. That behaviour predates this PR, and `TestManaPickOptions` now labels the case as current behaviour, not intended.


krakenhavoc:
Owner decision (2026-09-17): a manual tap should produce the colour in one click when exactly one of the source's printed colours is in the controller's commander identity (a Scrubland in mono-white gives {W}); every printed colour stays available from the menu. Sources with 2+ identity colours, or none, prompt as before. The auto-tapper and bots are unchanged.

Work in progress for that is pushed to branch `wip/836-one-click-default` (4984edfe): server, client and docs changes are in place, but the tests written for the always-prompt behaviour still expect a prompt and need rewriting, and the client vitest is missing. This PR stays open until that is finished.

krakenhavoc:
Correction to the comment above: this PR is already merged (1e18f33), so the all-five behaviour is on `develop`. The one-click identity default is therefore a follow-up: branch `wip/836-one-click-default` (4984edfe) holds the server, client and docs work, and needs its tests rewritten for the new behaviour (they still expect a prompt) plus the client vitest before it becomes its own PR.

Issue #837, opened 2026-09-17
Bot decision harness and eval loop for the local model (qwen3:14b) — ADR 0052


krakenhavoc:
## Why

The `assisted`/`strong` tiers were built and tested entirely against `model/fake.go`. Nothing measures play quality, and the local model path (Ollama, `qwen3:14b`, 16 GB card) has never been observed with real telemetry. Two transport-level problems probably make the local bot much worse than the design intends, and both are cheap to confirm:

- **Silent truncation.** Ollama defaults to a 4k context below 24 GiB VRAM and drops the *front* of an over-long prompt without an error. The static system block is the rules primer plus a 100-card decklist with oracle text (likely 6–10k tokens), so the instructions are the first thing lost. The `/v1` endpoint has no field for `num_ctx`.
- **Thinking not suppressed.** The transport sends Ollama's native `think:false` on `/v1/chat/completions`; Ollama's docs say that endpoint takes `reasoning_effort` (`"none"`), not `think`. If `think` is ignored, qwen3 thinks on every call and the 128/256 `max_tokens` cap cuts the answer off → `FallbackMalformed` on every window → the seat plays the heuristic under the `assisted` label.

Beyond that, the prompt is thin for a 14B model: no oracle text outside the bot's own decklist, no game history (although `GameView.Log` is on the wire), no combat framing, pending-choice options never rendered, no structured output, no temperature, one budget for every window.

## Decision (2026-09-17)

Build **both** an eval harness and a rebuilt decision harness, **eval first**, one lever per PR, every PR after the baseline carries suite + arena numbers. Target: qwen3:14b on the existing box. Reasoning time per decision is fine but must be **adjustable at the table** (players/admin can lower it mid-game). Success = win rate vs `heuristic`, agreement on a labelled position suite, fewer obvious blunders (reviewable decision logs).

## Scope

**ADR 0052** records the design (merged, #840). Then, in order:

- [x] PR 1 — decision trace (`aiseat.Tracer`), runner observer, per-game JSONL decision log (`CMDCTRL_BOT_DECISION_LOG`), whole-decision latency percentiles on `aiseat.Stats` (#505 part 1), `boteval probe` (tests the two transport hypotheses against the real endpoint) — **#842**
- [ ] PR 2 — labelled position suite (`aiseat/suite`, ungated heuristic regression gate) + `boteval suite run|harvest|render`, ~20 seed positions — **#850**, reviewed and awaiting merge
[...]

Issue #505, opened 2026-09-14
S31: per-decision latency percentiles are never computed and no admin surface can read the bot metrics


krakenhavoc:
Issue #89's Latency-discipline checklist asks for:

> - [ ] Per-decision p50/p99/p999 latency instrumentation **in admin tools**

Half of it shipped. The other half — the half a human can read — did not, and the collected half is not the metric that was asked for.

## What exists

`server/internal/aiseat/model/metrics.go` (sub-PR 7) is good instrumentation: a bounded ring of `DecisionRecord` (layer, rule, escalations, model id, `Latency`, `ModelLatency`, token usage including cache read/write, fallback cause) plus aggregate `Stats`, with `AbsorptionRate()`, `EscalationRate()` and `Triggers()` derived from it. `model.Policy.Stats()` and `.Records()` are exported. `rules.Meter.Stats()` and `aiseat.Runner.Stats()` are the same idea at their own layers.

## What is missing

**1. No percentiles anywhere in production code.** `Stats` carries `ModelLatency` (a sum) and `MaxModelLatency`. `DecisionRecord.Latency` is stored per record and never aggregated. The only p50/p99 in the repo is computed inline in a test:

```go
// server/internal/aiseat/heuristic_game_test.go:337
p50 := all[len(all)*50/100]
p99 := all[len(all)*99/100]
```

`grep -rn "p999" server/` returns nothing. So the numbers quoted in the sprint log (p50 8.5µs, p99 52µs) came from a `t.Logf`, not from anything the running server can report.

**2. Nothing is reachable from an HTTP handler.** `Manager.Runners(gameID)` exists and its doc comment literally says *"(tests, admin tools)"* — but `lobby.BotHost` narrows `*aiseat.Manager` to `Tiers()` / `StartBots()` / `StopBots()`, so no handler can call it. And `Runner` has no accessor for its policy, so even with `Runners()` exposed, `model.Policy.Stats()` stays out of reach. There is no `/metrics`, no `/debug`, no expvar and no pprof on this server; the entire observability HTTP surface is `GET /healthz` returning the string `ok`.

The upshot: a decision-latency regression on a live table is invisible. Collecting a metric nobody can read is half the feature.

## Suggested shape

Three pieces, all small, but the first two are inside `internal/aiseat`:

[...]

PR #842, opened 2026-09-17
feat(aiseat): decision trace, per-game decision log, runner latency percentiles, and boteval probe (#837)

krakenhavoc:
## Summary

- **Decision trace.** `aiseat/trace.go` adds `Trace`/`Candidate`/`Prompt`/`TokenUsage` as plain JSON data and the optional `Tracer` extension (the same shape as `Conceder`). `model.Policy`, `rules.Filter` and `heuristic.Policy` implement it, with compile-time assertions; `model.Policy.BuildRequest` assembles the exact call a window would make and makes none. `aiseat.Input` gains JSON tags — that is what makes a recorded window replayable.
- **Runner observer + latency percentiles (#505 part 1).** `decide()` returns an outcome struct instead of `(int, string)`, so "the policy chose the pass" and "the policy timed out and the runner took the pass" stop being the same fact. One `DecisionEvent` per window after the `room.Apply` result is known, carrying the runner's own fallback cause (`timeout` / `policy-error` / `out-of-range` / `decline-pass` / `forced-pass` / `forced-always-legal`). A 1024-entry ring of whole-decision durations gives `Stats.Latency` (`Percentiles{Count,P50,P99,P999,Max}`, nearest rank); `Stats` gets JSON tags and `Rejection.Err` now marshals as a string instead of `{}`.
- **`aiseat/decisionlog`.** One JSONL file per game, mutex-serialised across the seat goroutines, `0700` directory / `0600` files, a 256 MiB per-game cap with a single WARN, three fullness modes (`escalated` default / `all` / `model`), and `Scan` with a 64 MiB token cap that also reads `.jsonl.gz`. Wired through `Manager.SetDecisionLogger`; the log is closed by `StopBots` **and** by a watcher on every runner's `Done()`, because a game that simply ends is never stopped. `CMDCTRL_BOT_DECISION_LOG[_MODE]` in `loadConfig` (off by default; an invalid mode fails the boot like `CMDCTRL_BOT_MAX_THINK`), plus the `AISEAT_DECISION_LOG` test knob in `playGameIn`.
- **`cmd/boteval probe`.** One real request in the funnel's exact shape against a configured endpoint, printing endpoint/model, system+user bytes and a token estimate, `prompt_tokens`/`completion_tokens`, the server's prefix-cache hits, `finish_reason`, whether the reply was empty, whether a reasoning field came back, the first 300 chars, the parsed index or the parse error, and wall time — then a verdict line for **TRUNCATION** and for **THINKING**. Always exits 0. `make -C server build-boteval`.
- **Rebased onto `develop` after #839 landed.** That PR shipped the sending-side fix (`reasoning_effort:"none"` beside `think:false`) and the same response-side `reasoning` / `reasoning_content` decoding this branch had. Conflicts were resolved keeping #839's behaviour and a single copy of the decoding and its tests (`TestOpenAIRequestTurnsThinkingOffTheWayOllamaHonours`, `TestOpenAIResponseReasoningIsSurfaced`); my duplicate was deleted. **The only transport change left here is one decode**: `usage.prompt_tokens_details.cached_tokens` → a new `Usage.CachedPromptTokens`, kept distinct from `CacheReadTokens` (Anthropic's *billed* cache-breakpoint read, still zero on this endpoint), with its own test. Nothing about what is **sent** changes.
- The `THINKING` verdict stays, and should now read **suppressed** — the probe goes through the ordinary `OpenAIClient`, so both thinking-off fields travel with it, and its test asserts exactly that rather than the old "sends no `reasoning_effort`". It is kept because a server that honours neither switch is invisible from the table: the seat still plays, it just plays Layer B under a model tier's name.
- **ADR 0052's decision-log size estimate is corrected with the measured numbers** (15–20 MiB/game → 38–59 KiB per window, 114–236 MiB for a four-seat game in `escalated`), and the 256 MiB cap is re-described as a real limit rather than a theoretical one.
- **Privacy.** The decision log is operator-only and off by default: each record is the seat's own filtered view, but the file aggregates every bot seat at one table. Never served over HTTP, never attached to a bug report. Written into `docs/bot.md` (new "Decision log" section, plus a "probe" paragraph under "The model endpoint") and `AGENTS.md`.

## Sprint

S31 — AI bot seat (follow-up)

## Issues

Relates to #837; part 1 of #505.

## Test plan

All run via Docker from the worktree root (no local Go toolchain).

- `docker run --rm -v "$PWD":/src -v gomodcache:/go/pkg/mod -v gobuildcache:/root/.cache/go-build -w /src/server golang:1.22 sh -c 'gofmt -s -l . && go vet ./... && go test -race ./internal/aiseat/... ./internal/legal/... ./internal/docsguard/... ./cmd/...'` → **exit 0**, all packages `ok`.
- Full suite: same image with `go test -race ./...` → **exit 0**, 34 packages `ok` / no-test-files, 0 failures.
- Gated whole-game tests with the decision log on:
  `docker run --rm -e AISEAT_GAME_TESTS=1 -e AISEAT_DECISION_LOG=/dl -v "$PWD":/src -v /tmp/dl:/dl … golang:1.22 sh -c 'go test -race -timeout 20m -run "TestDecisionLogReplaysOffline|TestFourHeuristicBotsPlayToAWinner" -v ./internal/aiseat/'` → **exit 0**.
  - `TestDecisionLogReplaysOffline` (2-seat `heuristic` tier, `all` mode): **569 lines, 569 replayed (Layer A = 479, Layer B = 90)**, every line parses, `rules.Resolve` reproduces every recorded Layer A verdict and index, `heuristic.New().Decide` reproduces every recorded `HeuristicIndex`. 21.6 MB, ~38 KiB/record, 11 turns.
  - `TestFourHeuristicBotsPlayToAWinner`: 3 seeds, all pass, logs of 2222 / 3965 / 3501 windows.
- Lint: `docker run --rm -e GOTOOLCHAIN=auto … golang:1.24 make lint` → **exit 0**, clean.
- The probe was **not** run against the dev box from here; its verdict logic, the request it sends (`think:false` **and** `reasoning_effort:"none"`) and its reading of the measured qwen3 reply shape (empty `content`, `finish_reason: "length"`, thinking in `message.reasoning`, `prompt_tokens_details.cached_tokens`) are covered by `httptest` unit tests, and `representativeInput` is exercised by a test that plays a real game and builds the prompt from it.
- All of the above was re-run after the rebase onto `develop` (`fd01ecf1`): fmt/vet + `go test -race ./internal/aiseat/... ./cmd/...` → exit 0; full suite → exit 0, 34 packages; `make lint` → exit 0.

**Run the probe with a dump.** Without `--dump` / `CMDCTRL_SCRYFALL_DUMP` the static block is the deck name alone (~1.5 KB) and the truncation verdict is measuring a prompt an order of magnitude smaller than a real seat sends.

### Measured record size

| Mode | Seats | Bytes/record |
|---|---|---|
| `all` | 2 | ~38 KiB |
| default (`escalated`) | 4 | ~51–59 KiB |

A four-player board view is most of a record, so a long four-seat game writes 100–250 MiB. Worth knowing before turning it on: `escalated` only saves anything for a policy that runs Layer A (the `heuristic` **tier**, i.e. `rules.NewFilter(heuristic.New(), …)`); the bare `heuristic.New()` the whole-game tests seat reports every window as Layer B and keeps them all in full. Noted in AGENTS.md.


krakenhavoc:
Review follow-ups are in `1553645e`. All nine, plus the one declined with a comment.

| # | Change |
|---|---|
| 1 | `runner.step` emits on both paths that dropped the event after `decide()` had already banked the latency: the forced branch with no pass and no always-legal move (`Forced = no-legal-answer`) and the `ctx.Err()` check after pacing (`Forced = cancelled`), both `Applied=false`. |
| 2 | `Fallback` and `Forced` are now two axes rather than one field. `Fallback` keeps `timeout` / `policy-error` / `out-of-range` / `decline-pass`; `Forced` carries `forced-pass` / `forced-always-legal` / `no-legal-answer` / `cancelled`. Both reach `DecisionEvent` and `Record.Final`. A window that timed out and was *then* forced onto the always-legal answer reports both — the pair that explains a fail-to-find. |
| 3 | `truncationVerdict` defers to thinking evidence: an unparseable reply is not truncation evidence when `Reasoning != ""` or `finish_reason == "length"` with empty content. A genuinely short `prompt_tokens` still wins over both. |
| 4 | `CMDCTRL_BOT_DECISION_LOG_MODE` fails the boot only when `CMDCTRL_BOT_DECISION_LOG` is set; otherwise it warns and takes the default. |
| 5 | `rules.Filter.DecideTraced` over a non-`Tracer` inner policy no longer claims Layer `"B"` and an index — it names the inner policy (`"random"` for `*RandomPolicy`) and leaves `HeuristicIndex` at `Decline`. |
| 6 | `GameLog` writes from one goroutine behind a bounded queue. `Observe` marshals on the caller's goroutine and does a non-blocking send; a full queue drops and counts like the byte cap, with `DroppedQueue` / `DroppedCap` / `DroppedError` broken out in `Stats`. `Close` drains, then flushes and closes. |
| 7 | The static system block is written **once per file** and carried by `SystemHash` on every later record. `Scan` is unchanged; the package doc says how a reader reconstructs it. |
| 8 | `model.Truncate` / `model.OneLine` are exported and the probe uses them instead of a local byte-slice that could split a rune. |
| 9 | `Manager.StartBots` opens the decision log **before** taking `m.mu` — no disk I/O under the lock every other `StartBots`/`StopBots` queues behind. |

**Declined, as agreed:** `heuristic.DecideTraced` prices the window twice. The funnel already does exactly that on every Layer B window in production (`Rank` then `Fallback.Decide`), and both passes are microseconds against a seconds-long deadline. There is now a comment on `DecideTraced` saying so.

**Queue depth is 256, not 64.** `MaxActionsPerWake` is 64 and four seats can be woken by one commit, so 256 back-to-back windows is the burst a real table reaches; a 64-deep queue lost a quarter of the records in the concurrency test. The bound is ~15 MiB in `all` mode on a four-player board.

**One deviation worth naming.** Review item 1 suggested `no-legal-answer` and `cancelled` as *fallback* labels. They are on the new `Forced` axis instead, because both are the runner overriding a window it had already decided — putting them in `Fallback` would clobber the `timeout` that item 2 asked to preserve. `Fallback` answers "why was the policy's answer not used", `Forced` answers "what happened instead".

### Tests

New: `forced_internal_test.go` (all three `forcedAnswer` cases), `TestRunnerObserverReportsAWindowCancelledDuringPacing`, `TestRunnerObserverKeepsTheFallbackWhenItForcesAnAnswer` (real rejections via the Myriad Landscape scenario), `TestFilterOverAnUntraceablePolicyNamesIt`, `TestRecordKeepsBothFallbackAndForced`, `TestSystemBlockIsWrittenOncePerFile`, `TestSystemHashIsEmptyWithoutAPrompt`, `TestQueueOverflowDropsAndCountsRatherThanBlocking`, `TestCloseDrainsTheQueue`, `TestThinkingEvidencePreemptsTheTruncationVerdict`, three `loadConfig` cases, and `TestOpenAIUsageReportsThePrefixCacheHits`.

`TestProbeReadsTheQwenReplyOffTheWire` now asserts **both** verdicts on the measured shape: `TRUNCATION: not detected` + `THINKING: not suppressed`.

The `no-legal-answer` branch is covered by unit test rather than end to end, and the test says why: it needs the enumerator and the engine to disagree in a window offering neither a pass nor an unconditional answer, and the enumerator's contract is that everything it offers is legal when offered — so it is not constructible from a real table. The other two forced cases are covered end to end.

`TestObserveIsSafeFromEverySeatAtOnce` now asserts what it was always about — every record that reached the file is one whole line, and every record that did not is accounted for.

### Verification (Docker)

- fmt + vet + `go test -race ./internal/aiseat/... ./internal/legal/... ./internal/docsguard/... ./cmd/...` → exit 0.
- Full `go test -race ./...` → exit 0, 34 packages, 0 failures.
- `make lint` (golang:1.24) → exit 0.
- Gated `TestDecisionLogReplaysOffline` → 594 windows, 594 replayed (A=509, B=85), **0 dropped**.
- Gated `TestFourHeuristicBotsPlayToAWinner` with the log on, three four-seat games: 2171 / 3964 / 3498 records, **0 dropped** by queue, cap or error — and the wall clock went **85s → 65s**, which is the synchronous-disk-on-the-seat-goroutine cost item 6 was pointing at, measured.

Issue #796, opened 2026-09-17
A resolving ability needs a free yes/no for its own controller at resolution ("you may …" inside an effect) — Eden and the reflexive-trigger tail


irobinson010:
Noticed while building the reflexive-trigger seam (#636, PR #795).

Triggers have a CR 603.5 "you may" prompt before they go on the stack (`OptionalPrompt`), and search / scry / put-from-hand primitives carry their own decline. What does not exist is a plain yes/no a **resolving** effect can ask its **own controller** mid-resolution and continue from — the shape of "you may [do X]. If you do, [Y]" where X is not a search or a cost the engine already prompts for. Eden is the card that hit it; the registry's "Opponent's non-mana choice at resolution" row is the opponent-facing sibling.

**Fix shape.** A small `MayChoice{Question, OnYes, OnNo}` primitive over the existing `PendingChoiceTriggerPrompt`-style yes/no (or `ChooseCardsPrompt` with count 0..0 if that is cleaner), addressed to `item.Controller`, with the two continuations as package-level effects reading the item (undo-safe, like delayed triggers). It is gated by #791 like every other prompt. Add the row to `docs/engine-seams.md` (Open) with Eden and any other card the batch reports named for it.

irobinson010:
Claimed (2026-09-18): an agent has this together with #568 on branch `feat/796-568-resolution-choices` off `develop` (one PR, separate commits). Please don't pick it up in parallel.

irobinson010:
PR #918 (branch \`feat/796-568-resolution-choices\`, against \`develop\`) implements this as \`effects.MayChoice{Player, Question, YesLabel, NoLabel, LifeCost, OnYes, OnNo}\` over the existing \`confirm\` kind — no new \`PendingChoiceKind\` — and converts Eden, Seat of the Sanctum (now \`full\`), Mask of Memory and Crossway Troublemakers.

irobinson010:
Done in #918, merged to `develop`: `effects.MayChoice{Player, Question, LifeCost, OnYes, OnNo}` on the existing confirm kind (no new kind), addressed to the controller by default; Eden, Mask of Memory and Crossway Troublemakers converted, Eden and Mask now full.

Issue #568, opened 2026-09-14
An opponent's non-mana choice at resolution — Fact or Fiction and 5 more


krakenhavoc:
Split out of #74 (S22), which closes with this deferred. Also a named cluster in `docs/decklists/card-coverage-roadmap.md`.

A spell that asks an **opponent** to make a choice during its resolution. The catalog can prompt the controller (#552 shipped chained prompts, #549 shipped the broadcast reveal) but not another seat.

**Cards blocked:** Fact or Fiction (oracle_id `437b2dab-15e0-4b9a-a204-58622d37a3b3`), Torment of Hailfire, Combustible Gearhulk, Charismatic Conqueror, Painful Quandary, Chain of Vapor.

**The extra work beyond #552's machinery**, called out by its author as deliberately fenced off: a choice over a zone the chooser does not own needs a **redaction pass**. Fact or Fiction reveals five cards and an opponent splits them into two piles — the reveal makes them public (#549 handles that), but the general shape does not, and `PendingChoiceView.Options` shipping cards a viewer should not see is exactly the leak PR #513 fixed.

Fact or Fiction is the card a draw player would actually notice missing.

irobinson010:
Claimed (2026-09-18): an agent has this together with #796 on branch `feat/796-568-resolution-choices` off `develop` (one PR, separate commits). Please don't pick it up in parallel.

irobinson010:
PR #918 (branch \`feat/796-568-resolution-choices\`, against \`develop\`) implements this as one new kind — \`PendingChoiceOptionPick\` (\`option_pick\`) — plus a pile split built from two chained prompts and one redaction function (\`protocol.redactChoiceCards\`); Fact or Fiction, Torment of Hailfire, Combustible Gearhulk, Charismatic Conqueror and Painful Quandary are converted, Chain of Vapor is left with a note (a spell cannot copy itself from inside its own resolution yet).

irobinson010:
Follow-up filed: #920 (a resolving spell cannot copy itself — the one gap left for Chain of Vapor).

irobinson010:
[...]

Issue #883, opened 2026-09-18
S40 — S30's tail: protection, regeneration, emblems, winning by effect


krakenhavoc:
**Phase:** 7 · **Goal:** close the object kinds and keywords S30 deferred

Planned 2026-09-18 as part of the S36–S49 backlog pass over the 210 open issues. Milestone: `S40 — S30's tail: protection, regeneration, emblems, winning by effect`.

> **Verification status.** This tracker was built from issue titles, labels, bodies and `docs/engine-seams.md`, not from a per-issue re-check against `origin/develop`. Unlike [#734](https://github.com/krakenhavoc/cmd_and_ctrl/issues/734), no member issue here carries a "still reproducible on `develop`" verdict. Re-verify before starting: the repo merges to `develop` while the default branch is `main`, so `Closes #N` never fires and shipped work can sit open.

## Members (8)

- [ ] #662 — ADR then implementation: protection (CR 702.16) — DEBT checked against the source object; supersedes ADR 0038 §7
- [ ] #667 — Regeneration (CR 701.19) — Asceticism and the regeneration batch skips
- [ ] #748 — ADR then implementation: infect, wither and toxic (CR 702.90, 702.80, 702.164) — damage results as -1/-1 counters and poison, in the one damage tail
- [ ] #749 — ADR then implementation: winning the game by effect (CR 104.2b), "can't lose / can't win the game", and gates on the loss checks
- [ ] #623 — ADR: emblems (CR 114) — representation, persistence, wire and client
- [ ] #626 — Invasion of New Phyrexia (S27 battle 4/4) — blocked on emblems, set-level Validate and a reflexive tap prompt
- [ ] #796 — A resolving ability needs a free yes/no for its own controller at resolution ("you may …" inside an effect) — Eden and the reflexive-trigger tail
- [ ] #568 — An opponent's non-mana choice at resolution — Fact or Fiction and 5 more

## Dependencies and sequencing

#626 depends on #623. #662's blocking half shares a function with S37's #750 - sequence S37 first or coordinate. ADR 0056 (#748) and ADR 0057 (#749) are already merged; only the implementation is pending.

## Exit criteria

1. Every member issue above is closed, or carries a named decision explaining why it is not.
2. The sprint section in `docs/sprints.md` is written and its index row added.
3. Any new engine primitive is reflected in `docs/engine-seams.md`, with the card counts it unblocks.
4. Lands on `main` with the next promotion.

[...]

Issue #636, opened 2026-09-16
Reflexive-trigger seam (CR 603.12): queue a targeted trigger from a resolving effect — blocks #626


krakenhavoc:
## Why

Invasion of New Phyrexia's back face, Teferi Akosa of Zhalfir (#626), has a −3: *"Tap any number of untapped creatures you control. When you tap one or more creatures this way, ..."*. That is a **reflexive trigger** (CR 603.12): a trigger created during resolution that fires only if the condition was met, with its **target chosen when it goes on the stack** (CR 603.3d).

The 2026-09-16 audit of the S27 follow-ups found no general mechanism for this. Cards with reflexive triggers currently fold them into the resolving effect and declare a caveat, e.g.:
- `invasion_of_tarkir.go` ~:43
- `ziatora_the_incinerator.go` ~:44-48
- the `batch29_helpers.go` reflexive helpers (Ziatora's damage)
- `batch08_helpers.go` ~:197
- Caesar, Legion's Emperor (`batch26_test.go` ~:139)

#626's body already says to give this its own issue if nothing existing fits; nothing does.

## Proposed shape (from the audit; confirm in review)

- [ ] A seam like `g.QueueReflexiveTriggerForEffect(sourceID, controller, TriggeredAbility)` that builds from the source's last known information and goes through the existing trigger dispatch. It should reuse `queueHarvestedTriggerLocked` → `PendingTriggers` (the `delayed.go` DelayedTrigger precedent: source-independent, clone-safe) together with `queuePickTargetLocked` (`triggers.go` ~:389) so the target is picked when it goes on the stack.
- [ ] The condition is evaluated by the resolving effect: queue only if the condition held (e.g. at least one creature tapped).
- [ ] A "tap any number of untapped creatures you control" choice. Likely a `ChooseCardsPrompt` with `Min: 0` over untapped creatures, possibly using #624's set-level `Validate`; confirm.
- [ ] Clone, snapshot and drift-test classification for anything new on `Game`; bot enumerator coverage for any new pending choice (the #499/#618 class: a new `PendingChoiceKind` needs a `legal` case or bot seats get an empty move list).
- [ ] Tests: the trigger fires only when the condition is met; the target is chosen on stacking, not at resolution of the parent; no target is chosen when there are no legal targets; clone/snapshot round-trip.

## Open question for the owner

[...]

PR #918, opened 2026-09-18
feat(engine): choices at resolution — a controller's "you may", and an opponent's pick or pile split with redaction (CR 608.2)

irobinson010:
## Summary

Two related seams, one branch, one commit each: a choice made at
**resolution** — first for the effect's own controller, then for
somebody else's seat.

### #796 — a free yes/no for the controller at resolution

Triggers have had the CR 603.5 "you may" since S19, and search / scry /
put-from-hand each carry their own decline. Nothing could ask the plain
"you may [do X]. If you do, [Y]" in the MIDDLE of a resolution, where X
is neither a search nor a cost the engine already prompts for.

`effects.MayChoice{Player, Question, YesLabel, NoLabel, LifeCost, OnYes,
OnNo}` is that primitive, and it adds **no prompt kind**: it is built on
`PendingChoiceConfirm` (#552), which is already a two-way question with
card-supplied continuations, already addressed by `Chooser`, already
classified in the choice gate, already enumerated for bots and already
rendered by `ChoicePromptModal`. `Player` defaults to the controller
("you may"); naming another seat is the same call, which is the half
#568 builds on. Branches take a `*Context` bound FRESH to the game the
resolver hands back, so they rewind with an undo.

### #568 — an opponent's non-mana choice at resolution

One new kind: `PendingChoiceOptionPick` (`option_pick`,
`server/internal/game/option_pick.go`) — "choose one of the following",
addressed to any seat, answered with the INDEX (`{option_index: N}`,
routed by kind because zero is both the field's zero value and the
commonest answer). An option is a label, a declared life cost, and
optionally the cards it is about.

**A pile split needs no kind at all**: it is two chained prompts to two
different seats (#552) — a `choose_cards` to the splitter with a floor
of zero (an empty pile is a legal split), then an `option_pick` back to
the controller whose two options each carry a pile.
`Game.QueuePileSplitForEffect` / `effects.PileSplit` is that composition.

Legality is enforced at **queue** time: `ResolveOptionPick` validates
the index and nothing else, the effect builds its option list out of
what the chooser can actually do (CR 608.2), and it must put a branch
that always works first — that is the one `legal.choiceMoves` marks
`AlwaysLegal`.

**Redaction is one function**, `protocol.redactChoiceCards`, applied to
`Options` and to every option's cards:

1. A viewer who is not a knower of a card does not get it — dropped,
   not sent as a back, because an instance ID out of a hidden zone is
   the correlation handle PR #513 closed.
2. The **chooser** keeps unknown cards as answerable backs only over
   their own material, or over another player's **hand**, whose size is
   already public (CR 400.2) — that is the coercive-discard case #513
   settled, and it is preserved. A pool that is neither (five cards off
   the top of a *library*) shows the chooser only what a reveal made
   public.

Rule 2 is therefore a contract on the card as well: a cross-seat prompt
over a non-hand zone must reveal first (#549), which every printed card
of this family does.

## Sprint

S40 — S30's tail: protection, regeneration, emblems, winning by effect
(tracker #883).

## Issues

Closes #796
Closes #568
Refs #883

## Cards converted

**#796 (3):**

| Card | What changed |
|---|---|
| Eden, Seat of the Sanctum | `caveats` → **`full`**. Was two activations sharing the `{5}` and the tap with the sacrifice decided up front; now one activation — mill two, then ask, and on yes the #636 reflexive trigger picks its target *after* the sacrifice, so a permanent card the mill just binned is a legal pick. Both caveats cleared. |
| Mask of Memory | `caveats` → **`full`**. The optional draw was always taken; it is a real question now, and the linked discard (CR 607.2) happens only on the branch that drew. |
| Crossway Troublemakers | One caveat cleared: the pay-2-life decision moved from the trigger's announce-time prompt to the ability's resolution, where the card prints it. Its attack-trigger caveat stands. |

**#568 (5 new files):**

| Card | Shape |
|---|---|
| Fact or Fiction | pile split (reveal 5 → opponent splits → controller takes a pile). Ships with one caveat: with more than one opponent the splitter is the next seat in turn order rather than one you choose — that needs the still-open "choose a player" prompt seam. |
| Torment of Hailfire | option pick, X repetitions × each opponent, asked strictly one at a time so a later repetition sees what an earlier one cost. Options are dropped for a player who cannot take them. |
| Combustible Gearhulk | `MayChoice` addressed to the target opponent; damage read off the cards actually milled. |
| Charismatic Conqueror | `MayChoice` addressed to the entering permanent's controller. |
| Painful Quandary | `MayChoice` addressed to the caster; an empty hand is not asked at all. |

**Left:** **Chain of Vapor**. Its copy leg is "they may copy *this*
spell", and `CopySpellForEffect` reads the source off `Game.StackMeta`,
which `resolveTopOfStackLocked` deletes before it calls `OnResolve` — a
spell cannot copy itself from inside its own resolution today. Noted in
`docs/engine-seams.md`. Braids, Arisen Nightmare, Tempt with Bunnies and
Druid of Purification are unblocked by this seam but still need their
own files and their own gap check.

## Bots

`docs/bot.md` gains a section. `confirm` (so `MayChoice`) keeps its
existing policy — both branches offered, the accept branch priced by
`cost.life`, the decline always-legal. `option_pick` offers every branch
in printed order, each carrying its declared life cost, with the FIRST
marked always-legal by the kind's contract, so a bot-only table
terminates. A pile split needs no policy of its own: its first half is
an ordinary `choose_cards` over public revealed cards (empty pile =
always-legal answer) and its second half is the option pick.

## Docs

- `docs/decisions/0018-triggers-on-the-stack.md` — dated amendment
  (2026-09-18, #796 / #568): the resolution-time kinds, `Player`
  addressing, and the redaction rule. No new ADR: the redaction rule is
  an application of #513's existing decision to a new field, not a new
  decision.
- `docs/engine-seams.md` — "Opponent's non-mana choice at resolution"
  moved from Open to Closed with its card counts; a second Closed entry
  for #796's free yes/no; the #636 entry's "Eden is still waiting" line
  corrected.
- `docs/protocol.md` — a `PendingChoiceView — #568 additions` bullet
  (`option_pick`, `pick_options`, `option_index`, and the redaction
  rule a client must expect).
- `AGENTS.md` §7 — a "Choices made at resolution" section, one
  paragraph per primitive.
- `docs/decklists/card-coverage-roadmap.md` census regenerated
  (`go test ./internal/cards/coverage/ -update`).

## Test plan

- [x] `cd server && go build ./... && go vet ./... && go test ./...` — clean
- [x] `AISEAT_GAME_TESTS=1 go test ./internal/...` — clean
- [x] `cd server && make lint` (golangci-lint v1.64.8 via `go run`) — clean
- [x] `cd client && npm run check` — 478 files, 0 errors
- [x] `cd client && npm run lint` (eslint + prettier) — clean
- [x] `cd client && npm test` — 58 files, 979 tests
- [x] `go test ./internal/cards/coverage/ -update`; census block in the
      roadmap doc regenerated (not hand-merged) after the rebase
- [x] Engine: `MayChoice` runs exactly one branch, in order, only after
      the answer; can be addressed to another seat and only that seat
      may answer; blocks the table; undoes across the prompt and the
      restored question answers the same way
- [x] Engine: an option pick addressed to an opponent runs the chosen
      branch, refuses another seat's answer and an out-of-range index,
      and undoes with its options intact
- [x] Engine: a pile split chains to the controller's pick, returns the
      piles in split order, and falls back to one pile when the
      splitter has left (CR 800.4a)
- [x] Protocol: a prompt to seat B over seat A's **revealed** cards
      shows exactly those; over **unrevealed** cards shows none; a
      bystander sees the prompt and no hidden cards; the coercive
      discard still hands its chooser backs. Verified failing with the
      cross-seat clause backed out.
- [x] `legal`: every branch enumerated for its chooser and nothing for
      any other seat; the life branch priced, the first branch
      always-legal; `TestEveryChoiceKindIsClassifiedAndEnumerated` passes
- [x] Cards: Fact or Fiction end to end (including an empty pile and an
      undo across both links); both branches of Combustible Gearhulk,
      Charismatic Conqueror and Painful Quandary; Torment's three
      branches and its option list shrinking for a player who cannot
      take them; Eden, Mask of Memory and Crossway Troublemakers each
      get their decline path
- [x] Client: 4 render tests on the new modal arm — one button per
      branch in printed order, `option_index` submitted (including 0),
      not the card-grid fallback, and an option whose cards were
      redacted away stays live

## Out of scope

- `server/internal/game/spell_copy.go:142` (`stackSpellLocked`) +
  `server/internal/game/mutations.go:1685` (`delete(g.StackMeta, …)`
  before `fireEffectResolverLocked`): a resolving spell is unreachable
  by `CopySpellForEffect`, which is what holds Chain of Vapor. Not a
  one-liner on this path.
- `server/internal/protocol/view.go:2770` — a `choose_cards` prompt
  drops its `Options` and bounds for every non-chooser, which
  over-redacts Fact or Fiction's split (those five cards are public).
  Harmless — the reveal frame shows them to the table anyway — and
  changing it is a separate visibility decision.


irobinson010:
CI note: the first run went red on `TestBotsStopPassingWhenTheLoopBreakerFires` (`server/internal/aiseat`), which this branch does not touch — it queues no `option_pick`. It is a pre-existing race in the test's `waitFor`, reproducible on `origin/develop` at the same rate (`go test ./internal/aiseat/ -run TestBotsStopPassingWhenTheLoopBreakerFires -count=120 -cpu=1`: 5/120 failures on develop, 6/120 here). Filed as #924; the run has been re-run.

Issue #756, opened 2026-09-17
ADR then implementation: gain and exchange control from spells and abilities (CR 613.1b, 701.12) — Act of Treason first, then Agent of Treachery and Sower of Temptation


krakenhavoc:
## Why

Layer 2 works, but only an Aura can use it. No spell or ability can gain control of a permanent, for any duration, and nothing can exchange control. Checked on develop (`ad9413f`):

- **The layer engine handles control.** `Layer2Control` is a real bucket (`server/internal/game/layers.go:55`, `:320`). `layerPassLocked` captures `BaseController` on the first recompute after entry (`layers.go:593`), `MoveCard` clears it (`game/zone.go:192`), and the printed characteristic resets to it every pass (`game/characteristic.go:164`). So control reverts by itself when an effect stops applying, and two control effects sort by timestamp (CR 613.7).
- `materialiseControlLocked` (`layers.go:660`) writes the result back onto `Card.Controller`, sets summoning sickness (CR 302.6) and removes the permanent from combat (CR 506.4).
- **The only producer** is `ControlAttachedBySource()` (`cards/effects/attachments.go:294`), used by Mind Control and Control Magic.
- **Until end of turn nearly works today.** `activeStaticAbilitiesLocked` feeds the turn-scoped registry into the same layer sort without filtering by layer (`layers.go:261-268`). A `StaticUntilEOT` (`cards/effects/until_end_of_turn.go:233`) with `Layer: Layer2Control` and an `AppliesTo` pinned to the target's `(InstanceID, EnteredBattlefieldAt)` should give Act of Treason, reverting at cleanup. No card or test uses it.
- **A stale comment says otherwise:** `game/turn_scoped_statics.go:36-39` still calls layer 2 "deferred engine-wide". It was not deferred after S24 (ADR 0036 decision 17). The audit said ADR 0035 repeats this; on develop it doesn't (ADR 0035 §7 is about layer 1 only).

What's missing:
- A durable gain-control effect: no stated duration (CR 611.2a), "for as long as you control ~" or "for as long as ~ remains on the battlefield" (CR 611.2b). The only registry expires at cleanup. That is the sibling seam **permanent-duration-effects** (#755).
- Exchange of control (CR 701.12a–b): all or nothing, both effects with one timestamp.
- Choosing "an opponent" to give control to (the engine-seams "Choose-a-player / choose-an-opponent prompt" row, no issue). With two players the choice can be skipped.
[...]

Issue #755, opened 2026-09-17
ADR then implementation: continuous effects that outlast the turn (CR 611.2a–b) — "until your next turn", "for as long as", and no stated duration


krakenhavoc:
## Why

A continuous effect from a resolved spell or ability can last only until end of turn. CR 611.2a says an effect with no stated duration lasts until the end of the game, and CR 611.2b covers "for as long as". The engine can't express either one, or "until your next turn". Checked on develop (`ad9413f`):

- The S32 registry is `ScopedStatic` / `g.TurnScopedStatics` (`server/internal/game/turn_scoped_statics.go:63-105`, `game/game.go:355`). Its only duration field is `ExpiresAfterTurn int` (`:100`).
- The only way in is `RegisterTurnScopedStaticForEffect` (`:132`), which always stamps `ExpiresAfterTurn: g.Turn.Number` (`:141`). The cleanup sweep `ClearExpiredTurnScopedStaticsLocked` (`:164`, called at `game/game.go:1075`) drops it at the first cleanup.
- The file's own "What is NOT here" note (`:28-35`) names "until your next turn", "for as long as ~ remains exiled" and "until ~ leaves the battlefield" as missing.
- `Turn.Number` counts rounds, not seat-turns (field comment at `:93-99`; `turn.go:188` bumps it only on wrap). "Until your next turn" therefore needs the controller's seat as well, not a bigger number.
- The replacement twin, `TurnScopedReplacements` (`game/game.go:345`), is cleared wholesale at cleanup (`game/replacements.go:848-852`). It has no expiry field at all.
- Both registries hold closures and are counted in the snapshot census (`game/snapshot.go:459-499`), so any entry makes a game a less-than-full restore point. Today that lasts one turn. An indefinite entry would last the whole game.
- `server/internal/cards/effects/batch31_test.go:130-137` keeps Tree of Redemption out of the catalog as "a continuous effect with no duration".

What already exists and can be reused: the card-side helpers in `cards/effects/until_end_of_turn.go` key the affected set on `(InstanceID, EnteredBattlefieldAt)` (`eotAffected`, `:40-51`). That already gives CR 400.7 (a flickered object is a new object) and the CR 611.2c locked-in set.

## Cards

[...]

Issue #881, opened 2026-09-18
S38 — Layers: dependency ordering, ability grants, ability removal


krakenhavoc:
**Phase:** 7 · **Goal:** CR 613.6-613.8 hold under interacting continuous effects

Planned 2026-09-18 as part of the S36–S49 backlog pass over the 210 open issues. Milestone: `S38 — Layers: dependency ordering, ability grants, ability removal`.

> **Verification status.** This tracker was built from issue titles, labels, bodies and `docs/engine-seams.md`, not from a per-issue re-check against `origin/develop`. Unlike [#734](https://github.com/krakenhavoc/cmd_and_ctrl/issues/734), no member issue here carries a "still reproducible on `develop`" verdict. Re-verify before starting: the repo merges to `develop` while the default branch is `main`, so `Closes #N` never fires and shipped work can sit open.

## Members (8)

- [ ] #668 — ADR: CR 613.8 dependency ordering in layer 4 — Urborg + Song of the Dryads / Arixmethes, Maskwood + crewed Vehicles / Warring Triad
- [ ] #754 — ADR then implementation: abilities granted to other permanents (CR 113.10, layer 6) — Cryptolith Rite, Rishkar, Gemhide Sliver, Necrotic Sliver
- [ ] #755 — ADR then implementation: continuous effects that outlast the turn (CR 611.2a–b) — "until your next turn", "for as long as", and no stated duration
- [ ] #756 — ADR then implementation: gain and exchange control from spells and abilities (CR 613.1b, 701.12) — Act of Treason first, then Agent of Treachery and Sower of Temptation
- [ ] #669 — Ability removal drops a silenced source from every layer (CR 613.6), and Song of the Dryads wipes granted abilities (CR 305.7)
- [ ] #670 — "Every creature type" is stored as a keyword, so layer-6 ability removal wipes a layer-4 grant (Maskwood Nexus)
- [ ] #690 — Toughness SBA still skips coded */* CDA creatures at computed toughness 0 (Consuming Aberration, Lord of Extinction)
- [ ] #691 — Printed 0/0 X-counter creatures survive an X=0 cast: player-facing caveats (and fix Wildwood Scourge's comment)

## Dependencies and sequencing

#668 must land before or with #754 - Maskwood plus a crewed Vehicle is exactly the dependency pair #668 describes. This is the highest card-unlock sprint in the plan.

## Exit criteria

1. Every member issue above is closed, or carries a named decision explaining why it is not.
[...]

Issue #760, opened 2026-09-17
ADR then implementation: one announce-time cast gate — "can't cast" restrictions (CR 101.2) and "cast only if" conditions (legendary sorcery, Rakdos)


krakenhavoc:
## Why

No static, effect or spell condition can stop a player casting a spell. Checked on develop (`ad9413f`):

- `Game.CastSpell` (`server/internal/game/mutations.go:446`) is the only entry point for a cast. Its checks run in this order: split second (`:463`), source zone (`:465`), face (`:514-530`), then the exile grant, alternative cost, cast path and no-mana-cost checks. None of them asks "is this player forbidden from casting this spell?". No `CantCast`, `CastRestriction`, `CastCondition` or `ErrCantCast` exists in the tree.
- The bot enumerator `legal.castMoves` (`server/internal/legal/cast.go:35`) repeats the split-second check (`:37`), so any new gate needs a twin there. The view's `CastableHere` stamp (`server/internal/protocol/view.go:849-865`) would otherwise tell the client a banned card is castable.
- The closest existing template is `CatalogCostModifiers` (`cost_modifier.go:197`, with `CostQuery` at `:113`). It is a static on another permanent, looked up at cast time with the caster and the face already chosen.
- The per-turn cast count is `CastTally{Total, Noncreature}` (`game.go:795-798`). Rule of Law needs `Total` and Deafening Silence needs `Noncreature`; Ethersworn Canonist needs a nonartifact count that doesn't exist yet.
- "Until your next turn" can't be expressed: `turn_scoped_statics.go:93-99` says so, because `Turn.Number` counts rounds. Avatar's Wrath, Reflector Mage, Hope of Ghirapur and Single Combat need it.
- `restrictions.go` has per-permanent bits only. `CantActivate` and `CantActivateMana` already cover Grand Abolisher's activation half; nothing covers a player.

## Cards

23 cards where this is the only core blocker and 31 with it as any core blocker. The verifier estimated about 25% of the sampled cards are also blocked by something else (≈17 real).

Examples: Rule of Law, Eidolon of Rhetoric, Deafening Silence, Archon of Emeria, Drannith Magistrate (must also cover casts from the command zone, graveyard and exile), Silence, Ranger-Captain of Eos, Grand Abolisher, Ethersworn Canonist (needs a nonartifact count), Dragonlord Dromoka.

**Misattributed or also blocked:**
[...]

PR #921, opened 2026-09-18
feat(engine): continuous effects carry a Duration, and spells can gain or exchange control (CR 611.2, CR 613.1b, CR 701.12)

irobinson010:
## Summary

Two seams from S38, one duration model and one control primitive.

**Durations (#755).** `ScopedStatic.ExpiresAfterTurn int` becomes
`ScopedStatic.Duration`, and one function — `durationExpiredLocked` in
`server/internal/game/duration.go` — decides when any continuous effect in the
game is over. Four kinds:

| Oracle text | Kind | Ends |
|---|---|---|
| "until end of turn" | `UntilEndOfTurn` | that turn's cleanup step (CR 514.2) — unchanged behaviour |
| "until your next turn" | `UntilYourNextTurn` | as that player's next turn begins, before untap (CR 500.1 / 502.1), and when a departed player's turn *would have* begun (CR 800.4m) |
| "for as long as ~ remains on the battlefield" / "for as long as you control ~" | `ForAsLongAs` | when the condition goes false, re-checked at the top of every layer pass (CR 611.2b) |
| nothing printed | `Indefinite` | never (CR 611.2a) |

The sweep runs at exactly three moments — the cleanup step, the turn-began hook
and the top of `recomputeLayersLocked` — and nowhere else. No card schedules its
own expiry.

"Until your next turn" counts seat-turns, not rounds: `Player.TurnsBegun` lands
from [ADR 0059](docs/decisions/0059-turn-machinery.md) Decision 1 (which
explicitly reserved it for #755), bumped by the single rotation seam for the seat
whose turn begins **and** for every eliminated seat the rotation steps over, so
CR 800.4m falls out rather than being special-cased. `Turn.Number` is not read by
any duration, which retires ADR 0035 §3's "counts ROUNDS" caveat.

`Duration` is plain data — no closures — so it clones, and it is the half of a
scoped static #515 can persist verbatim. An optional pin ends an entry once the
one object it exists to affect is gone (CR 400.7), which is what keeps an
indefinite effect from sitting in the snapshot census for the rest of the game.

The registry is renamed to match what it now holds: `Game.ScopedStatics`,
`RegisterScopedStaticForEffect`, `scoped_statics.go`. The census keeps its
`turnScopedStatics` wire key so snapshot files written by older binaries still
decode.

**Control (#756).** Layer 2 has worked since S24 and only an Aura could reach it.
`GainControlForEffect(source, target, controller, duration, label)` registers one
layer-2 `ScopedStatic` in the same bucket Mind Control uses, so
`Card.BaseController` is the reversion, CR 613.7 timestamp ordering against a
second control effect is free, and `materialiseControlLocked` stays the one place
the consequences fire. `ExchangeControlForEffect` is CR 701.12: both objects
checked before either half is registered (701.12b — all or nothing), and the two
halves share **one** timestamp, so a later control-changer beats both or neither.

**Bug fixed along the way (#871).** `materialiseControlLocked` cleared
`AttackingTarget` / `BlockingTarget` on a control change but left the attack and
block *announcements* (`announcedAttacks`, `announcedBlocks`,
`announcedBecameBlocked`). A creature stolen mid-combat therefore stayed marked as
already-declared and could never fire a "whenever ~ attacks" trigger again that
combat. One `removeFromCombatLocked` helper now clears the declaration and its
announcements together. The assertion fails with the fix backed out (verified).

## ADR

[ADR 0063 — Durations of continuous effects, and control from spells and
abilities](docs/decisions/0063-durations-and-control.md). Amends ADR 0035 §3
("Expiry is stamped, not implied") and ADR 0036 decision 17 (layer 2 is no longer
Aura-only); implements the `Player.TurnsBegun` third of ADR 0059 Decision 1 and
leaves `Turn.Seq`, `Round` and extra turns with #753.

Numbering was re-checked against every remote head immediately before pushing:
0061 and 0062 are the parallel replaceable-events and hand-abilities ADRs, 0064 is
the emblems ADR, and 0063 is free.

## Sprint

S38 — Layers: dependency ordering, ability grants, ability removal (tracker #881).

## Issues

Closes #755
Closes #756
Refs #881, #871

## Cards converted

Five new catalog cards, one per duration plus the exchange:

- **Act of Treason** — gain control until end of turn, untap, haste. `full`. The
  haste clause is not decoration: summoning sickness is per controller (CR 302.6),
  so the test asserts both that the flag is set and that haste lifts it.
- **Agent of Treachery** — gain control of target permanent with no stated
  duration (CR 611.2a), plus its "at the beginning of your end step, if you
  control three or more permanents you don't own, draw three cards"
  intervening-if. `full`.
- **Sower of Temptation** — flying; gain control for as long as it remains on the
  battlefield (CR 611.2b). `full`.
- **Mass Diminish** — "until your next turn, creatures target player controls have
  base power and toughness 1/1", with flashback. `full`.
- **Switcheroo** — exchange control of two target creatures (CR 701.12). `full`.

Census: 1723 → 1728 registry keys, regenerated with
`go test ./internal/cards/coverage/ -update`.

**Caveats grepped and not converted** (`until your next turn` / `for as long as`
across the catalog), with the reason each is not cheap:

- **Teferi, Time Raveler** (`teferi_time_raveler.go:78`) — the +1 is "until your
  next turn, you may cast sorcery spells as though they had flash". A cast-timing
  permission, not a layer effect; it needs #760's gate, not a duration.
- **Alela, Cunning Conqueror** (`alela_cunning_conqueror.go:30`) — the goad marker
  is already cleared at the controller's next turn by a delayed trigger. Its
  caveat is that goad is not *enforced*, which no duration fixes.
- **Aerial Extortionist** (`aerial_extortionist.go:10`) — "for as long as" a card
  remains exiled, which is the exile-return machinery, not a continuous effect.

## Test plan

Run from `server/` on the rebased branch:

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./...` — clean
- [x] `AISEAT_GAME_TESTS=1 go test ./internal/...` — clean
- [x] `make lint` (golangci-lint v1.64.8) — clean
- [x] `gofmt -l internal/ cmd/` — empty
- [x] census block and clone baseline regenerated
      (`go test ./internal/cards/coverage/ -update`), roadmap in sync
- [ ] client — untouched. No protocol or wire change: a control change reaches
      the client through `CardView.controller`, which the existing Mind Control
      path already exercises. `grep` over `client/src` and `internal/protocol` for
      the renamed registry and the new field finds nothing.

New engine tests (`server/internal/game/duration_test.go`,
`control_effects_test.go`):

- each duration expires at exactly its boundary: end of turn; the player's next
  turn beginning and **not** any of the three intervening opponents' turns; an
  "until your next turn" made on the holder's own turn lasting a full round;
  `ForAsLongAs` ending when the source leaves and when it is flickered back
  (CR 400.7); `ForAsLongAs` never starting when the condition is already false;
  "for as long as you control ~" ending on a control change; `Indefinite`
  surviving nine turns
- a departed player's "until your next turn" ends when their turn *would have*
  begun (CR 800.4m), not on their concession
- `Start` stamps the starting seat's first turn (the one turn that does not come
  through the rotation seam)
- until-end-of-turn is not ended early by either new sweep site, and still ends at
  cleanup
- the pin drops an entry when its object leaves; the census labels name the
  duration
- gain control moves the permanent, grants per-controller summoning sickness,
  removes it from combat **and clears its announcement**, and reverts to
  `BaseController` on expiry; two control effects sort by timestamp, and the
  later one ending hands the creature to the *earlier thief*, not the owner
- gain control refuses a permanent that is not on the battlefield, and ends when
  the stolen permanent is flickered
- exchange both ways, with one shared timestamp; a failed exchange registers
  nothing and moves nothing
- a stolen commander keeps its commander state and its owner
- undo across a control change, undo across an expiry (the snapshot still holds
  the theft the original's sweep dropped), snapshot round-trip with the census
  refusing the restore point and `TurnsBegun` surviving

Card tests (`mass_diminish_test.go`, `control_cards_test.go`): one assertion per
card, plus Mass Diminish's CR 611.2c locked-in affected set and Agent of
Treachery's end-step draw.

Two new bug assertions were checked by backing the fix out: the #871 announcement
clear, and the CR 400.7 battlefield-entry stamp in the `ForAsLongAs` condition.
Both fail without their fix.

## Out of scope, stated

- "An opponent gains control" (Humble Defector, Wishclaw Talisman) — the primitive
  already takes an arbitrary controller; those cards wait on the choose-a-player
  prompt.
- "When you lose control of ~" (Khârn the Betrayer) — there is no event for a
  layer-2 delta.
- Control of a spell (Commandeer) and of a player's turn (Mindslaver) — neither is
  layer 2.
- `TurnScopedReplacements`, ADR 0057's `TurnScopedGameEndGates` and ADR 0045's
  `TurnScopedBlockRules` are not folded into the `Duration` model. ADR 0059
  expects #755 to do it eventually; no card needs a longer duration from them yet,
  and building it in three registries without a card to prove it would be
  speculative.


irobinson010:
CI note: the first run went red on `TestBotsStopPassingWhenTheLoopBreakerFires` (`internal/aiseat`), which this branch does not touch — it is the known flake filed as #924 today, verbatim the same two assertions, and it reproduces on `develop` at the same rate. Locally the test passes 20/20 alone and 3/3 as a package on this branch. Re-run: `cmdctrl-ci` green.

Issue #661, opened 2026-09-16
CR 514.3a not implemented: a trigger during cleanup waits until the next player's upkeep instead of getting priority


krakenhavoc:
## Bug

CR 514.3a isn't implemented. The cleanup step gives nobody priority (CR 514.3). But if a state-based action happens or a trigger is waiting during cleanup (typically from the hand-size discard), the rules say: "triggered abilities are put on the stack, then the active player gets priority. … Once the stack is empty and all players pass in succession, another cleanup step begins."

The engine auto-advances out of cleanup as soon as nobody owes a discard, without checking for pending triggers:
- `server/internal/game/game.go:1042-1045`: `if len(g.DiscardPending) == 0 { g.advanceCursorLocked(); g.runStepEntryHooksLocked() }`
- `server/internal/game/mutations.go:2930-2932`: after the last hand-size discard, `DiscardSelection` re-runs the step-entry hook, which advances the same way.

Triggers queued in cleanup sit in `PendingTriggers` until the next priority boundary, which is the **next player's upkeep** (untap grants no priority; see the comment at `game.go:957-965`).

Catalog cards that watch `EventDiscardCard` and hit this when the discard is the cleanup one: Surly Badgersaur, Mary Read and Anne Bonny, Sangromancer, Marauding Mako, Scrounging Skyray. Madness (#657) depends on this fix.

## Repro (verified with a throwaway effects-package test on develop `7c1ae9b`; not committed)

1. `newCatalogGame`. Seat 0 controls Sangromancer ("Whenever an opponent discards a card, you may gain 3 life").
2. Advance to seat 1's end step and give seat 1 eight cards in hand. Advance into cleanup: `DiscardPending = {seat 1: 1}`.
3. Seat 1 answers with `DiscardSelection`.

**Actual:** right after the discard the game is already in **seat 2's upkeep**, with priority at seat 2. Sangromancer's "may" prompt is open there (`PendingChoices` = 1). Anything the trigger checks "this turn", or an until-end-of-turn effect it creates, lands in the wrong turn.

**Expected (CR 514.3a):** the trigger goes on the stack during seat 1's cleanup and seat 1 gets priority. The trigger resolves, a new cleanup step runs, and only then seat 2's turn begins.

The fix PR turns this into a committed test that fails on develop.

## Scope

[...]

Issue #630, opened 2026-09-16
Loyalty 'already activated this turn' survives a planeswalker leaving and returning the same turn (CR 400.7)


krakenhavoc:
## Bug

A planeswalker whose loyalty ability was activated this turn, and which then leaves the battlefield and comes back **in the same turn**, still counts as "already activated". The client greys out every loyalty ability with "Already activated this turn", and the server refuses the activation. Under CR 400.7 the returning permanent is a **new object**, so its loyalty abilities should be available (CR 606.3 applies per object).

Found while re-testing #376 in the app on 2026-09-16. It is related to, but probably **not**, what #376's reporter hit (their log shows no earlier activation that turn).

## Repro (verified in the app, local server at `ffbca1c`, which includes 6768113)

1. Cast a catalogued planeswalker, e.g. **Teferi, Hero of Dominaria**, and let it resolve.
2. Activate any loyalty ability (here +1: loyalty 4 → 5).
3. The same turn, return it to hand. Both of these reproduce:
   - normal play: cast **Venser, Shaper Savant** and target Teferi with its enters trigger
   - sandbox: context menu **Move to → Hand**
4. Recast Teferi.

**Expected:** Teferi enters with loyalty 4 and both abilities available.
**Actual:** loyalty 4, but both rows are disabled with "Already activated this turn". The replay's recast snapshot has `loyalty_activated: true`.

Evidence: in the replay, Teferi keeps instance ID `2e826c94` through battlefield → hand → stack → battlefield, and the recast snapshot (seq 34) carries `loyalty_activated: true`.

## Cause

- `Game.LoyaltyActivatedThisTurn` (`server/internal/game/game.go:145`) is keyed by **instance ID** and cleared only on turn advance (`onTurnAdvanceLocked`).
- Instance IDs survive zone changes, and nothing deletes the entry when a card leaves the battlefield (`grep -rn 'delete(g.LoyaltyActivatedThisTurn' server/` → nothing).
- It's written at `activated.go:472` (and the sandbox `activate_loyalty` path in `mutations.go`). It's read at `activated.go:378`, `legal/abilities.go:60` and `protocol/view.go`, so the server gate, the bot enumerator and the client flag all see the stale value.

## Fix direction

[...]

Issue #884, opened 2026-09-18
S41 — Turn machinery and per-turn accounting


krakenhavoc:
**Phase:** 7 · **Goal:** extra turns, extra phases, the untap step and the cleanup step all behave

Planned 2026-09-18 as part of the S36–S49 backlog pass over the 210 open issues. Milestone: `S41 — Turn machinery and per-turn accounting`.

> **Verification status.** This tracker was built from issue titles, labels, bodies and `docs/engine-seams.md`, not from a per-issue re-check against `origin/develop`. Unlike [#734](https://github.com/krakenhavoc/cmd_and_ctrl/issues/734), no member issue here carries a "still reproducible on `develop`" verdict. Re-verify before starting: the repo merges to `develop` while the default branch is `main`, so `Closes #N` never fires and shipped work can sit open.

## Members (6)

- [ ] #753 — ADR then implementation: turn machinery — extra combat and main phases (CR 500.8) and extra turns (CR 500.7); fix Turn.IsNewTurn and define Turn.Number
- [ ] #826 — ADR then implementation: untap-step choices (CR 502.3) — "untap no more than N" and "may choose not to untap" (Winter Orb, Static Orb, Winter Moon)
- [ ] #661 — CR 514.3a not implemented: a trigger during cleanup waits until the next player's upkeep instead of getting priority
- [ ] #567 — Cumulative upkeep (CR 702.24) — Mystic Remora
- [ ] #630 — Loyalty 'already activated this turn' survives a planeswalker leaving and returning the same turn (CR 400.7)
- [ ] #811 — Hobgoblin Bandit Lord counts Goblins as they are now, not as they entered, so it can count creatures the printed card doesn't

## Dependencies and sequencing

ADR 0059 (#753) is already merged. #826 sits beside the merged #751 untap primitive. Six issues rather than 8-12 because #753 alone is three decisions wide and touches the turn loop that every other sprint's tests run through.

## Exit criteria

1. Every member issue above is closed, or carries a named decision explaining why it is not.
2. The sprint section in `docs/sprints.md` is written and its index row added.
3. Any new engine primitive is reflected in `docs/engine-seams.md`, with the card counts it unblocks.
4. Lands on `main` with the next promotion.

irobinson010:
#661 and #630 closed with #935.

irobinson010:
#826 and #567 closed with #962 (ADR 0070 untap-step choices; cumulative upkeep).

[...]

Issue #717, opened 2026-09-16
Model the second combat damage step with its own priority window (CR 510.4) when first or double strike is present


krakenhavoc:
## Why

The rules have **two combat damage steps** when any attacker or blocker has first strike or double strike, with a **priority window between them** (Aug 7, 2026 CR 510.4, restated in 702.4b and 702.7b; 506.1 says "there are two combat damage steps"). CR 510.3 / 510.3a give the active player priority after each step, with damage triggers put on the stack first.

The engine has **one** `StepCombatDamage` (`server/internal/game/turn.go` ~:31) that runs two damage passes back to back, with no priority between them. As a result:
- a "whenever this deals combat damage" trigger from **first-strike** damage resolves only **after** regular damage has been dealt
- no player can respond between the steps (e.g. pump or remove a creature after first-strike damage)

**Decision (2026-09-16, ADR 0053 for #187):** #187's beat animation ships as **presentation-only** now. Modelling the real second step is tracked here as separate work.

## Scope (design first: an ADR or an amendment to ADR 0006 or 0014)

- [ ] Turn structure: a second combat-damage step (or substep with its own priority window) when first or double strike is present. Covers `turn.go`, the `turn.ts` step list, `PhaseIcon`, stops and auto-pass, and `Step` events.
- [ ] Triggers from first-strike damage go on the stack and resolve in the window before regular damage.
- [ ] Participation per CR 702.7c (who deals damage in the second step), consistent with the fix for **#716**.
- [ ] Legal-move enumerator (`internal/legal`), bot policies, snapshots and replays, protocol docs.
- [ ] Revisit ADR 0053's client beat sequencer: with two real frames, most of it may become unnecessary.
- [ ] Tests: a first-strike damage trigger resolves before regular damage; a player can respond between the steps; no second step when nobody has first or double strike.

Related: #187 / ADR 0053, #702, #715, #716, #693 (510.x citations), ADR 0014.

irobinson010:
Claimed (2026-09-18, S37): an agent has this together with #716 on branch `feat/717-716-second-damage-step` off `develop` (one PR, separate commits). Please don't pick it up in parallel.

irobinson010:
[...]

PR #935, opened 2026-09-18
fix(engine): cleanup gives priority when something happens (CR 514.3a), and per-object per-turn state dies with the object (CR 400.7)

irobinson010:
## Summary

Two turn-machinery bugs, one commit each.

### The cleanup step gives priority when something happens in it (CR 514.3a, #661)

Cleanup grants nobody priority (CR 514.3) — until a state-based action
is performed or a triggered ability is waiting to go on the stack.
Then, per CR 514.3a, the SBAs happen, the triggers go on the stack,
the active player gets priority **in** the cleanup step, and once the
stack is empty and all players pass in succession, **another cleanup
step begins**.

The engine had no such window: the cleanup hook advanced out of the
turn as soon as nobody owed a CR 514.1 discard, so a trigger queued by
that discard waited on `PendingTriggers` until the next priority
boundary — the *next player's upkeep*, because untap grants none
either. Seat 1 discards to hand size and Sangromancer's "may gain 3
life" prompt opens in seat 2's upkeep. Five catalog cards watch
`EventDiscardCard` and all five hit it; madness (#657) needs the same
window.

- One cleanup exit decides it: `exitCleanupStepLocked`
  (`server/internal/game/cleanup.go`), called from the `StepCleanup`
  case of the step-entry hook and from `DiscardSelection`'s resume —
  the two ways a cleanup step's turn-based actions can finish. They
  used to decide separately, which is how the discard path inherited
  the bug.
- The condition is the rule's: an SBA was performed, or a trigger is
  waiting. `runStateChecksLocked` now reports the first (one body, no
  reporting copy — every other caller ignores the value); the second
  is read off `PendingTriggers` before the drain empties it. A trigger
  held behind its own CR 603.5 "you may" counts as waiting, via
  `blockingChoiceLocked` — an optional trigger reaches
  `PendingTriggers` only once the answer arrives, and Sangromancer's
  prompt is exactly the shape the issue reported.
- Priority in cleanup is ordinary priority: `Turn.PriorityHolder`
  becomes the active seat, so `PassPriority`, the enumerator's
  `holdsPriority`, the bot runner and the client all work unchanged.
  **No client change was needed** — `PhaseDisplay` derives "somebody
  holds priority" from `priority_holder >= 0`
  (`client/src/lib/components/board/PhaseDisplay.svelte:65`) and
  `NO_PRIORITY_STEPS` (`client/src/lib/turn.ts:56`) is only consulted
  for which steps may be *pinned* as stops
  (`priorityStops.ts:47`, `settings.ts:208`), not for the indicator or
  the pass button.
- The wrap begins another cleanup step instead of the turn: hand size
  is checked again and the CR 514.2 sweep runs again, which closes
  ADR 0035 §3's "an effect created during cleanup survives into the
  next turn". The cursor does not move for it, so that step opens its
  own event batch — the third caller of `beginEventBatchLocked`, and
  the only step the rules begin again without the cursor moving now
  that #717 made the second combat damage step a real one.
- A quiet cleanup is untouched: no SBA, no trigger, no prompt, nothing
  on the stack, and the turn ends in the same call it always did, with
  no new stop for autopass or the bots.
- Not a loop: ADR 0055's breaker counts resolutions per
  `(source, label)`, so an extra cleanup adds at most one, and a card
  that really re-triggered every cleanup would be paced by a full
  priority round and trip the breaker like any other loop.

### Per-object per-turn state dies with the object (CR 400.7, #630)

Activate Teferi's `+1`, bounce him to hand the same turn, recast him:
both loyalty rows come back greyed with "Already activated this turn",
the server refuses the activation, and the recast snapshot still
carries `loyalty_activated: true`. `LoyaltyActivatedThisTurn` is keyed
by instance ID and flushed only when a turn begins, and an instance ID
survives a zone change — but the permanent that came back is a new
object and CR 606.3 applies per object.

- `battlefieldExitLocked` (`server/internal/game/battlefield_exit.go`)
  is the Game-side half of `MoveCard`'s one battlefield-exit cleanup:
  it takes the CR 603.10 LKI snapshot the three exits were already
  taking, then forgets what the leaving object did. All three exits
  call it (the destroy / sacrifice route, the effect route, the
  sandbox move), so the fix is true of a bounce, an exile, a
  sacrifice, a tuck and a context-menu shove alike. The blink path was
  already safe for a different reason (the exile return mints a new
  instance ID, ADR 0026 §5) and now also stops leaving a dead row
  behind.
- Siblings fixed in the same place: `announcedAttacks`,
  `announcedBlocks`, `announcedBecameBlocked` — same key, same rule,
  cleared only at end of combat. A stale row there is unreachable
  today (a permanent that leaves is out of combat and returns with no
  attacking or blocking target), so this is the shape being closed,
  not a second reported bug.
- Siblings fixed on the card side, in `MoveCard`'s own block:
  `SummonedThisTurn` and `EnteredBattlefieldAt`. Every battlefield
  entry stamps both unconditionally, so the returning permanent was
  already right; what this fixes is the card in between, which was
  telling the wire it was summoning sick from the graveyard it had
  died to.
- Deliberately not moved: `TurnTally`'s per-ability counts
  (`Resolved`, `Triggered`, `LoopRun`, `LoopAllowance`) are per object
  too, and a returning object arguably should trigger its "only once
  each turn" ability again — but clearing them on exit would let a
  blink loop reset ADR 0055's `LoopRun` every iteration and never trip
  the breaker. That trade needs the breaker keyed on something other
  than the source instance first.
- No new field, so clone / snapshot / drift need nothing: the fix is
  deletes from maps they already carry.

Docs: AGENTS.md §7 gets a paragraph for each; dated amendments to
[ADR 0006](docs/decisions/0006-priority-foundation.md) (the step /
priority machinery, noting CR 514.3a), [ADR 0035
§3](docs/decisions/0035-until-end-of-turn-effects.md) (its known
simplification is gone) and [ADR
0032](docs/decisions/0032-planeswalkers.md) (the once-per-turn gate is
per object). No catalog completeness changed, so the census is
untouched.

## Sprint

S41 — Turn machinery and per-turn accounting (tracker #884).

## Issues

Closes #661
Closes #630
Refs #884, #657

## Test plan

- [x] `go build ./... && go vet ./... && go test ./...` from `server/` — green
- [x] `AISEAT_GAME_TESTS=1 go test ./internal/...` — green
- [x] `make lint` (golangci-lint v1.64.8, via `go run`) — clean
- [x] Rebased onto `origin/develop` at `caee7b78` and re-ran all of the above
- [x] Every new bug assertion fails with its fix backed out (checked by
      reverting `exitCleanupStepLocked` to the old unconditional advance,
      and by emptying `forgetPerObjectTurnStateLocked` plus the two
      `MoveCard` lines)
- [x] #661: the issue's Sangromancer repro (opponent discards to hand
      size in cleanup → the "may" prompt, the trigger and the life all
      land in that turn), Marauding Mako, and one assertion each for
      Surly Badgersaur, Scrounging Skyray, Mary Read and Anne Bonny
      (`server/internal/cards/effects/cleanup_discard_trigger_test.go`)
- [x] #661: a quiet cleanup ends the turn with no priority window; a
      discard trigger gets priority and a second cleanup step; a trigger
      in the second cleanup gives a third; an SBA in cleanup (an "until
      end of turn" pump ending on a creature with -1/-1 counters) opens
      the same window; an "until end of turn" effect created in cleanup
      expires in the repeated one; undo across the window
      (`server/internal/game/cleanup_priority_test.go`)
- [x] #661: the enumerator offers the holder a pass inside the window
      and the table passes out of it, so no bot wedges
      (`server/internal/legal/cleanup_priority_test.go`)
- [x] #630: activate → bounce → recast is allowed, with the instance ID
      asserted unchanged; the walker that stayed put is still gated; a
      flicker keeps its printed loyalty and leaves no row behind; undo
      across the refresh; the combat maps and the summoning-sickness
      pair (`server/internal/game/battlefield_exit_turn_state_test.go`)
- [x] #630: the wire half — a returned planeswalker stops reporting
      `loyalty_activated` (`server/internal/protocol/loyalty_view_test.go`)
- [ ] Client lint / check / test — not run: no client file is touched
      (the engine change rides on `priority_holder`, which the client
      already reads)

One flake seen once and not reproducible:
`TestBotsStopPassingWhenTheLoopBreakerFires` (`internal/aiseat`) failed
on a single run while other agents were hammering the machine, and
passed on every run since (3× alone, 2× in the full package, plus two
clean full `AISEAT_GAME_TESTS=1` sweeps). It is a goroutine-timing test
over the bot runner and touches nothing either commit changes.


irobinson010:
Out of scope, noticed while doing this and filed rather than fixed here: #936 — `TurnTally.Resolved` / `Triggered` are keyed by the source's instance ID too (`server/internal/game/turn_tally.go:66-76`, read at `server/internal/cards/effects/batch16_helpers.go:206`), so an "only once each turn" gate survives the permanent leaving and returning. It is not fixed at this PR's battlefield exit because `TurnTally.LoopRun` shares the key, and clearing it on exit would let a blink loop reset ADR 0055's breaker every iteration. The reasoning is recorded in `forgetPerObjectTurnStateLocked`'s doc comment.

Also noticed, not filed (a documented design decision, not a defect): `DrawnThisTurn` keeps a card's ID after it leaves hand and returns as a new object, so Sylvan Library would offer a card that was drawn by a previous object — `server/internal/game/game.go:179-193` states the choice and why.

irobinson010:
CI is red for an infrastructure reason, not this branch: the self-hosted runner `gunner-0` is out of disk. The first run died in `go vet` with `no space left on device` (run 35380631059 / 35380905553), and the `--failed` rerun did not even get through **Set up job** — `##[error]No space left on device` while downloading `actions/checkout`. Nothing in the job reached our code.

Locally, from `server/` at `bdaee049`: `go build ./... && go vet ./... && go test ./...` green, `AISEAT_GAME_TESTS=1 go test ./internal/...` green, `make lint` (golangci-lint v1.64.8) clean — all re-run after the rebase onto `caee7b78`. Worth a rerun once the runner has space.

Issue #697, opened 2026-09-16
Face-down flag survives admin moves and casts out of exile (CR 400.7); StackOverlay hover guard never fires for unknown cards


krakenhavoc:
## Why

Two face-down lifecycle bugs on develop (`7c1ae9b`). #646 listed both as out of scope.

**1. `FaceDown` survives a move out of exile unless the move goes through `zone_route.go` (CR 400.7, 708).**
- `MoveCard` (`server/internal/game/zone.go:164-233`) resets battlefield-only state and the DFC face, but never `FaceDown`.
- The shared exit route clears it itself after calling `MoveCard` (`zone_route.go:233-242`), and so does the exile→battlefield return (`effect_api.go:2006`). Two callers that can take a card out of exile skip that:
  - **Admin/sandbox move, live today:** `MoveCardByIDAsCommander` (`mutations.go:3136`) calls `MoveCard` and then `markCardKnownInZoneLocked`, with no reset. If a player uses `move_card` on a card Necropotence exiled face down, it lands in hand or on the battlefield still marked `FaceDown=true`. It keeps that flag in snapshots, on the wire (`face_down`, `protocol/view.go:681`), and in what the bots read (`aiseat/model/prompt.go:396`, `aiseat/heuristic/score.go:290`).
  - **Cast path, latent:** `CastSpell` moves an exiled card to the stack with `MoveCard(src, g.Stack, cardID)` (`mutations.go:874`). It resets `ExilePlay` and sets the face, but not `FaceDown`. Nothing can cast a face-down exiled card today: `ExileTopFaceDownForEffect` gives no `ExilePlay` grant, and no catalog card declares `ZoneExile` in `CastableZones`. Foretell (#658) casts exactly that card, though, and CR 406.3a turns it face up just before the cast.

**2. `StackOverlay.svelte`'s hover guard never fires.**
- `previewCardFor` returns null only when `c.known_by_you === false` (`client/src/lib/components/board/StackOverlay.svelte:185-189`).
- `KnownByYou` is `omitempty` (`server/internal/protocol/view.go:689`), so the server never sends `false`. `Card.svelte:130-139` already says so in a comment.
- So a stack spell, or an ability's source card from the battlefield or exile (`artCardFor`, `StackOverlay.svelte:108-111`), that the viewer doesn't know still goes to `hoveredCard`. The zoom panel then shows a blank card, because `redactCardForViewer` has stripped its name and art.
[...]

Issue #656, opened 2026-09-16
ADR: face-down objects (CR 406.3a / 708) — owner-known exile, 2/2 projection, catalog suppression


krakenhavoc:
## Why

Face-down objects (CR 406.3a, CR 708) exist only as a flag. `Card.FaceDown` (`server/internal/game/card.go:241-248`) is set in one place: face-down **exile** through `routeCardToZoneLocked` (`game/zone_route.go:220-247`). Its only user is Necropotence. Nothing puts a permanent face down.

Two families need more, and they need different things:

| | foretell (#94, S29) | morph / manifest / disguise / cloak (#95, S30) |
|---|---|---|
| zone | exile | battlefield (and the stack for morph and disguise casts) |
| who may look | **the owner** (CR 702.143d; 406.3a allows it) | the controller (CR 708.5) |
| characteristics | none (CR 406.3a) | a 2/2 with no name, text, subtypes or mana cost (CR 708.2); ward {2} for disguise and cloak (CR 702.168a, 701.58a) |
| reveal | all foretold cards at game end or when the owner leaves (CR 702.143f) | when it leaves the battlefield (CR 708.9) |

What's on develop (`7c1ae9b`):
- **The owner can't see their own face-down exile.** The face-down branch clears every knower, the owner included (`zone_route.go:226-231`).
- **Catalog hooks don't know about face-down.** Every hook keys on `CatalogKey(card)` → oracle ID (`game/effect_hooks.go:37-42`, one lookup through `CardDef` since #627). A face-down permanent would still run its real triggers, statics, replacements and abilities. CR 708.2a says it has no text.
- **The wire leaks identity to viewers who don't know the card.** `redactCardForViewer` (`protocol/view.go:2411-2450`) clears name, costs and faces but leaves `ManaAbilities`, `TargetMode`, `Auto`, `Restrictions` and `ActivatedAbilities`, which `viewOfCard` sets for every card (`view.go:2546-2554`). `ExilePlay.CostOverride` (`view.go:2592-2601`) would send a foretell cost to the whole table. #646 fixes today's leak (Necropotence, and your own library) and pins it with a zone × viewer test that fails on any new unredacted `CardView` field. The ADR builds on that rule.
- **Spectators and admins know every card** (`FilterViewFor`, `view.go:2220-2225`). Say whether that stays true for face-down objects.
[...]

Issue #886, opened 2026-09-18
S43 — Hand special actions and face-down objects


krakenhavoc:
**Phase:** 7 · **Goal:** cards that live in the hand, or under a face-down projection, work

Planned 2026-09-18 as part of the S36–S49 backlog pass over the 210 open issues. Milestone: `S43 — Hand special actions and face-down objects`.

> **Verification status.** This tracker was built from issue titles, labels, bodies and `docs/engine-seams.md`, not from a per-issue re-check against `origin/develop`. Unlike [#734](https://github.com/krakenhavoc/cmd_and_ctrl/issues/734), no member issue here carries a "still reproducible on `develop`" verdict. Re-verify before starting: the repo merges to `develop` while the default branch is `main`, so `Closes #N` never fires and shipped work can sit open.

## Members (7)

- [ ] #655 — ADR: hand special actions and activated abilities from hand — foretell, suspend, cycling
- [ ] #656 — ADR: face-down objects (CR 406.3a / 708) — owner-known exile, 2/2 projection, catalog suppression
- [ ] #658 — Foretell (CR 702.143) — Saw It Coming, Behold the Multiverse, Cosmic Intervention
- [ ] #659 — Suspend (CR 702.62) — Rift Bolt, Lotus Bloom, Ancestral Vision; retire the declared-ZoneExile shape
- [ ] #660 — Cycling (CR 702.29): activated abilities from hand with a discard-this-card cost — six catalog cards wait on it
- [ ] #657 — Madness (CR 702.35) — Fiery Temper, Big Game Hunter
- [ ] #697 — Face-down flag survives admin moves and casts out of exile (CR 400.7); StackOverlay hover guard never fires for unknown cards

## Dependencies and sequencing

Depends on S42's #652 (granted permissions). Internally #656 before #658 - foretell needs the owner-known face-down exile. #660 and #659 both need #655's activated-abilities-from-hand shape.

## Exit criteria

1. Every member issue above is closed, or carries a named decision explaining why it is not.
2. The sprint section in `docs/sprints.md` is written and its index row added.
3. Any new engine primitive is reflected in `docs/engine-seams.md`, with the card counts it unblocks.
4. Lands on `main` with the next promotion.

irobinson010:
[...]

Issue #659, opened 2026-09-16
Suspend (CR 702.62) — Rift Bolt, Lotus Bloom, Ancestral Vision; retire the declared-ZoneExile shape


krakenhavoc:
## Why

Suspend (CR 702.62) isn't implemented. The mechanic:
1. **Special action from hand**, whenever you could begin to cast the card: pay the suspend cost and exile it with N time counters (CR 116.2f, 702.62a, 702.62c).
2. **At the beginning of your upkeep**, remove a time counter (a trigger that works in exile).
3. **When the last counter is removed**, you may cast it without paying its mana cost, during that trigger's resolution (CR 608.2g). A creature cast this way has haste.

The shape S29 prepared for it is wrong. `CastSpell` skips the per-instance grant check when the card's own text declares `ZoneExile` (`server/internal/game/mutations.go:536-547`; also `game/cast_zones.go:24-28` and `:183`), and `TestExileCastAllowedByDeclaration` (`game/cast_zones_test.go:283-306`) calls it "the shape suspend and foretell will use". A card-level declaration makes **every** exiled copy castable at any time: suspended, exiled by Path to Exile, or still holding counters. Suspend allows the cast only as the last counter comes off. The shape that fits already exists per instance: cascade's `grantFreeCastLocked` (`game/cascade.go:171-187`) plus `PendingChoiceMayCast`.

Other gaps on develop (`7c1ae9b`):
- **Counters survive leaving exile** on develop: `MoveCard` clears `Counters` only for battlefield exits (`game/zone.go:172-174`), so a suspended creature would enter carrying its time counters (CR 400.7). #649 fixes this.
- **Cards with no mana cost:** Lotus Bloom and Ancestral Vision have an empty mana cost, which CR 118.6 treats as unpayable. Today an empty cost parses as free, so they'd be castable from hand for nothing. They need the CR 118.6 fix (#648) while still allowing suspend's free cast (CR 118.6a).
- **Exile-zone triggers:** the harvester doesn't scan exile (`game/triggers.go:244-278`). How the two triggers are built is decided in #655.

**Cards:** Rift Bolt ({2}{R}, suspend 1), Lotus Bloom (no mana cost, suspend 3), Ancestral Vision (no mana cost, suspend 4). Drop the #94 body's other picks:
- Greater Gargadon: its ability works while suspended, which needs activation from exile.
- Hypergenesis: needs #654.
- Living End: fine after the CR 118.6 fix, but not a first card.

[...]

PR #940, opened 2026-09-18
feat(engine): face-down objects — ADR, owner/controller visibility, the 2/2 projection, catalog suppression, and a single reset on zone change (CR 406.3a, CR 708)

irobinson010:
## Summary

Face-down was one `bool` on `Card` with one caller (Necropotence) and no rules behind it: nothing put a permanent face down, nothing turned one face up, the layer engine had never heard of the flag, catalog hooks fired for a face-down card, the owner could not see their own face-down exile, and the flag survived a move out of exile on two paths. This turns it into an object model, so foretell (#658) and the morph family (#95) become implementation rather than design.

**One state.** `Card.FaceDownKind` says *why* a card is face down: `exiled` (CR 406.3), `foretold` (CR 702.143b), and the four CR 708.2 permanent states `manifested` / `morphed` / `disguised` / `cloaked`. `SetFaceDown` / `ClearFaceDown` are the only writers, so "face down with no rule attached" cannot be built. A `Card` has no zone, and every answer below comes off the kind alone — which is what lets them live on `Card` methods with no `*Game` in reach.

**One viewers rule.** `faceDownViewersLocked` is the whole of "who may look": **nobody** for a plain face-down exile (CR 406.3, the exiling player included), the **owner** for a foretold card (CR 702.143d), the **controller** for a CR 708.5 permanent. It is written into `KnownBy`, which stays the single source of truth the wire projection reads, and the face-down landing skips `markCardKnownInZoneLocked` — exile and the battlefield are public *zones*, and that skip is the only thing that makes an object private in one.

**One projection.** A face-down permanent **is** a 2/2 colourless creature with no name, text, subtypes or mana cost (CR 708.2), so it enters at **layer 0** in `printedCharacteristic`. Every later layer and every reader gets it for free: `Effective()`, targeting, "creature you control" predicates, combat, the SBAs and the wire. The four accessors that deliberately bypass `Characteristic` when the layer cache is cold (`HasCardType`, `HasSubtype`, `HasSupertype`, `EffectiveColors`) plus `HasKeyword` consult the same function, so there is one definition of the 2/2 and five reads of it. `PrintedIsCreature` / `PrintedIsLand` stay the real card — they are the CR 707.2 copiable surface.

**One suppression predicate.** `CatalogKey` returns the **empty key** for a face-down permanent, and `catalogDef("")` already means "no entry", so CR 708.2a ("no text") is one `if` at the one place a `Card` becomes a catalog key: no triggered, activated, mana, static or replacement abilities, no cost modifiers, no ETB hook, no printed keywords, no `auto` / `target_mode` on the wire. Exile is deliberately *not* suppressed — a foretold card needs its entry for the cast out of exile (#658).

**One reset point.** `MoveCard` clears the state on every zone change (CR 400.7). That is #697's first bug, and it closes both leaks without either caller learning about face-down.

**Reveals.** CR 708.9 (a face-down permanent that leaves the battlefield) and CR 702.143f's "when its owner leaves" half go through the S22 reveal frame, which marks every seat and moves nothing.

**`ManifestForEffect` ships** (CR 701.40a) — the ordinary library→battlefield entry with one `ZoneEntryOptions.FaceDown` option, which lifts the CR 110.4 nonpermanent refusal, sets the state and its viewers instead of marking the table, and therefore makes the ETB hook and every `EventETB` trigger see the empty catalog key. It ships because the model is untestable without it: a helper that sets three fields proves the projection and nothing about the entry pipeline. **Manifest the mechanic does not ship, and neither does morph** — the `turn_face_up` special action (CR 116.2g), the face-down *cast* (CR 708.4), the morph costs and disguise's ward {2} are #95's.

**Wire and client.** `face_down_kind` is public and labels the card back. `face_visible` is the per-viewer look permission, stamped beside `known_by_you`. The CR 708.2 body survives the #646 redaction for a face-down permanent — an opponent has to see the 2/2 to block it — while everything that names the card underneath (`scryfall_id`, `mana_cost`, `faces`, `layout`, the ability lists) is still stripped, and most of it is never stamped because `CatalogKey` is silent. The client draws the real face plus a badge for the permitted viewer, and a labelled back for everyone else. **Bots** read a face-down permanent as the vanilla 2/2 it is instead of pricing it at the `Unknown` floor, and key their redaction on the look permission rather than on the knowledge set.

**#697's second half:** `StackOverlay`'s hover guard tested `known_by_you === false`, which the server never sends (`omitempty`), so it never fired and an unreadable spell opened a blank zoom panel. The rule moves into `client/src/lib/stackPreview.ts` beside `cardBack.ts`'s, so the overlay and the card cannot disagree.

## ADR

[ADR 0069 — Face-down objects (CR 406.3a, CR 708)](https://github.com/krakenhavoc/cmd_and_ctrl/blob/feat/656-697-face-down-objects/docs/decisions/0069-face-down-objects.md).

Numbering per AGENTS.md §4: every remote ref was listed with `ls-tree docs/decisions/` twice — 270 refs at first writing, 275 immediately before this push. Highest present anywhere was `0064-emblems.md`, then `0065-modal-and-multi-target-clauses.md`. 0069 was free on both passes.

Two corrections the ADR records rather than papering over:

- **Disguise and cloak's ward {2} is not a keyword.** #656's checklist asked where it hangs once the catalog key is blank; the answer is *not* `Characteristic.Abilities`. `canonicalKeywords` is a closed set with no `ward` token, deliberately, because "a bare token has nowhere to put the cost" (`game/keywords.go`, `cards/effects/attachments_batch3.go`). Ward is a structured CR 702.21a triggered ability (`effects.Ward`). ADR 0069 §3 designs it as a `faceDownTriggeredAbilities(kind)` beside the projection and leaves it to #95 — `Ward` lives in the catalog package, which `game` cannot import, and inverting that boundary must not ride along here.
- **The suppression predicate is at `CatalogKey`, not `CatalogAbilityKey`.** The latter is the *documented* "what does this permanent DO" accessor and reads like the right home, but four ability readers deliberately bypass it — the layer pass's static gather, the LTB trigger's last-known identity, `HasKeyword`, and the ETB hook — and each would have needed its own arm.

## Sprint

S43 — Hand special actions and face-down objects (#886).

## Issues

Closes #656
Closes #697
Refs #886, #658, #659, #95

## Test plan

Everything below was run on this branch at its final rebase onto `origin/develop` (`ce954f74`, the modal-clauses merge #937). The one rebase conflict was `docs/engine-seams.md`, where #937 retired the modal-clause row next to the face-down one; resolved by keeping #937's deletion and this branch's rewritten face-down row.

- [x] `go build ./...`, `go vet ./...` — clean
- [x] `go test ./...` — all packages pass
- [x] `AISEAT_GAME_TESTS=1 go test ./internal/...` — pass
- [x] `make lint` (golangci-lint v1.64.8 via `go run`) — clean; `gofmt -s -l .` empty
- [x] Client `npm run lint` (ESLint + Prettier), `npm run check` (svelte-check: 482 files, 0 errors, 0 warnings), `npx vitest run` (61 files, 1000 tests)

**New engine tests** (`server/internal/game/face_down_test.go`, one section per ADR decision):

- [x] the viewers table: a plain face-down exile is known to nobody including the exiling player, a foretold one to its owner alone, a manifested permanent to its controller alone on a public zone
- [x] the CR 708.2 body reaches every reader — `Effective()` name / P/T / types / abilities, `CurrentPower` / `CurrentToughness`, `IsCreature`, `HasSubtype`, `HasSupertype`, `IsColorless`, `HasKeyword` — while `PrintedIsCreature` and the printed type line stay the real card
- [x] a face-down card in **exile** gets no 2/2 body and keeps its catalog key (the cast out of exile needs it)
- [x] a face-down permanent's triggered / static / activated / mana / replacement abilities and printed keywords all come back empty, and it is not flagged `unimplemented`, against a fixture whose catalog entry is loud face up
- [x] `ManifestForEffect` manifests the top card, accepts a **nonpermanent** card (CR 701.40a) and the object is a creature
- [x] CR 708.9: a face-down permanent bounced to a hand is revealed to the table and emits `EventRevealCards`; a face-down **exile** moved to a hand is not
- [x] the reveal runs BEFORE the destination's knowledge rule, which is the one place the two disagree: a face-down permanent tucked into a library is revealed to the table *and* then unreadable by everyone (CR 401.2). Reveal last would have left every seat able to read a library card by position — caught in review and pinned
- [x] CR 702.143f: a leaver's face-down exiles are revealed
- [x] `SetFaceDown` / `ClearFaceDown` keep the flag and the kind in step; a kindless `FaceDown: true` is not a CR 708.2 permanent
- [x] undo across a manifest restores the state *and* the 2/2
- [x] an old snapshot with `faceDown: true` and no kind restores as `FaceDownExiled`

**New lifecycle tests** (`server/internal/game/face_down_lifecycle_test.go`, #697):

- [x] `move_card` from a face-down exile to hand, graveyard and battlefield, and a route bounce, all land face up
- [x] a card cast out of a face-down exile reaches the stack face up (CR 406.3a) — foretell's own path, built with a hand-made grant because nothing in the catalog can make one yet

**New protocol tests** (`server/internal/protocol/face_down_view_test.go`):

- [x] a face-down permanent ships its public CR 708.2 body (`type_line` "Creature", 2/2, no name, no colours, no abilities) to an opponent while `scryfall_id`, `mana_cost`, `faces`, `auto`, `target_mode`, `unimplemented` and both ability lists stay stripped; the controller additionally gets the art and `face_visible`
- [x] `face_visible` follows the viewers table for both exile kinds, and a face-down **exiled** card ships no body at all
- [x] #646's zone × viewer allowlist table still passes, with `face_down_kind` added to the allowlist and the reflection guard satisfied

**New bot tests:**

- [x] `heuristic`: a face-down permanent is priced exactly as an identical vanilla 2/2 and counts as a creature; a face-down *exiled* card is still `Unknown`
- [x] `model`: the prompt describes a face-down permanent (the seat's own included) as "a face-down 2/2 creature", and never names a face the seat may not look at even when `known_by_you` is set

**New client tests:**

- [x] `stackPreview.test.ts` (7 tests): the guard refuses an absent `known_by_you`, a redacted stack spell, an explicit `false`, a missing card and an opponent's morph; it allows a known card and a face-down card the viewer may look at
- [x] `faceDownCard.render.test.ts` (6 tests): an opponent's morph draws a back labelled with the public kind and never requests art; the controller's own manifest draws the real face plus the badge; an ordinary card carries no badge

**Each fix fails with itself backed out** — verified by reverting the hunk and re-running:

- [x] removing `c.ClearFaceDown()` from `MoveCard` fails all four `move_card` rows and the cast row
- [x] removing the `printedCharacteristic` guard fails 8 assertions in the projection test
- [x] removing the `CatalogKey` guard fails the suppression test
- [x] removing the wire re-stamp and `face_visible` fails 3 protocol assertions
- [x] moving the CR 708.9 reveal back after the destination's knowledge rule fails the library-tuck test
- [x] restoring the old bot readers fails all 3 prompt tests and both heuristic assertions
- [x] restoring `known_by_you === false` in `previewableCard` fails 3 of its 7 tests

### Not run / known

- **No browser check.** Nothing has looked at the rendered badge in a running client; it is covered by a render test only.
- **No e2e / Playwright.**
- **Census not regenerated** — `TakeCensus` counts catalog specs and `Completeness` declarations only, and no catalog card changed. `go test ./internal/cards/coverage/` passes, which is the guard.
- **`TestBotsStopPassingWhenTheLoopBreakerFires` / `TestRunnerObserverRecordsTheDeclineToPass` are load-flaky on this machine**, independent of this branch: both are 30 s wall-clock waits, and with a dozen agents running suites in parallel they time out intermittently. Confirmed by an interleaved A/B (6 runs each, alternating) of this branch against a pristine `origin/develop` tree: this branch 0/6 failures, develop 1/6. Both pass in the final full `go test ./...` above. This is the aiseat test race already being fixed separately.

## Out of scope / follow-ups

- **An LTB trigger on a face-down permanent still fires** (`server/internal/game/triggers.go:481`). The harvest reads the card *after* `MoveCard` has cleared the state, so `CatalogKey` answers again and a face-down permanent's real dies-trigger would go on the stack. Nothing in the engine can make a face-down permanent with a dies-trigger today, and the fix belongs with the last-known-information work the harvester already does for CR 603.10. Recorded in ADR 0069's Consequences.
- **The game-end half of CR 702.143f** (reveal all foretold cards when the game ends) has no game-end sweep to hang on; left to #658. The "owner leaves" half is built (`server/internal/game/leave_game.go:55`).
- **Disguise / cloak ward {2}** — designed in ADR 0069 §3, deferred to #95 with the rest of the morph family.
- **Spectators and admins still see every card**, face-down objects included (`server/internal/protocol/view.go`, `FilterViewFor`). #656 asked; ADR 0069 §2 keeps it — a spectator view is a debugging surface, not a seat.


irobinson010:
CI note: the two failures on this PR so far are the self-hosted runner running out of disk (`##[error]No space left on device` in **Set up job**, before checkout). Not this branch — `develop` itself and several other in-flight branches failed the same way in the same window, and `ci/runner-disk-cleanup` is running. I will re-run once that lands. Everything the workflow does was run locally against this exact tree at its rebase onto `ce954f74`: `go build`/`go vet`/`go test ./...`, `AISEAT_GAME_TESTS=1 go test ./internal/...`, `make lint` (golangci-lint v1.64.8), `gofmt -s -l` empty, and client `lint` / `check` / `vitest` (61 files, 1008 tests).

Issue #754, opened 2026-09-17
ADR then implementation: abilities granted to other permanents (CR 113.10, layer 6) — Cryptolith Rite, Rishkar, Gemhide Sliver, Necrotic Sliver


krakenhavoc:
## Why

Nothing can give a permanent an activated, mana or triggered ability it doesn't print. Layer 6 grants keywords only. Checked on develop (`ad9413f`):

- `ActivatedAbilitiesForCard(c Card)` (`server/internal/game/activated.go:242`) returns the instance's own `c.ActivatedAbilities` or the catalog entry for `CatalogAbilityKey(c)`. `ManaAbilitiesForCard(c Card)` (`game/mutations.go:3626`) does the same plus the CR 305.6 land-type abilities. Neither takes the game or a layer slot, so no effect can add to them.
- `Characteristic` (`game/characteristic.go:33-62`) has `Abilities []string` (keyword tokens) and `AbilitiesRemoved`. There is no slot for an ability shape.
- Triggers come only from the permanent's own key: `CatalogTriggers` at `game/triggers.go:298`, `:338`, `:414`, `game/simultaneous.go:135` and `game/sagas.go:71`.
- The accessors are the single choke point: `legal/abilities.go:54` and `:378` and `protocol/view.go:1718`, `:2763`, `:2970` read through them. A fix there reaches the client and the bot.

What already works, and is **not** this seam:
- A static "has 'whenever …'" grant, with the trigger on the granting card watching the recipient: Dionus, Elvish Archdruid (Full), Agent of the Iron Throne (caveats), and `WardAttached` (`cards/effects/attachments_batch3.go:231`). The only approximation is who controls the trigger.
- Springleaf Parade puts a mana ability on its own token template (`cards/effects/batch18_helpers.go:26-30`). That can't reach permanents that already exist.

Still missing: activated and mana grants to other permanents, including other players' Slivers, and a triggered ability granted until end of turn by a spell that has already resolved (CR 611.2c). Nothing is left to carry that trigger, and `DelayedTrigger` fires only on step entry (#663). `server/internal/cards/effects/batch31_test.go:133` keeps Manaweft Sliver out of the catalog until this lands. ADR 0040 (`docs/decisions/0040-mana-pipeline.md:211`) names the gap for Chromatic Lantern and The World Tree.

## Cards

[...]

Issue #750, opened 2026-09-17
ADR then implementation: conditional blocking restrictions (CR 509.1b) — "can't be blocked except by…", "can't block X", and blocker-count limits; follow-on to #705's game-aware CanBlock


krakenhavoc:
## Why

The block check has only on/off bits and flying. Checked on develop (`ad9413f`):

- `CanBlock(attacker, blocker *Card)` (`server/internal/game/keywords.go:310-330`) can't see the game. It reads `Restricted(attacker, CantBeBlocked)`, `Restricted(blocker, CantBlock)` and flying against flying or reach, and nothing else.
- Three callers must always agree (ADR 0045 §3): `DeclareBlocker` (`mutations.go:4304`, check at `:4340`), the move enumerator `legal.combatMoves` (`legal/combat.go:104`) and the auto-pass signal `seatOwesBlockDecisionLocked` (`blockers.go:99`, check at `:135`).
- `Restriction` is a set of bits with no parameters (`restrictions.go:95-110`). A condition on the *other* creature (its power, subtype, colour, keyword) can't be a bit. The file (`restrictions.go:55-83`) and ADR 0045 ("What this deliberately does not express") both leave these out and put their home in `CanBlock`, "once CanBlock can see the game".
- **Block-count limits:** `BlockerCountValid` (`keywords.go:345-356`) knows only menace (at least 2), and only tests call it. Menace is actually enforced by silently undoing a lone block at the start of the damage step, in `assignAndDealCombatDamageLocked` (`mutations.go:4506-4540`), not when blockers are declared. ADR 0045 and `restrictions.go:69-71` both say `BlockerCountValid` does that close-out, which is wrong. A maximum (Hungering Hydra: "can't be blocked by more than one creature") has nowhere to go.

**Relationship to #705.** #705 (landwalk, open) already has the checklist item "Give the block-legality check access to game state … without breaking the legal-move enumerator's read-only contract". That is the `CanBlock(g, …)` signature change, so this issue doesn't repeat it and starts from it. The audit's order is: #705 changes the signature and ships landwalk, then this issue, then protection (#662, CR 702.16f) adds its check in the same function.

## Cards

17 cards where this is the only core blocker and 26 with it as any core blocker. The verifier estimated about 20% of the sampled cards are also blocked by something else (≈14 real).

[...]

Issue #744, opened 2026-09-17
ADR then implementation: dice and coin flips (CR 705, 706) — a public RNG for effects, random picks, and what undo does to a roll


krakenhavoc:
## Why

The game has a seeded, persisted RNG, but no card effect can reach it. There is no die roll, no coin flip, no general "at random" pick, and no roll or flip event. Evidence on develop (`ad9413f`):

- `g.rng` / `g.rngState` (`server/internal/game/game.go:302`, set up at `:581-582`) are saved and restored by `snapshot.go:668` and `:1119` (ADR 0041 decision 3).
- Only three effect paths draw from it: `DiscardRandomForEffect` (`effect_api.go:329`, `g.rng.IntN` at `:342`), `ShuffleLibraryForEffect` (`effect_api.go:1882`), and the search shuffle (`effect_api.go:1523`).
- Cascade's `bottomInRandomOrderLocked` (`cascade.go:247`) is private and only moves cards **out of exile**.
- `git grep` finds no `RollDie`, `FlipCoin`, `EventRoll` or `EventCoin` anywhere in `server/`.
- The one workaround is `b27PseudoRandomIndex` (`cards/effects/batch27_helpers.go:253`), which hashes the latest event sequence number. Urza's Bauble ships with a caveat because of it (`urzas_bauble.go:22-27`, caveat at `:33`).

**Undo.** `cloneLocked` shares the RNG pointer with every undo snapshot (`clone.go:41-50`: "Undo therefore does NOT rewind the random stream, which is deliberate and long-standing"), and `RestoreFrom` keeps it (`clone.go:517-518`). `Room.Undo` (`ws/room.go:363`) is gated only by the caller and the per-turn `UndosRemaining` budget. Nothing blocks an undo after hidden or random information comes out. So undoing a shuffle or a random discard and doing it again already gives a new result. With dice, the same thing becomes "undo until you roll a 20".

No catalog card rolls or flips today (Ancient Copper Dragon, Hoarding Ogre, Vexing Puzzlebox and Deadbridge Chant are all absent from `cards/effects`).

## Cards

51 cards where this is the only core blocker, 62 with it as any core blocker. The verifier estimated 45–50% of the sampled sole-blocker cards are also blocked by something else (≈26 real).

- Ancient Copper Dragon, Ancient Gold Dragon, Ancient Silver Dragon, Hoarding Ogre: a d20 into Treasures or counters. Nothing else missing.
[...]

Issue #662, opened 2026-09-16
ADR then implementation: protection (CR 702.16) — DEBT checked against the source object; supersedes ADR 0038 §7


krakenhavoc:
## Why

Protection (CR 702.16) isn't implemented. Nothing reads it, and `canonicalKeywords` (`server/internal/game/keywords.go:55`) has no protection token. It is four checks against the **source** (DEBT):

| check | rule | what the hook has on develop (`7c1ae9b`) |
|---|---|---|
| **D**amage from a source with the quality is prevented | CR 702.16e | `RepEventDamage` carries only `DamageSource` (an ID). The source may already be gone, e.g. an attacker that died to first-strike damage; see `game/mutations.go:4694-4702` |
| **E**nchanted / equipped / fortified by something with the quality | CR 702.16c-d | `attachmentLegalLocked(c)` gets the Aura or Equipment itself (`game/attach.go:217`), but catalogued Auras return early at `:244-247`, before any new check |
| **B**locked by a creature with the quality | CR 702.16f | `CanBlock(attacker, blocker *Card)` gets both objects (`game/keywords.go:310`) |
| **T**argeted by a spell with the quality, or an ability from a source with it | CR 702.16b | `CanBeTargetedBy(c, zone, caster uuid.UUID)` (`keywords.go:138-148`) and `specMatchLocked(caster, …)` (`game/targets.go:320`) get only the **controller's ID** |

So the targeting path is the real refactor. ADR 0038 §7 (`docs/decisions/0038-protection-style-keywords.md:141-147`, and again `:235-241`) says so but cites CR 702.16e; the targeting rule is **702.16b**. One targeting call site has no source object at all: the spell-copy re-target resume (`game/spell_copy.go:243`) runs before the copy exists, and the original may have been countered. That frame keeps a value copy for this reason.

Catalog cards shipping without their protection, as declared caveats:
- Baneslayer Angel: from Demons and from Dragons, a subtype quality (changelings count)
- Sword of Fire and Ice: red and blue
- Sword of Feast and Famine: black and green
- Animar, Soul of Elements: white and black

Waiting, not in the catalog:
[...]

PR #949, opened 2026-09-18
feat(cards): batch 41 — 46 cards from EDHREC ranks 4254–4353

krakenhavoc:
Roadmap batch 41 (`edhrec_rank` 4254–4353). **46 cards registered**, 40 full and 6 with a declared caveat.

Scope was the 2026-09-17 status comment's *Ready today* (55) and *Ready with a declared caveat* (16) groups — 71 unregistered cards. There is no separate "Re-triage against today's engine" comment on #448; that status comment is already the current triage against `develop`. All 71 were checked against `effects.All()` first: none were already registered.

## Declared caveats (#259 — weaker than printed, never stronger)

| Card | Ships without |
|---|---|
| Hanweir Garrison | meld with Hanweir Battlements |
| Hordewing Skaab | loots once, not once per damaged opponent |
| Canopy Cover | the conditional block restriction (#750). The hexproof half is **exact** — CR 702.11b defines hexproof as that sentence verbatim |
| Sword of Body and Mind | protection from green and blue (#662) |
| Spectral Deluge | foretell (#658) |
| Nighthowler | bestow |
| Goblin Ringleader | the rest go to the bottom in a random order, not one you choose |

Canopy Cover is the #259 rule doing its job: shipping the unconditional "can't be blocked" would have been **stronger** than printed, so the evasion clause ships as nothing at all and a test pins that it stays nothing.

## Notable implementations

- **Diffusion Sliver** reuses the CR 702.21a ward machinery with its `AppliesTo` widened from "this permanent" to "a Sliver creature you control" — same trigger, same pay-or-counter prompt charged to the *spell's* controller, same per-target-instance firing (CR 115.7).
- **Ossification** is the Oblivion Ring shape: an entry trigger that exiles and a leave trigger that returns, keyed on a shared stack label through `b27ExiledWith`. The permanent returns under its **owner's** control (CR 610.3) as a new object (CR 400.7).
- **March of the World Ooze** declares its two static clauses in two layers — 7b for "base power and toughness 6/6" and 4 for "are Oozes in addition to their other types" — so a creature with two +1/+1 counters under it is an 8/8, not a 6/6.
- **Grumgully** is a real CR 614 entry replacement, so the persist interaction (CR 704.5q) works and Doubling Season applies to it.
- **Undead Butler** uses a CR 603.12 reflexive trigger: the "you may" belongs to the parent, and the reflexive's target is chosen after the Butler has already left the graveyard, so it can never return itself with no "another" clause needed.

## Shared code

`batch41_helpers.go` behind the `b41` prefix (#231). Four rows added to `tokens_table.go`; Phyrexian Rebirth's X/X Horror is built in the helpers because its size is decided at resolution and a table keyed on printed characteristics has no key for it.

The clone detector flagged three real duplications and all three were removed rather than baselined: Angel of Finality now shares Bojuka Bog's "exile target player's graveyard" body, Ossification uses Duplicant's `b27ExileChosenTarget`, and `b41HasACounter` is the negation of Damning Verdict's `b10NoCounters`.

## Skipped, with reasons

**25 of the 71 in scope were skipped.** The ones worth calibrating on:

*Blocked by machinery the re-triage did not account for:*
- **Horn of the Mark, Ao the Dawn Sky, Sandstone Oracle, Chaos Wand** — all four need a look-at-then-choose flow the engine does not have for the shapes they print (a multi-select bounded by total mana value; a "choose an opponent" prompt; a free cast from exile).
- **Amazing Acrobatics** (choose one or both, two targeted modes), **Suplex** ("if it would die this turn, exile it instead" is a replacement with a turn duration), **Level Up** and **Dancer's Chakrams** (grant a *triggered/static ability* to the attached creature, #754), **Master Chef** (grants two replacement abilities to commanders, and the count stacks per commander), **Dance with Calamity** (unbounded player-driven exile loop plus free casts), **Foil** (a two-card alternative cost), **Verity Circle** ("becomes tapped … if it isn't being declared as an attacker" needs a tap event that knows about declaration), **Jaws of Defeat** (the rules meaning of "the difference between its power and its toughness" is not settled by a CR paragraph I could cite — skipped rather than guessed), **Tato Farmer** (rad counters, and "a land milled this turn"), **Minn** (the token carries a static ability, #521), **Riveteers Ascendancy** (once-each-turn limiter plus a relative-mana-value clause), **Shadow of the Goblin** ("played from anywhere other than your hand" for a *land*), **Baeloth** (goad enforcement), **Highway Robbery** / **Pippin** / **Lozhan** / **Demonic Junker** / **Replication Technique** / **Ob Nixilis Reignited** — each needs a second mechanic beyond its filed blocker.

**Findings the re-triage should absorb:** every card in the batch printing "look at the top N, take some, put the rest on the bottom" is blocked by the *same* missing move, not by #744's ordering question — there is no library→hand move for a *named set* of revealed cards, and the library-search API shuffles and prompts over the whole library. Goblin Ringleader was writable only because `BounceToHand` happens to route from any zone (the same trick `b21RevealUntilBasicLandToHand` uses); Horn of the Mark and Ao need the multi-select on top of it.

## CI

GitHub Actions is not dispatching runs at the moment (self-hosted runner issue, unrelated). All gates were run locally in the devcontainer against the **rebased** branch:

- `gofmt -l` clean, `make fmt-check` clean
- `make vet` clean
- `make lint` (golangci-lint v1.64.8 + staticcheck) clean
- `go test ./internal/cards/... ./internal/game/` — all green
- `go test ./internal/cards/coverage/ -update` regenerated: `len(effects.All())` 1737 → 1783

Branch is rebased on `origin/develop` @ `35c487d3`; `git rev-list --count HEAD..origin/develop` is 0 and the diff against develop is additive (4092 insertions, 13 deletions — the deletions are the clone removals above plus the census lines).

Expect a conflict on the census block in `card-coverage-roadmap.md` — every parallel batch regenerates it. The fix is to re-run `go test ./internal/cards/coverage/ -update` after merging, never to resolve it by hand.

Refs #448


krakenhavoc:
Superseded by #966, which merged as `c63ecae8`. Batch 41's cards landed there in full.

These five card PRs each regenerate the catalog census and the clone baseline, so merging any one invalidated the other four — and `develop` was taking engine merges faster than a single card PR could re-verify against it. #966 merged all five as one integration, resolved the census and clone baseline once by regenerating, and went green on the combined result.

Three problems only showed up in combination and are fixed in #966: a duplicate `tokens_table.go` map key from two batches adding the same token, Shivan Devastator missing the `XMatters: true` that #946's new guard requires, and `unparam` firing once batch 40 added the catch-up land cycle's fifth caller.

Closing this without merging. The branch is left in place.

Issue #743, opened 2026-09-17
ADR 0020 addendum then implementation: activation conditions (CR 602.1b, 602.5) — "Activate only if …" and "only during your turn" on non-mana activated abilities


krakenhavoc:
## Why

A non-mana activated ability can carry only one activation restriction: sorcery speed. Nothing can say "Activate only if an opponent controls four or more lands" (Tectonic Edge), "only during your turn" (Sanctum of Eternity) or "only if this creature is attacking". Evidence on develop (`ad9413f`):

- `game.ActivatedAbilityShape` (`server/internal/game/activated.go:211-231`) has Label, Cost, Targets, SorcerySpeed and Effect. `effects.ActivatedAbility` (`cards/effects/spec.go:467-473`) has the same five fields.
- `ActivateCatalogAbility` (`activated.go:335`) gates timing only on `SorcerySpeed` or a loyalty cost (`:384`), and the legal enumerator mirrors that (`legal/abilities.go:56`).
- The view has `sorcery_speed` and nothing for a condition (`protocol/view.go:1049`).

The shape to copy already exists for **mana** abilities:
- `ManaAbilityShape.Condition func(g, controller, source) bool` (`game/effect_hooks.go:183-197`, #352), checked before any cost in `ActivateManaAbility` (`mutations.go:3373-3374`), in the auto-tapper (`mutations.go:1190`, `autotap.go:189`) and in the enumerator (`legal/abilities.go:383`).
- It returns `ErrConditionNotMet` (`errors.go:267-272`), and `effects.ControlsAtLeast` (`cards/effects/mana_derivation.go:274`) builds conditions.
- `Game.TurnTally` (#586), `EventsThisTurn()` (`turn_tally.go:121`) and `StartingLife` already give most conditions something to read.

Shipped cards that settle for less:
- **Sanctum of Eternity** uses sorcery speed for "during your turn" and declares a caveat (`cards/effects/sanctum_of_eternity.go:21-27`, caveat at `:33`).
- **Glint-Horn Buccaneer** leaves out its "only if this creature is attacking" ability (`glint_horn_buccaneer.go:14-18`, `:24`). It also needs a discard-a-card cost (#660).
- **Beledros Witherbloom** leaves out "Activate only once each turn" (`beledros_witherbloom.go:27-34`). That is the activation-tally half, not this gate.

## Cards

[...]

Issue #660, opened 2026-09-16
Cycling (CR 702.29): activated abilities from hand with a discard-this-card cost — six catalog cards wait on it


krakenhavoc:
## Why

Cycling (CR 702.29) is an **activated ability that works only while the card is in hand**: "[Cost], Discard this card: Draw a card." Typecycling (CR 702.29e) searches for a card of the named type instead. The #94 body and several card comments call it "an alternative cast path"; it isn't one.

Two engine seams block it. Both are already registered, and neither names the cycling cards:
- **No discard cost component.** `AbilityCost` (`server/internal/game/activated.go:35-134`) can't discard. See `docs/engine-seams.md:38` ("Discard-a-card cost on an activated ability": Fauna Shaman, Tortured Existence, Cryptbreaker, Survival of the Fittest).
- **Abilities can't be activated from hand.** `ActivateCatalogAbility` finds its source on the battlefield only (`activated.go:314`). The bot enumerator loops over the battlefield (`legal/abilities.go:38`), and the view projects abilities for battlefield cards only (`protocol/view.go:1599`). See `docs/engine-seams.md:40`.

And one rules detail: "When you cycle this card" triggers from **whatever zone the card ends up in**, normally the graveyard, not from the hand (CR 702.29c). "Cycles or discards" triggers only once per cycle (CR 702.29d). The harvester has no scan that finds a just-cycled card in the graveyard, and `EventDiscardCard` carries no "cycled" marker (`game/additional_cost.go:183`).

Six catalog cards ship without cycling. Five declare it in `Caveats`:
- Raffine's Tower (`raffines_tower.go:28`)
- Ketria Triome (`ketria_triome.go:41`)
- Marauding Mako (`marauding_mako.go:21`)
- Scrounging Skyray (`scrounging_skyray.go:22`)
- Magmakin Artillerist (`magmakin_artillerist.go:28`), which also has a "when you cycle" trigger

The sixth, Sylvan Reclamation, omits basic landcycling with no `Completeness` set (`sylvan_reclamation.go:10-13`), so it counts as Unreviewed.

Also buildable once this exists: **Decree of Pain** ("When you cycle this card, all creatures get -2/-2 until end of turn"). Drop the #94 body's other picks:
- New Perspectives: no cycling of its own; "pay {0} rather than pay cycling costs" is the #625 seam.
- Decree of Justice: its cycle trigger needs a resolution-time pay {X}.
[...]

Issue #450, opened 2026-09-13
Cards batch 43: EDHREC rank ~4456–4555 (100 cards)


krakenhavoc:
**Umbrella:** #293 — Card catalog coverage: the next 2000 Commander cards by play rate
**Batch 43 of 60** · `edhrec_rank` 4456–4555 · 100 cards, rank order.

Cards the catalog does not have, ranked by how often Commander decks actually
play them. Source and method are on the umbrella issue; nothing here was
recalled or guessed — one pass over `data/scryfall/default-cards.json`
produced every row.

**Notable in this batch:** 37 of the 100 need no new machinery; the dominant blocker is **cost modification** (12 cards); 47 creatures, 9 lands.

## Mechanic triage

Grouped by the machinery each card needs, ordered by how many cards each
group buys. This is the part that makes the batch actionable — the table
further down is only the shopping list.

### No new machinery — 37 cards

Expressible today with the primitives already on `main` (`Spec.Targets` /
`Modes` / `Static` / `Replacements` / `Triggered` / `Activated` /
`ManaAbilities` plus the `*ForEffect` helpers). Start here.

> Naya Panorama · Gigantosaurus · Ifnir Deadlands · Urza's Factory · Boon of the Spirit Realm · Éowyn, Shieldmaiden · Darien, King of Kjeldor · Peek · Archetype of Courage · Angelic Field Marshal · Raise the Alarm · Burning-Rune Demon · Caldera Pyremaw · Flood of Tears · Dreadbore · Senu, Keen-Eyed Protector · Rundvelt Hordemaster · Nightshade Harvester · Waterbending Lesson · Tower Winder · Shaman of the Pack · Seafloor Oracle · Finneas, Ace Archer · Captain Sisay · Labyrinth of Skophos · Voidslime · Kalain, Reclusive Painter · Indomitable Ancients · Baron, Airship Kingdom · Zuko's Exile · Electrickery · Desert · Dragonlord Atarka · Nemesis of Reason · Nyx-Fleece Ram · Planar Bridge · Everything Comes to Dust

### Needs a named mechanic that does not exist yet

Each card is filed under its *dominant* blocker; a card needing two
pieces is listed once and its other blockers show in the table's
"blocking machinery" column.

#### Cost modification, alternative casts and costs computed at activation — 12 cards — tracking #93

[...]

PR #955, opened 2026-09-18
feat(cards): roadmap batch 43 — 47 cards from ranks 4456–4555

krakenhavoc:
## What this is

Roadmap **batch 43** (#450, `edhrec_rank` 4456–4555). Scope was the union of the issue's 2026-09-17 re-triage lists — *Ready today* (63) and *Ready with a declared caveat* (3), 66 candidates — not the older "No new machinery — 37" group in the issue body.

**47 registered** (38 `full`, 9 `caveats`), **19 skipped** with a reason each. None of the 66 turned out to be already in the catalog.

## Three cards exist only because #937 landed mid-branch

The branch was cut from `develop` before **#937 — modal and multi-target clauses (ADR 0065)** merged. Rebasing onto it during the work unblocked three cards that were going to be skips, and each is a first of its kind in the catalog:

| Card | What #937 gave it |
|---|---|
| **Skemfar Shadowsage** | A modal **triggered** ability. The mode is chosen as the trigger goes on the stack (CR 603.3c). Doing it at resolution instead would have made the card *stronger* than printed — more information before the choice — which is the #259 direction, so it was a skip until this landed. |
| **Gimli's Reckless Might** | Two target **clauses** with different predicates on one trigger: "target attacking creature you control" fights "up to one target creature you don't control". |
| **Lost Jitte** | A modal **activated** ability, three bullets carrying three different target clauses, paid with a charge counter at announce. |

## Skips, and why

**Still blocked, and the re-triage was wrong about two of them** — this is the calibration data #293 wants:

| Card | Listed as | Actually needs |
|---|---|---|
| **Goblin Rabblemaster** | ready | "Other Goblins you control attack each combat if able" is an attack **requirement** (CR 508.1d). `game.Restriction` has prohibitions only — `CantAttack`, `CantBlock`, … — and no requirement bit. Dropping the clause removes a **drawback**, which makes the card stronger than printed (#259), so it is a skip rather than a caveat. |
| **Glowing One** | ready | **Rad counters.** Poison counters exist; rad counters do not, and the card is two-thirds rad. |

The other 17 were listed as blocked and stayed blocked, or were ready-but-thin and are better left for the mechanic that unblocks them: Birthing Ritual, Octomancer, Banquet Guests, Burning-Rune Demon, Descendants' Fury, Endless Evil, City of Death, Everything Comes to Dust, Nerd Rage, Overlord of the Boilerbilges, Bone-Cairn Butcher, Prismari the Inspiration, Sokrates Athenian Teacher, Waterbending Lesson, Bilious Skulldweller, Rundvelt Hordemaster, Cloud Ex-SOLDIER.

## Declared caveats (9)

All weaker than printed; none stronger. Written for a player deciding whether to sleeve the card.

- **Agonasaur Rex** — cycling is not implemented (#660), so the cycling draw and its counters never happen.
- **Archetype of Courage** — only the first half. A plain layer 6 removal for the opponents' clause would silently stop working against a first striker played *after* the Archetype (CR 613.7 timestamp order), which is worse than not implementing it, because it looks like it works.
- **Cryptcaller Chariot** — a multi-card discard makes its Zombies one trigger at a time (CR 603.1 batching).
- **Desert** — see the engine finding below.
- **Dragonlord Atarka** — the 5 damage is divided evenly rather than as the player chooses. Fury's caveat exactly; #937 explicitly left divided damage out of scope.
- **Interplanar Beacon** — the `{1}, {T}` filter is omitted. The mana grammar cannot say "two mana of **different** colors", and two independent pipe picks would produce `{U}{U}`, which is *stronger* than printed.
- **Restart Sequence** — freerunning omitted for the same reason: the alternative-cost slot carries a price, not a gate, so an ungated `{1}{B}` offer would be stronger than the printed card.
- **Tower Winder** — the search looks in your library only; there is no combined library-and-graveyard search.
- **Vizkopa Guildmage** — only the lifelink half. The drain installs an **event-keyed floating trigger** for the turn; the engine's two duration mechanisms are turn-scoped *statics* (`RegisterScopedStaticForEffect`) and *step-keyed* delayed triggers (`ScheduleDelayedTriggerForEffect`), and neither can express "watch an event kind until end of turn".

## Engine finding (not fixed here)

**Combat is cleared one step too early, CR 511.3.** `Game.runStepEntryHooksLocked` calls `clearCombatLocked` on **entering** `end_combat`; CR 511.3 removes creatures from combat as that step **ends**. The existing comment says the clear was already deferred once so the client keeps its combat arrows through the damage steps — it just stopped one step short.

Consequence: there is no attacking creature *during* the end of combat step, so **Desert**'s "target attacking creature… activate only during the end of combat step" has no legal target and the enumerator never offers it. Anything else phrased "at end of combat, …target attacking creature" has the same problem.

Left unfixed on purpose: moving the clear to the step's exit is a cross-cutting combat change and does not belong in a card batch. Desert's ability is **declared anyway**, on Darksteel Citadel's posture — the day the clear moves, the card starts working with no change to its file — and the caveat says so in plain words. Worth its own issue.

## Shared-vocabulary changes

`TestNoNewExactClonesInTheCatalog` caught three duplicate bodies and all three were folded rather than baselined away:

- `IsBasicLandOfAnySubtype` in `helpers.go` — the New Capenna Overlook cycle and the new Alara Panorama cycle.
- `lookAtTargetPlayersHandThenDraw` in `helpers.go` — Gitaxian Probe and Peek.
- `b33PutChargeCounterOnSelf` — Coalition Relic and Lost Jitte (doc comment widened).

`DuringStep` joins `activation_conditions.go` with its first card, deliberately *not* `SorcerySpeed`: a sorcery-speed gate means "your main phase, empty stack", which would forbid exactly the window Desert's clause opens (#743).

Five new rows in `tokens_table.go`: the Assembly-Worker, the Moogle, the hexproof Merfolk, the trample-and-haste Human Knight, the hasty Goblin.

## Gates

Run in the devcontainer image with the **whole repo** mounted (so `census.go` can read the roadmap), on this branch **rebased onto `develop` at 35c487d3**:

- `gofmt -l` clean
- `make lint` (golangci-lint v1.64.8) clean
- `go test ./internal/cards/... ./internal/game/` — green
- `go test ./internal/cards/coverage/ -update` re-run; registry keys 1737 → **1784** (+47)

GitHub Actions is not dispatching runs on this repo right now (self-hosted runner issue, unrelated), so the local devcontainer gates above are the real signal and they are green.

**Expect a conflict on the census block** in `docs/decklists/card-coverage-roadmap.md` — every parallel batch regenerates it. The fix on conflict is to re-run `go test ./internal/cards/coverage/ -update` after merging, never to hand-edit the table.

Refs #450


krakenhavoc:
Superseded by #966, which merged as `c63ecae8`. Batch 43's cards landed there in full.

These five card PRs each regenerate the catalog census and the clone baseline, so merging any one invalidated the other four — and `develop` was taking engine merges faster than a single card PR could re-verify against it. #966 merged all five as one integration, resolved the census and clone baseline once by regenerating, and went green on the combined result.

Three problems only showed up in combination and are fixed in #966: a duplicate `tokens_table.go` map key from two batches adding the same token, Shivan Devastator missing the `XMatters: true` that #946's new guard requires, and `unparam` firing once batch 40 added the catch-up land cycle's fifth caller.

Closing this without merging. The branch is left in place.

Issue #691, opened 2026-09-16
Printed 0/0 X-counter creatures survive an X=0 cast: player-facing caveats (and fix Wildwood Scourge's comment)


krakenhavoc:
## Why

Creatures that enter with X +1/+1 counters and are printed 0/0 survive being cast for **X=0** (or entering with no counters), because the toughness SBA's placeholder skip treats "toughness 0, no counters, never had any" as a placeholder. In paper they die immediately (CR 704.5f). #683 fixes the "lost its last counter" case but keeps the never-had-a-counter skip, so this remains.

Mikaeus, the Lunarch gets a player-facing caveat for it on the #625 branch (`2524db3`). The others only mention it in Go comments, or not at all.

**Decision (2026-09-16):** track player-facing caveats for the rest, and fix a false comment.

## Scope

- [ ] Add a player-facing caveat ("Cast for X=0 it stays on the battlefield as a 0/0 instead of dying", worded per card) to each affected catalogued card, after checking its oracle text and how it enters. Candidates from the #683 review:
  - Hangarback Walker
  - Goldvein Hydra
  - Primordial Hydra
  - Lifeblood Hydra
  - Hydroid Krasis
  - Wildwood Scourge
  - Multani (no-land case)

  Grep the catalog for other printed-0/0 X-counter creatures.
- [ ] **Wildwood Scourge's Go comment is false:** it says a Scourge cast for X=0 "dies to the toughness check, as in paper". It never had a counter, so the placeholder skip keeps it. Fix the comment and add a test pinning the real behaviour.
- [ ] Update `Completeness` where a card was declared complete but has this gap (#259: undeclared stronger-than-printed).
- [ ] Longer term: a real fix, e.g. mark "entered with X counters, X=0" or give imported printed-0/0 creatures a non-placeholder flag. It could share the design with the CDA issue and #683's `VariableToughness`. Record that option; the caveats are the immediate scope.

Related: #683, #625, #259.

Sprint: S27 follow-up

irobinson010:
Claiming this together with #690 on branch `fix/690-691-cda-toughness-sba-and-x-zero` (base `develop`). Engine mechanics first, cards as proof.

irobinson010:
Fixed on #999 (base `develop`, CI green). The scope here was "caveats now, a real fix later" — the real fix turned out to be cheap enough to do instead, so the caveats are **deleted** rather than written.

[...]

Issue #690, opened 2026-09-16
Toughness SBA still skips coded */* CDA creatures at computed toughness 0 (Consuming Aberration, Lord of Extinction)


krakenhavoc:
## Bug

**A coded `*` creature whose power/toughness come from a characteristic-defining ability (CDA) is never killed by the toughness state-based action (CR 704.5f) when its computed toughness is 0 and it has no counters.**

The SBA skips `Toughness == 0 && len(Counters) == 0` creatures as placeholders (`server/internal/game/mutations.go` ~:2270-2277 on `develop`). #683 (PR pending, branch `fix/zero-toughness-683`) narrows that skip with `LostLastCounter` and `VariableToughness`, but it deliberately **keeps** the skip for `*` creatures (`VariableToughness`). That's right for uncoded `*` cards whose stats the engine can't compute. It's **wrong for coded CDA cards**, whose toughness the layer system does compute:

- **Lhurgoyf** (`*/1+*`) can never reach 0, but `*/*` CDA creatures can.
- **Consuming Aberration**, with empty opponent graveyards, is a 0/0 and should die.
- **Lord of Extinction** (`*/*`, cards in all graveyards) is a 0/0 with empty graveyards and should die.

Found during the #683 split review (2026-09-16).

## Fix direction

- [ ] Let the skip apply only to `*` creatures **whose stats the engine doesn't compute**. For example, skip when `VariableToughness && !hasCharacteristicDefiningAbility(c)`, where the CDA check is whatever marks a registered layer-7a P/T definition for that card (check how CDA statics are registered in S16 layers).
- [ ] Make sure the check reads the **effective** toughness after layers, not printed.
- [ ] Tests: a coded CDA creature at computed toughness 0 dies; the same card at toughness ≥ 1 lives; an uncoded `*` placeholder still survives.
- [ ] Sweep the catalog for coded `*` creatures and confirm none relied on the skip.

Depends on / stacks with: #683. Related: #259, #35/#620 (any new bool field goes in the `game.Card` bool block).

Sprint: S27 follow-up

irobinson010:
Claiming this together with #691 on branch `fix/690-691-cda-toughness-sba-and-x-zero` (base `develop`). Engine mechanics first, cards as proof.

irobinson010:
Fixed on #999 (base `develop`, CI green). Not a card fix and not a new bool on `game.Card` — the skip became one predicate.

[...]

Issue #683, opened 2026-09-16
Toughness SBA skips printed 0/0 creatures that lost their last counter (CR 704.5f) — LostLastCounter + VariableToughness


krakenhavoc:
## Bug

**A printed 0/0 creature that loses its last counter survives the toughness state-based action (CR 704.5f).**

`server/internal/game/mutations.go` ~:2270-2277 on `develop`:

```go
// Printed-0 creatures (Toughness == 0 and no counters applied)
// are skipped: that's the placeholder / unparseable-stats
// convention documented on Card.Power — the SBA would else
// destroy every demo seed card.
if c.Toughness == 0 && len(c.Counters) == 0 {
```

The skip exists for **placeholder or unparseable stats** (`*`, `1+*`, `?`, demo seed cards). But it can't tell those apart from a **real printed 0/0** such as Hangarback Walker, Mikaeus the Lunarch, Primordial / Kalonian / Goldvein / Lifeblood / Mossborn Hydra, Hydroid Krasis or Walking Ballista. When such a creature has **used up its +1/+1 counters**, `applyCounterLocked` sets `Counters = nil`, the skip applies, and a 0/0 stays on the battlefield. That is stronger than printed (#259).

Today this is only reachable through an X=0 cast, or through counter removal from effects. #625's counter-removal **cost** makes it ordinary play: Mikaeus paying his team pump with his last counter, or Fain, the Broker removing the last counter from any creature you control.

## Fix already written (on the #625 branch, to be split out)

Found by the adversarial review of #625, and implemented on local branch `feat/counter-removal-cost-625`. Commits `ed049bf` (engine part), `d8ba4c9` and `31cad03`:

- **`Card.LostLastCounter`:** set by `applyCounterLocked` only when a card's counters go from **some to none**. It is not set when "removing" a counter the card never had (`d8ba4c9`). The SBA skip no longer applies to such a card.
- **`Card.VariableToughness`:** set by the deck importer (`deck.go` `nonNumericStat`) when printed toughness is present but non-numeric (`*`, `1+*`, `?`); also per `Face`, carried by `SetFace`, copy effects (`CopiableValuesOf`, `applyCopy`, `restorePrintedSelf`) and snapshots. A `*` creature keeps the placeholder skip even after losing its last counter, as on `develop` (`31cad03`).
- The new SBA condition skips a creature only when `Toughness == 0 && len(Counters) == 0 && (!LostLastCounter || VariableToughness)`.
[...]

Issue #625, opened 2026-09-16
Heart of Kiran: alternative crew cost as a second ability with a real 'remove a loyalty counter' cost


krakenhavoc:
## Why

Heart of Kiran ships with Crew 3 only (`server/internal/cards/effects/heart_of_kiran.go`, `CompletenessCaveats`). The printed alternative is missing: "You may remove a loyalty counter from a planeswalker you control rather than pay Heart of Kiran's crew cost."

**Decision (2026-09-16): model it as a second activated-ability entry**, not as an alternatives slot on `AbilityCost`. The client already lists abilities separately, so this shows as two menu entries ("Crew 3" / "Crew: remove a loyalty counter"), with no new "choose how to pay" UI.

## The constraint the card file already names

`heart_of_kiran.go:14-17` rejects a *fake* second entry: one that could be activated with no planeswalker in play would be **stronger than printed**, which #259 forbids. So the second entry needs a **real cost component**. That means adding to `AbilityCost` even though there's no alternatives slot:

- [ ] New cost component, e.g. `AbilityCost.RemoveCounterFrom{Kind: "loyalty", Filter: planeswalker you control, N: 1}`:
  - **Legality:** activatable only while a planeswalker you control has at least 1 loyalty counter
  - **Payment:** choose which planeswalker if there are several (a prompt, or auto when there's exactly one), then remove the counter. Removing a planeswalker's last counter sends it to the graveyard through the existing SBA (CR 704.5i).
  - **Not loyalty activation:** it doesn't count as activating a loyalty ability (no `LoyaltyActivatedThisTurn` stamp, CR 606)
- [ ] Heart of Kiran registers the second `ActivatedAbility` with that cost and the same `CrewEffect("Heart of Kiran")`; it drops the caveat and declares `complete` if nothing else is missing.
- [ ] Update the comment in `heart_of_kiran.go`, and the AGENTS.md §7 "alternative cost on an activated ability" gap (#621 added it): record that "rather than pay" on an activated ability is modelled as a separate ability entry with its own real cost.
- [ ] Bots: the enumerator offers the second ability only when it's payable.
- [ ] Tests: unpayable with no planeswalker or no loyalty; payable and crews; picks among several planeswalkers; a planeswalker at 1 loyalty dies to the SBA; doesn't set the once-per-turn loyalty gate.

Size: S–M. Related: #92 (S27), #621, #259.

Sprint: S27 follow-up

[...]

PR #999, opened 2026-09-19
fix(engine): CR 704.5f applies to every creature whose toughness the engine knows (#690, #691)

irobinson010:
Closes #690
Closes #691

The toughness state-based action skipped every creature with printed
`Toughness == 0` and no counters. That shape is the importer's stand-in
— `strconv.Atoi("*")` fails and writes 0 — and the skip was the only
thing stopping CR 704.5f from eating every `*` creature the engine
cannot size. It was covering two things it never should have, and it
skipped the **whole** creature, so the lethal-damage (CR 704.5g) and
deathtouch (CR 702.2c) checks below it never ran either: a Consuming
Aberration was not merely un-killable at toughness 0, it was
un-killable at toughness 9 with 9 damage on it.

## The shapes

**One predicate, `game.Card.ToughnessIsKnown`** (`server/internal/game/card.go:844`),
read at exactly one place, `server/internal/game/mutations.go:2845`.
The question is "does the engine know what this object's toughness
IS", because that — not "is the number 0" — is what the convention was
ever about. Decided in order:

1. A nonzero `Toughness` or any counter. The fast path, unchanged.
2. `Characteristic.PTDefined` — **#690**.
3. Otherwise `VariableToughness` keeps the skip — #683, unchanged.
4. `LostLastCounter` — #683, unchanged.
5. A printing behind the object (`ScryfallID`) — **#691**.

**`Characteristic.PTDefined`** (`server/internal/game/characteristic.go:160`),
set by `definesPowerAndToughness` in the layer pass
(`server/internal/game/layers.go:458` and `:518`) for the **7a and 7b
buckets only**. 7c modifies, 7d counts counters, 7e switches — all
three need a number something else defined. A layer 7a CDA is the
engine answering the very question the stand-in stands in for, so a
Lord of Extinction this engine sizes has a real toughness, empty
graveyards included. Deliberately **not** compared by
`sameCharacteristic` (`server/internal/game/layer_dependency.go`,
documented in place): CR 613.8a asks what an effect does to an object,
not what the pass did, and `applyRaw` never writes it. Rebuilt with
the rest of the `Characteristic` every recompute, so no snapshot field
and no drift-plan entry — `effective` is already classified `rebuilt`.

**Branch 5 is the #691 rule.** Branch 3 already took every printing
whose toughness did not parse, so what branch 5 sees is a real printed
0/0: a Hangarback Walker cast for X=0, a Wildwood Scourge that entered
with no counters, a declined Clone. **The engine does not refuse the
X=0 cast** — CR 601.2b makes X a number the caster announces and 0 is
legal. It simply does not survive one, which is what paper does. The
bot enumerator already declines to *offer* X=0 for a card declaring
`XMatters` (`server/internal/legal/x.go:49`), a separate rule about
what is worth putting in front of a player; that is untouched.

**What stays skipped** is what the convention was always for: objects
with no printing, no computed body and no counter history — a test
fixture that typed a creature line without a body, a 0/0 token
template. Measured: 1 test in the whole suite changed behaviour on
branch 5 (below), against 146 for the version of this change that made
every 0/0 die.

**Importer safety net** (`server/internal/deck/deck.go:529`,
`variableToughness`): a MISSING printed toughness on a creature is now
stamped `VariableToughness` rather than read as a printed 0. Every
creature prints one, so a Scryfall record the dump reader cannot read
must not become a death sentence under branch 5 — it keeps the skip
and lives as a visible, reportable 0/0 instead. A joined multi-face
type line is exempt: Scryfall leaves that toughness empty on purpose
and `SetFace(0)` overwrites the flag from the face. This is what the
one behaviour-changed test found — `TestVigilantAttackerDoesNotTap`'s
`prepare`-layout fixture has a creature face with no P/T strings.

## Cards

Two player-facing caveats declared the engine **stronger** than the
printed card and are now false, so they are gone: **Mikaeus, the
Lunarch** and **Marketback Walker** ("Cast for X=0, it stays on the
battlefield as a 0/0 instead of dying"). Two more were the same gap
under another name — **Clone** and **Stunt Double** ("if you decline
the copy, the 0/0 stays on the battlefield instead of dying") — and
both drop to `CompletenessFull`. Census regenerated with
`go test ./internal/cards/coverage/ -update`: full 1563 → 1565,
caveats 408 → 406 (regenerated, never hand-merged, against the
post-#991 develop).

**Wildwood Scourge's Go comment** — the one #691 names as false — said
a Scourge cast for X=0 "dies to the toughness check, as in paper" and
it did not. It is true now, and the comment says what made it true
rather than being quietly deleted. **Lord of Extinction's** comment
had the same shape and is the #690 half.

Comments corrected on the rest of the family, each of which declared
the X=0 survival as an engine gap: Hangarback Walker, Goldvein Hydra,
Primordial Hydra, Lifeblood Hydra, Hydroid Krasis, Shivan Devastator,
Multani (the no-land case, reached through a 7c modify rather than an
entry counter, so it is branch 5 and not branch 2).

## Tests

`server/internal/game/toughness_known_test.go` (new) — a coded CDA at
computed toughness 0 dies, at 3 lives, and now dies to lethal damage;
an uncoded `*` still survives; `ToughnessIsKnown`'s eight-row order
table; a printed 0/0 with a printing dies and without one is still
skipped; a printed 0/0 under an anthem lives at toughness 1, which the
old skip could not even reach because it fired before the toughness
was computed; a layer-7b set to 0/0 kills a token.

`server/internal/cards/effects/printed_zero_body_test.go` (new) — Lord
of Extinction dies with every graveyard empty and lives at one card;
Consuming Aberration dies against empty opponent graveyards; a 2/2
Lord dies to two damage; Hangarback Walker dies at X=0 making no
Thopters and lives at X=1; Wildwood Scourge, Mikaeus die at X=0;
Multani dies with no lands and lives with one; a declined Clone dies;
a 0/0 fixture with no printing is still skipped.

`server/internal/game/toughness_known_test.go`, added on the rebase for
#967's copiable values — both roads turn out to be right for the same
reason, that the body and the bit saying how to read it travel
together, so these pin rather than change:

- a **Clone of a coded `*`/`*`** answers branch 2, not branch 3:
  `CopiableValuesOf` carries the stand-in 0 *and* `VariableToughness`,
  but it carries the oracle ID too, so the layer pass gathers the same
  CDA for the copy and stamps `PTDefined` per object. The copy dies
  with the original at a computed 0 and lives with it at 3.
- a **CR 608.3f token copy** answers branch 5 with the *copied* card's
  printing, which is correct — the token's body came out of that same
  printing. A token copy of a printed 0/0 dies with no counters; one of
  an uncoded `*` keeps `VariableToughness` and stays skipped. The case
  worth guarding is a token taking the `ScryfallID` **without** the
  body it belongs to, and `tokenCopyOfSpell` → `CopiableValuesOf` →
  `setPrintedValues` cannot produce one; two rows assert that pairing
  and a third runs the token through the real state check.

Changed: `TestB13GoldveinHydraWithXZeroIsAZeroThatPaysNothing` pinned
the old outcome and now pins the printed one. The Mikaeus caveat
assertion in `counter_cost_cards_test.go` flipped from "the X=0 gap
must be declared" to "it must not come back".
`TestNonNumericToughnessIsStampedVariable` gains the missing-toughness
row and a non-creature row.
`TestPrintedZeroSkipStillCoversCreaturesThatNeverLostCounters` keeps
passing and gains a note saying why (its fixture has no printing).

## Back-outs run

- Drop branch 2 (`PTDefined`) → fails `TestCodedCDACreatureAtZeroToughnessDies`,
  `TestCodedCDACreatureDiesToLethalDamage`, `TestLayerSevenBSetDefinesTheBodyToo`,
  `TestLordOfExtinctionWithEmptyGraveyardsDies`,
  `TestConsumingAberrationWithEmptyOpponentGraveyardsDies`,
  `TestCodedStarCreatureDiesToLethalDamage`. Restored.
- Drop branch 5 (`ScryfallID`) → fails `TestToughnessIsKnownOrdersItsAnswers`,
  `TestPrintedZeroZeroWithAPrintingDies`, `TestB13GoldveinHydraWithXZeroIsAZeroThatPaysNothing`,
  `TestHangarbackWalkerCastForXZeroDiesAtOnce`, `TestWildwoodScourgeCastForXZeroDiesAtOnce`,
  `TestMikaeusCastForXZeroDiesAtOnce`, `TestMultaniWithNoLandsDies`,
  `TestDeclinedCloneDiesAsItsPrintedZeroZero`. Restored.
- Drop the importer safety net → fails `TestNonNumericToughnessIsStampedVariable`
  and `TestVigilantAttackerDoesNotTap`. Restored.
- Re-run after the rebase: dropping branch 2 also fails
  `TestCopyOfACodedStarCarriesPTDefined`; dropping branch 5 also fails
  `TestTokenCopyOfAPermanentSpellReadsTheCopiedBody` and
  `TestTokenCopyOfAPrintedZeroZeroDiesOnArrival`. Restored.

## Docs

ADR 0007 §7 amended (the exemption becomes the predicate, with the
X=0 answer and the damage-check bug spelled out; the "placeholder
card" framing is retired — the engine has objects it has printed data
for and objects it does not). ADR 0012 §3 amended with `PTDefined`.
AGENTS.md gains the rule under the layer/sub-layer table: **a CDA you
code must compute the number the card prints, 0 included** — do not
floor it at 1 to keep the creature alive. `docs/engine-seams.md`'s
#625 row points at the whole rule rather than half of it.

No seam opens or closes and no new ADR number was taken, so no race
with the concurrent branches.

## Out of scope

- **`server/internal/cards/effects/hangarback_walker.go:52` and 12
  siblings — the X entry counters still go on at RESOLUTION, not
  through the CR 614.1c entry pipeline.** Each declares the caveat
  "effects that watch you put counters on a permanent don't see them",
  and each says an entry replacement "cannot see the X announced for
  the spell". That last claim is **false today**:
  `server/internal/game/mutations.go:1946` hands the resolving
  `*StackItem` to `applyAltCostEntryCountersLocked`, right where the
  `ReplacementEvent` is built, so X is readable there. Moving the
  family onto `ReplacementEvent.EntersWithCounters` would delete 13
  caveats and put the counters inside the entry event where CR 616
  can see them. NOT done here: it needs a new `Spec` field and a new
  `Catalog*` hook, which collide directly with the open #967 and #988
  on `effects/spec.go`, `effects/carddef.go`, `game/carddef.go` and
  `game/snapshot.go`. Worth its own issue after both land; it does
  not change the X=0 outcome either way, since zero counters are zero
  counters whichever road they take.
- **`server/internal/game/alternative_cost.go:620` and
  `:227`** say escape's counters "ride the same `EntersWithCounters`
  map Hangarback Walker's own self-replacement writes". Hangarback
  writes no such thing — it is the `OnResolve` above. Left alone to
  avoid touching #988's file; it is the same follow-up.
- **A 0/0 token with no CDA and no counters still survives.** Branch 2
  covers the tokens that matter (Hallowed Haunting's Spirits,
  Simulacrum Synthesizer's Constructs are sized by a layer effect),
  branch 4 covers the ones that ever held a counter, and a token
  template with a genuinely printed 0/0 body and nothing sizing it
  does not exist in the catalog today. Pre-existing; noted rather
  than guessed at.
- **`server/internal/cards/effects/hydroid_krasis.go`'s comment
  claimed X=1 also left the Krasis counterless.** Never true — its
  `OnResolve` puts `ctx.X()` counters on and only the life and the
  draw are halved. Corrected here, since the line was being rewritten
  anyway.

## Gate

From `server/`: `go build ./...`, `go vet ./...`, `go test ./...`,
`AISEAT_GAME_TESTS=1 go test ./internal/... -p 2`, `make lint`
(golangci-lint v1.64.8), `make fmt-check` — all clean. `client/` is
untouched.

**Rebased once** onto `f9a26a3a`, covering #965, #967, #1001, #989,
#995 and #991. Only the generated census conflicted, and it was
regenerated rather than merged; `card.go` (`PrintedValues` /
`CopiableValuesOf`) and `mutations.go` (`spellMovedItselfLocked`,
`beginResolvingSpellLocked`, the permanent-copy token branch) merged
cleanly into different hunks and both sides were verified present
afterwards, as were the AGENTS.md and `docs/engine-seams.md` edits.
#991 did fold `Card.PaidOptionalCosts` into `Card.Provenance`, and it
touches neither branch of this rule: `ToughnessIsKnown` reads
`ScryfallID`, `Toughness`, `Counters`, `VariableToughness`,
`LostLastCounter` and the effective characteristic, and no cast-time
state. ADR numbers re-swept across all 311 remote heads: highest is
0074, and this branch claims none — it only amends 0007 and 0012.


irobinson010:
Follow-up filed as #1002 for the out-of-scope CR 614.1c entry-counter move (13 cards, 13 identical caveats), to be picked up once #967 and #988 have landed.

Issue #1005, opened 2026-09-19
A `carried` snapshot classification is a promise nothing enforces


irobinson010:
Noticed while landing #980, running the back-out for the `Card.ChosenPlayer` snapshot carry. The back-out did not bite, and the reason generalises.

**The claim.** `snapshot_test.go`'s header says the round-trip property means "a field added to the snapshot is covered the moment it is added, with no test edit", and that the blind spot — "a field that exists on `*Game` but on neither side of the snapshot" — is closed by `snapshot_drift_test.go`.

**What actually happens.** Neither test covers a `carried` field whose projection is missing.

- `TestSnapshotRoundTripIsExact` (`server/internal/game/snapshot_test.go:306`) compares `capture → JSON → restore → capture`. Delete a field from BOTH the `Card → CardSnapshot` projection (`snapshot.go:962`) and the `CardSnapshot → Card` one (`snapshot.go:1484`) and the test still passes: absent equals absent. It only catches a field dropped in ONE direction.
- `snapshot_drift_test.go` walks the field NAMES on `game.Card` and fails on one with no plan entry. It never checks that a field classified `carried` is actually copied by either projection. `"X", carried, ""` is a promise nothing enforces.
- The round-trip's fixture is `enrich` (`snapshot_test.go:69`), and a field it does not set is invisible to the property even in the one-directional case.

I verified this by deleting `ChosenPlayer` from both projections: every snapshot test passed.

**Scale.** Unknown, and that is the point. Three `carried` fields were provably uncovered when I looked — `NamedTribe`, `ChosenColor` and `ChosenPlayer`, the CR 614.12-family stored answers, none of which `enrich` was setting. #980 added all three to the fixture and added `TestTheStoredAnswersSurviveASnapshot`, which asserts on the RESTORED GAME rather than on a re-captured snapshot. That closes those three and nothing else.

**Why it matters.** These are exactly the fields a restore cannot re-derive — a player made the choice. A dropped one restores a Cavern of Souls naming no tribe or a True-Name Nemesis protected from nobody, silently, with nothing in the state to say what went. That is the failure mode the drift test's own doc comment says it exists to prevent.

**Fix shape.** Two candidates, and the first is probably enough:

[...]

Issue #993, opened 2026-09-19
The exile-payout lint should widen to the other exits, and six sites want triage first


irobinson010:
Found while fixing #911 (branch `fix/911-489-cancelled-exile-payouts-and-self-exile`, PR forthcoming).

#911 shipped `server/internal/cards/effects/exile_payout_guard_test.go`: a source scan that fails on a **condition, after a fire-and-forget exile, that reads a local the function assigned before the exile**. That shape is a clause about the card, and a clause about the card has to know whether the card actually went (CR 400.7, [ADR 0013 §5t](https://github.com/krakenhavoc/cmd_and_ctrl/blob/develop/docs/decisions/0013-replacement-effects.md)).

The lint is deliberately scoped to **exile** — that is the verb #911 is about, and the rule it enforces was settled for exile and only for exile. Every other exit is pausable and cancellable in exactly the same way (`destroyRoute`, `bounceRoute`, `tuckRoute`, `millRoute`, `sacrificeRoute`), and widening the scanner's `exileStartsTheClock` / `exilePrimitives` tables is a two-line change. Running it widened today finds **six** sites, each needing its own rules answer before the lint can be turned on:

| site | shape | first read |
|---|---|---|
| `server/internal/cards/effects/boomerang_basics.go:43` | `youControlled` read before `BounceToHand`, gates the draw | the closest analogue of #911. "Return target nonland permanent to its owner's hand. If you controlled that permanent, draw a card" — is the draw gated on the RETURN, or only on who controlled it? At minimum the ordering is wrong: a bounced commander's CR 903.9 prompt is open while the draw happens |
| `server/internal/cards/effects/batch14_helpers.go:336` | `top.IsLand()` after `MillCards{N: 1}`, inside a mill-until-a-land loop | the loop proceeds AROUND a paused leg and re-reads the top of the library. #893's `MillToZoneThenForEffect` is the shape |
| `server/internal/cards/effects/batch21_helpers.go:262` | `land` read before `MillCards`, used after | same family |
| `server/internal/cards/effects/dark_confidant.go:91` | `life` read before `BounceToHand`, gates a `return nil` | probably fine — the life loss is unconditional and the guard is "it cost nothing"; wants the allowlist entry with that reason rather than a change |
[...]

Issue #1019, opened 2026-09-19
A sacrifice PROMPT has no continuation, so "if you sacrificed a creature this way" pays out before the answer


irobinson010:
Found while widening the exit-payout lint in #993 (branch `chore/993-1005-exit-lint-and-carried-enforcement`).

`g.PlayerSacrificesForEffect` and `g.EachPlayerSacrificesForEffect` (`server/internal/game/sacrifice.go:164`, `:131`) do not sacrifice anything. They QUEUE a prompt per seat and return **how many seats were asked**. The permanent leaves the battlefield when the player answers, which is one or more actions later.

That makes every clause written after one of them a payout on a move that has not happened — the #911 / #993 shape, arriving through the prompt rather than through a fire-and-forget exit. The widened lint in #993 deliberately leaves both entry points out of its tables, because the message a lint prints has to name a fix and there is no continuation form to name: the choose-cards prompt behind `SacrificeChoice` has a `Then`, but these two queue their own prompt and hand the caller nothing.

**The site that shows it.** `server/internal/cards/effects/rise_of_the_witch_king.go:49`:

```go
ctx.Game.EachPlayerSacrificesForEffect(ctx.Source(), controller, spec, "…")
if ctx.Game.PlayerSacrificesForEffect(ctx.Source(), controller, spec, "…") == 0 {
    return nil
}
// … return a permanent card from your graveyard
```

"Each player sacrifices a creature of their choice. If you sacrificed a creature this way, you may return another permanent card from your graveyard to the battlefield." The gate is "was the controller handed a prompt", the card's own comment says so, and the card declares the consequence as a caveat: *"the returned permanent is on the battlefield before you answer your own sacrifice prompt"*. It is the right approximation for the machinery that exists; it is not the printed card.

**Scale.** `git grep -l 'PlayerSacrificesForEffect\|EachPlayerSacrificesForEffect' server/internal/cards/effects/*.go` lists 14 files today. Most are "each player sacrifices" with nothing hanging off the answer, which is fine. The ones with a clause after the prompt are what this issue is about; Rise of the Witch-king is the one the lint surfaced.

[...]

Issue #980, opened 2026-09-19
Protection from a player (CR 702.16k): True-Name Nemesis needs Card.ChosenPlayer and a player quality in the protection grammar


irobinson010:
Left open by design in ADR 0072 (PR #979). The protection grammar (`game/protection.go`) covers colours, card types, creature subtypes and `everything`; it deliberately refuses `protection from the chosen player` so nothing half-works. True-Name Nemesis is caveated to this and to #929's choose-a-player prompt.

**Fix shape (ADR 0072 designs it):** `Card.ChosenPlayer` set by an as-enters CR 614.12 choice through #929's `QueueChoosePlayerForEffect` (an entry choice like `ChooseColorAsEnters`), a stable seat-referenced token spelling (never a raw UUID in a display string), and the player quality read by the one `ProtectedFrom` predicate — DEBT against the source's *controller* rather than its characteristics. Sites: `game/builtin_replacements.go:213` and `game/keywords.go:156` name the gap. Classify `ChosenPlayer` `carried`; a client badge tooltip that names the seat.

irobinson010:
Claiming this together with #994 on branch `feat/980-994-protection-from-a-player-and-seat-options` (base `develop`). Engine mechanics first, cards as proof.

irobinson010:
Landed in #1007 (CI green, `8863c999`), together with #994 — the two turned out to be one piece of work, and building this prompt on that one is what stops there being two answers to "which seats may this prompt offer".

**ADR 0072 §7's design held.** Every bullet is as written: `Card.ChosenPlayer` on `Card.ChosenColor`'s pattern, classified `carried`, the token `protection from the chosen player`, the matcher against the source's controller. §7 has a dated amendment recording the four things the build had to decide that the design did not say, and §9's `ChosenPlayer` bullet moves from "will be" to "is".

**The field.** `Card.ChosenPlayer`, per instance, carried by clone and the snapshot, cleared at **both** CR 400.7 sites — `zone.go`'s battlefield exit and `resetAsNewObjectLocked`'s new-object reset, because a blinked permanent reaches the second without passing the first. Not a copiable value (CR 707.2), which needs no rule: `CopiableValuesOf` projects printed characteristics and never looks at it, exactly as ADR 0071 arranged for `ClassLevel`. There is a test saying so.

[...]

PR #1020, opened 2026-09-19
fix(cards): the payout lint covers every exit, and `carried` is a property the snapshot must satisfy (#993, #1005)

irobinson010:
## Summary

- **#993** — the exile-payout lint widens to every pausable exit (destroy, sacrifice, bounce, tuck, mill, a discard, a graveyard arrival). All six sites the issue listed are triaged: one changed an outcome, three were out of order, one was spinning, one stopped being a false positive when the primitive behind it grew a real continuation. Two allowlist entries remain, both carrying their reason.
- **#1005** — `carried` stops being a promise. `snapshot_carried_test.go` writes a distinctive value into every field the drift plan classifies `carried`, runs the real capture → JSON → restore, and reads it back **off the restored game**. 213 fields, reflection over the plan, so a `carried` row added tomorrow is enforced with no test edit.
- Two new continuation primitives, both wrappers over batches that already exist: `game.BounceToHandThenForEffect`, plus `BounceToHand.Then` and `SacrificePermanent.Then` in the catalog.

## Sprint

S49 — operability and the hygiene tail (#892). The lint half also answers S39 #882.

## Issues

Closes #993
Closes #1005

Filed on the way: #1019 (out of scope, below).

---

### The lint

`server/internal/cards/effects/exile_payout_guard_test.go` → `exit_payout_guard_test.go`. The AST shape it flags is unchanged — a **condition, after a fire-and-forget exit, reading a local the function assigned before it** — and the two tables now carry a **verb** rather than a bool:

| verb | entry points | the fix the message names |
|---|---|---|
| exile | `ExileCardForEffect`, `ExileCardsForEffect`, the with-permission and face-down forms, `ExileTarget`, `ExileAllMatching`, … | `ExileTarget.Then` / `ExileThenIfItWas` / `g.ExileCardThenForEffect` |
| destroy | `DestroyPermanentForEffect`, `DestroyPermanentsForEffect`, `DestroyTarget`, `DestroyAllMatching` | `DestroyAllMatching.Then` / `g.DestroyPermanentsThenForEffect` |
| sacrifice | `SacrificePermanentForEffect`, `SacrificeAllForEffect`, `SacrificePermanent` | `SacrificePermanent.Then` / `g.SacrificeThenForEffect` |
| bounce | `BounceToHandForEffect`, `BounceCardsToHandForEffect`, `BounceToHand`, `BounceAllMatching`, `ReturnAllToHand` | `BounceToHand.Then` / `g.BounceToHandThenForEffect` |
| tuck | `TuckToLibraryForEffect`, `TuckToLibraryAtDepthForEffect` | `g.TuckToLibraryThenForEffect` |
| mill | `MillNForEffect`, `MillToZoneForEffect`, `MillCards`, `MillToZone` | `MillToZone.Then` / `g.MillToZoneThenForEffect` |
| discard | `DiscardRandomForEffect`, `DiscardCards` | none yet — the message says where to write it |
| graveyard arrival | `PutIntoGraveyardForEffect` | none yet — the message says where to write it, and what the body is |

A finding prints **the site, the exit kind and the continuation form** (rule: a lint other agents will trip has to say what to do). Example:

```
boomerang_basics.go:43 gates on `youControlled`, read before the bounce at
boomerang_basics.go:40 — use BounceToHand.Then, ReturnAllToHand.Then or
g.BounceToHandThenForEffect / g.BounceCardsToHandThenForEffect, and gate the
clause on `bounced`
```

### The rules calls (ADR 0013 §5v)

§5t's test — "a clause GATED on a pre-exile fact is about the card" — is right for the family it was written about and is not the test. §5v restates it as **what the clause is ABOUT**:

- **about the object that moved** (`if it WAS a creature card`, `for each card exiled this way`, `if you do`) → gated. A move that did not happen leaves the clause with no referent, CR 614.10.
- **about a player, or about the spell's TARGET** (`its controller gains life equal to its power`, `if you controlled that permanent`, `then that permanent's controller may …`) → **not** gated, but it still has to WAIT: the order is observable when a commander is holding a CR 903.9 prompt open.

| site (file:line, pre-change) | call | change |
|---|---|---|
| `ruthless_technomancer.go:78` | sacrifice | **outcome fixed.** "If you do, create Treasures equal to that creature's power" was a live-board read on the next line — `sacrificedThisWayLocked`'s rule by hand, and right for everything but a PAUSED leg. A sacrificed commander is still on the battlefield while its owner answers, so the read said "not sacrificed" and no Treasure was ever made. Now `SacrificePermanent.Then`; both answers pay (CR 701.17a), a cancelled sacrifice pays nothing. |
| `boomerang_basics.go:43` | bounce | **order only.** The draw moves into `BounceToHand.Then` with the answer ignored. "If you controlled that permanent" is about the player's past control of the TARGET: true before anything moves, and no replacement rewrites it — a commander of yours that took the command zone was still yours. Path to Exile's shape (§5m item 2). |
| `chain_of_vapor.go:70` | bounce | **order only** — a seventh site the issue did not list, same family. "Then that permanent's controller may sacrifice a land" was being asked while that player's own CR 903.9 question was open. |
| `batch21_helpers.go:262` (Hermit Druid) | mill | **order only.** "Put that card into your hand and all other cards revealed this way into your graveyard" is one instruction about a settled run; the land was reaching hand with cards above it still in the library. |
| `batch14_helpers.go:336` (Consuming Aberration) | mill | **it was spinning.** The loop read the top of the library, milled one, and asked whether the card it had READ was a land. A paused leg leaves the card where it was, so the next pass read the same card and milled it again, to the 1000-iteration fuse. Now `MillToZone`'s own `Until` run, chosen up front against the pre-move copies (§5l). |
| `resolution_choice.go:242` | sacrifice | **the false positive stopped being one.** `then` is the continuation parameter — but the prompt's own body called the fire-and-forget sacrifice and ran the clause on the next line, so `SacrificeChoice.Then` did not mean "once the permanent has gone" the way its doc says. It does now, and the lint has nothing to allowlist. |
| `dark_confidant.go:91` | bounce | **allowlisted with the reason.** `if life == 0` is Solitude's guard with a different verb: an unconditional clause about a PLAYER (§5m item 5), with an early return so the engine is not asked for a life change of nothing. |
| `solitude.go:74` | exile | allowlist entry unchanged (§5t's). |

### The engine

`game.BounceToHandThenForEffect` is the **fourth** single-card wrapper over the batch, beside `ExileCardThenForEffect` (#870), `SacrificeThenForEffect` (#910) and `TuckToLibraryThenForEffect` (#783) — six lines over `BounceCardsToHandThenForEffect`, no second bounce path, because a second path with its own notion of what landed is how exile and destroy drifted apart (#815, #866). `BounceToHand.Then` and `SacrificePermanent.Then` are the catalog side, shaped exactly like `ExileTarget.Then`; the sacrifice passes `uuid.Nil` as the source because that is what the fire-and-forget form already stamps on `EventSacrifice`, so adding a `Then` changes sequencing and nothing else.

`resumeClause` (`helpers.go`) is the two lines every asynchronous clause ends with, named because the clone gate found the body in `ChoosePlayer` and `SacrificeChoice` and the third copy would have been written the same way.

### What keeps the widening from rotting

- **one synthetic offender per verb** in the self-test, each paired with the continuation the message names. A table rots one row at a time, and an entry point renamed in `game/` leaves its verb silently unguarded until a fixture fires through it.
- `TestExitPayoutAllowlistIsLive` — an entry with no reason fails, and so does an entry that no longer names a finding.
- `TestEveryExitVerbIsReachableFromTheTables` — a verb no table names, or a verb that names no fix, fails.
- `TestEveryExitEntryPointStillExists` — the tables are STRINGS, so a rename in `game/` or in this package leaves a row matching nothing while every other test keeps passing. Each of the 17 engine rows is checked against `(*game.Game)`'s method set by reflection, and each of the 14 catalog rows against this package's type declarations. Added while rebasing over three engine PRs at once, which is exactly when a table of names goes stale.

#1003's `TestEveryReplacementEventKindIsSwitchedOn` is the standard being copied.

---

## #1005 — the shapes

`snapshot_drift_test.go` walks field NAMES and never reads a value, and `TestSnapshotRoundTripIsExact` compares `capture → JSON → restore → capture`, which is structurally blind to a field missing from **both** projections: absent equals absent. Delete a field from `snapshotCard` **and** `restoreCard` and every snapshot test passed.

`server/internal/game/snapshot_carried_test.go`:

- `TestEveryCarriedFieldSurvivesTheSnapshot` — for each `carried` row, set a distinctive value on a fresh `newRestorableGame` + `enrich` fixture, capture → JSON → restore, read back off the **restored game**. **213 subtests** across `Game`, `Card`, `Player`, `Zone`, `StackItem`, `DelayedTrigger`, `PendingChoice`.
- the value generator is **plan-aware**: a domain type with a plan of its own is filled only in its `carried` fields, because filling a `rebuilt` one (`Card.ManaAbilities` is two closures the catalog re-supplies) would make the test assert the opposite of what that row says.
- **unexported fields are reached through `unsafe`** — half of `Game`'s carried rows are unexported (the RNG key, the event sequence, the combat lock-ins) and they are exactly the ones no other test can see.
- the failure **names the field and the projection**, by looking for the value in the captured JSON: in the file → `restoreX` dropped it; not in the file → `snapshotX` did.
- two escape hatches, both checked by `TestCarriedExemptionsAreLiveAndExplained`: `carriedFixture` (a type the generator cannot invent a plausible value for — a closed-set typed string; `Game.Seats`, `Game.StackMeta` and `Player.Emblems`, whose shape the restore re-derives) and `carriedNotRoundTrippable` (`Game.layerVersion`, advanced by one on purpose; `ScopedStatic.Duration`, which `Clone` carries and the snapshot does not).
- `TestEveryPlannedTypeHasACarriedProbe` — `driftPlans` is now **one list walked by three tests**, and a domain TYPE added to the plan with no probe fails CI. This is what makes the enforcement survive the rebase over #1008/#1011/#1014/#1016/#1017: a new `carried` row is enforced automatically, and a new domain type cannot arrive unenforced either.

**What it found:** no live hole beyond the three #980 had already closed by hand — which is the answer, and the reason the back-outs below matter more than the current pass. One hygiene bug: `cardFields` listed `ObjectEpoch` twice, so the `#936` comment was dead text a map had already thrown away; the two rows are merged.

Header comments in `snapshot_test.go` and `snapshot_drift_test.go` stop claiming coverage they do not have — the other half of the report.

---

## Rebased onto #1014 + #1016 + #1017

One rebase, onto `origin/develop` at `59926ecb`. The point of interest is that **the `carried` enforcement picked up both new rows on its own, with no edit to the test** — which is the property the test exists to have:

```
--- PASS: TestEveryCarriedFieldSurvivesTheSnapshot/StackItem.SourceEpoch
--- PASS: TestEveryCarriedFieldSurvivesTheSnapshot/PendingChoice.OwedInStep
```

213 enforced fields before the rebase, **222** after. The generator needed no new fixture for either: `SourceEpoch` is an `int` and `OwedInStep` a plain field the generator synthesises, so neither is in `carriedFixture`. #1016's removal of `PendingChoice.ForceBlocks` left nothing stale — my two exemption tables never named it, and `TestCarriedExemptionsAreLiveAndExplained` would have failed on the way through if they had. `NoChoiceIndex` is a constant (`option_pick.go:97`), not a struct field, so it is correctly absent from the plan and from the enforcement.

The widened lint was re-run across the whole merged tree: `TestNoExitPayoutWithoutAContinuation` and `TestExitPayoutAllowlistIsLive` both green, so no card file from the three merged PRs carries the pattern and both allowlist entries still name live findings at their current lines. Census regenerated with `-update`: no movement. ADR sweep re-run across all 319 remote heads: highest file is still `0074`, and the only head carrying a `§5v` is this branch.

## Test plan

- [x] `go build ./... && go vet ./... && go test ./...` from `server/` — clean.
- [x] `AISEAT_GAME_TESTS=1 go test ./internal/... -p 2` — clean.
- [x] `make -C server lint` (golangci-lint v1.64.8) — clean.
- [x] `make -C server fmt-check` — clean.
- [x] re-ran all of the above after `git rebase origin/develop` (onto fb24372f).
- [ ] `client/` untouched.
- [ ] census not regenerated: no card's `Completeness` or caveat changed. Ruthless Technomancer keeps both of its caveats (the announce-time target choice and the missing reanimation ability); the other four cards were and are `full`.

### Tests added

| file | what it pins |
|---|---|
| `server/internal/cards/effects/exit_payout_guard_test.go` | the widened lint, one synthetic offender + one answer **per verb**, the live-allowlist check, the verb-reachability check |
| `server/internal/cards/effects/exit_payout_cards_test.go` | all five migrated cards on a real board — a commander holding the CR 903.9 prompt open (both answers) and a CR 614 window that cancels the move outright |
| `server/internal/game/single_card_bounce_then_test.go` | `BounceToHandThenForEffect`: does not run while the prompt is open, runs exactly once on the answer, reports CR 400.7's arrived-object, and still runs (with `false`) for a cancelled move |
| `server/internal/game/snapshot_carried_test.go` | the `carried` enforcement, the two exemption lists, and the every-planned-type-has-a-probe guard |

### Back-outs run (revert → confirm the new test fails → restore)

| fix reverted | test that fails |
|---|---|
| Boomerang Basics → fire-and-forget bounce | `TestBoomerangBasicsWaitsForTheCommandZoneAnswer` |
| Chain of Vapor → fire-and-forget bounce | `TestChainOfVaporAsksTheChainAfterTheReturnSettles` |
| Ruthless Technomancer → live-board read | `TestRuthlessTechnomancerPaysForASacrificedCommander` |
| Hermit Druid → mill then bounce on the next line | `TestHermitDruidPutsTheLandInHandAfterTheMillSettles` |
| Consuming Aberration → the hand-rolled mill loop | `TestConsumingAberrationDoesNotReMillAPausedCommander` |
| `SacrificeChoice.Then` → fire-and-forget sacrifice | `TestChainOfVaporCopyQuestionWaitsForTheSacrificedLand` |
| the lint's tables narrowed to #911's two entries | `TestExitPayoutGuardCatchesTheShape` (7 per-verb offenders go quiet) and `TestEveryExitVerbIsReachableFromTheTables` |
| `NamedTribe` deleted from **both** projections | `…/Card.NamedTribe`, reporting "the CAPTURE projection dropped it" |
| `ChosenColor` deleted from `restoreCard` only | `…/Card.ChosenColor`, reporting "the RESTORE projection dropped it" |
| `restoreZone` made to use its fallback unconditionally | `…/Zone.Kind`, reporting "the RESTORE projection dropped it" |
| one engine row in `exitStartsTheClock` renamed | `TestEveryExitEntryPointStillExists`, naming the row and the verb it was guarding |
| one catalog row in `exitPrimitives` renamed | same, naming the missing type |

---

## Out of scope

- **`server/internal/game/sacrifice.go:164` and `:131`** — `PlayerSacrificesForEffect` / `EachPlayerSacrificesForEffect` queue a prompt and return how many seats were ASKED; nothing has left the battlefield when they return, and the prompt carries no continuation. `server/internal/cards/effects/rise_of_the_witch_king.go:49` reads that count as "if you sacrificed a creature this way" and returns the graveyard permanent before anybody has answered — which the card already declares as a caveat. Both entry points are deliberately **out of the lint's tables**, documented in the header, because a lint whose message names a fix that does not exist is worse than no lint. Filed as **#1019**.
- **`server/internal/cards/effects/primitives.go:135` (`DestroyTarget`)** — no `Then` field, because no catalog site needs one today; the lint's destroy message points at `g.DestroyPermanentsThenForEffect` with a set of one. Not added speculatively.
- **`server/internal/game/random_bottom.go:175` (`PutIntoGraveyardForEffect`)** and the discard entry points — in the lint's tables with zero findings today, and their messages say where to write the missing continuation (`routeAllThenLocked(zoneRoute{Dst: ZoneGraveyard}, …)`) rather than naming one that exists. No dead code shipped.

One thing found in my own test and fixed before it shipped: `restoreZone` (`server/internal/game/snapshot.go:1455`) falls back to the caller's expected kind for an empty one, and every caller's expectation IS the slot's real kind — so a `Zone.Kind` fixture of `ZoneGraveyard` on a graveyard would have passed forever even with the field dropped from both projections. The fixture is `ZoneExile` on a graveyard now, deliberately wrong for its slot, and the back-out above proves it bites. Every other fixture was checked for the same masking (the `VariableToughness`, `FaceDownKind` and `LandDropsPerTurn` backfills all produce a value the fixture differs from).

Fixed on this branch and worth calling out: the duplicated `ObjectEpoch` row in `cardFields` (`snapshot_drift_test.go`), the stale doc comment on `b14EachOpponentMillsUntilLand` ("milled one card at a time"), and `resumeClause` replacing the two copies the clone gate flagged.


irobinson010:
Note on the squash commit `8838d64e`: its subject reads "and  is a property the snapshot must satisfy" because the merge tooling dropped the word `carried` (an unescaped backtick in the subject string). The PR title and body are correct; the commit content is unaffected. Recording it here so the gap in the subject is not read as anything else.

Issue #830, opened 2026-09-17
Re-pointing a blocker leaves the "becomes blocked" trigger of the attacker it left (Cyberman Patrol afflict on an unblocked attacker)


krakenhavoc:
## What happens

The sandbox lets a defender re-point a blocker at a different attacker (the "Re-declare blocker" context-menu item). The first declaration has already emitted `EventBlock` naming the first attacker, so any "whenever this becomes blocked" trigger on that attacker is already harvested. Re-pointing doesn't withdraw it. The attacker ends the declaration **unblocked**, but its "becomes blocked" trigger still goes on the stack and resolves.

Example: Cyberman Patrol (afflict 3 for artifact creatures its controller controls) attacks. The defender declares a blocker on it, then re-points that blocker at another attacker. The Patrol is unblocked and deals combat damage. The defending player also loses 3 life from afflict, which the printed card doesn't do. Grazilaxx, Illithid Scholar's "you may return it" prompt appears for an attacker that is not blocked.

Found while drafting the ADR 0045 addendum (PR #822). The addendum's PR 2 fixes the related double `EventBlock` on a re-point. Its "Bugs found" list says this half "is **not** fixed here and needs its own issue".

## Evidence (origin/develop `deeec56e`)

- `server/internal/game/mutations.go:4401-4417`: `DeclareBlocker` sets `announce := blocker.BlockingTarget != attackerID`, overwrites `BlockingTarget`, and emits `EventBlock{Target: attackerID}` for the new pairing. Nothing is done about the attacker the blocker left.
- `DeclareBlocker` runs no state checks (#492), so the harvested trigger waits on `PendingTriggers` (or its optional prompt) and is drained at the next priority boundary, whatever the final blocks are.
- Readers of "becomes blocked": `cards/effects/batch26_helpers.go:91-100` (`b26ArtifactCreatureYouControlBecameBlocked`, Cyberman Patrol) and `grazilaxx_illithid_scholar.go:48`. Both dedupe with `b18AttackerAlreadyBlocked` (`batch18_helpers.go:108-132`), which walks event history. A withdrawn `EventBlock` still counts in that walk, so if a different creature later really blocks the first attacker, its trigger is declined as "not the first". The count then comes out right only by accident, and only if the stale instance is still around.
- Client: `client/src/lib/contextMenu.logic.ts:794` offers "Re-declare blocker".

## Rules

[...]

PR #1087, opened 2026-09-19
feat(engine): conditional block restrictions (#750)

krakenhavoc:
Implements the engine half of ADR 0045's addendum **Decision 11** (block rules with a parameter) and **Decision 12** (block counts as bounds), for #750.

Closes nothing yet — the issue's card wave and Decision 13 are still open (see *Not in this PR*).

## What changed

**`server/internal/game/block_rules.go` (new)**

`BlockRule` is one printed block restriction that carries a parameter, derived from its source when a block is checked rather than written onto a creature (the `Restriction` bits have no room for a condition on the *other* creature — ADR 0045 §2, and restrictions.go says as much). Two halves, per the ADR's struct:

- `Pair(g, attacker, blocker, source) bool` — "can't be blocked except by Walls", "can't be blocked by creatures with power 2 or less", "creatures with power less than this creature's power can't block creatures you control".
- `Count(g, attacker, source) (min, max)` — "can't be blocked by more than one creature".

`CatalogBlockRules` is the hook, keyed by `CatalogAbilityKey` so a source that has lost all abilities imposes nothing (CR 613.1f) while an Equipment's rule survives the equipped creature losing *its* abilities. `blockerBoundsLocked` is Decision 12's replacement for `BlockerCountValid`: menace built in as a minimum of 2, largest minimum and smallest maximum win, and a minimum above a maximum makes the attacker unblockable.

**`block_legality.go`** — fills the slot 4 the addendum reserved, and claims three of its reserved reason tokens (`cant_be_blocked_by`, `cant_be_blocked_except_by`, `cant_block_attacker`), each with a player-facing sentence. `BlockRefusal.Label` carries the rule's printed clause so the sentence names it and the client never re-derives a rule (§6). Because `DeclareBlocker` already routes through `BlockPairRefusalLocked`, pair rules are refused at declaration, withheld by the legal-move enumerator and honoured by the #328 auto-pass signal with no change to any of the three (§3).

**`blockers.go`** — the lock-in's close-out reads `blockerCountValidLocked` instead of `BlockerCountValid`, so a **maximum** is enforced alongside menace's minimum. An illegal set is reverted whole rather than trimmed: which blocker to drop under a maximum of one is the defender's choice, not the engine's.

**`block_rules_test.go` (new)** — pair refusal through `DeclareBlocker` (reason, source, label, sentence, and that the blocker is not staged); the `source` argument as the rule's scope; a rule dying with the permanent that printed it; a maximum reverting an over-full block and accepting a legal single; menace's bounds combining with a rule maximum; and the two menace cases below.

## The defect #750 documents, and where it stands

The issue (verified at `ad9413f`) reports that an illegal lone block on a menace attacker is accepted, logged, rendered, **fires "whenever this blocks" triggers**, and is then silently reverted at the damage step.

On current `develop` the trigger half is already gone: #830 moved the revert into `commitBlockDeclarationLocked`, where it runs as pass 0, *before* any `EventBlock` is emitted. That was previously untested, so `TestMenaceLoneBlockFiresNoBlockTriggers` now pins it: no `EventBlock` on the blocker (CR 509.3a), no `EventBecomesBlocked` on the attacker (CR 506.4), and the attacker is not recorded as blocked. `TestMenaceTwoBlockersAnnounceNormally` guards the legal case against the generalisation.

What is **not** fixed: the lone block is still *accepted* by `DeclareBlocker` and reverted silently at the lock-in instead of being refused with a reason. See below.

## Not in this PR, and why

- **Decision 13 — refusing an illegal COUNT at declaration.** This needs `DeclareBlockers(decls []BlockDeclaration)`, and `DeclareBlocker` becoming a one-entry call, so the engine has a complete declaration to judge. `DeclareBlocker` lives in `server/internal/game/mutations.go`, **304,966 bytes**, over this PR's 35,000-byte size ceiling. A count cannot be moved into the per-pair check instead: refusing the *first* of two menace blockers would make every menace attacker unblockable. No count `BlockReason` is minted here for the same reason — nothing can send one yet, and `block_legality.go` says a token only ships in the PR that first returns it.
- **`Game.TurnScopedBlockRules`** (Gingerbrute's until-end-of-turn rule) needs a field on `Game`, and `game.go` is 73,545 bytes — also over the ceiling.
- **`BlockRule.Limit`** (Silent Arbiter) — Decision 12 leaves it unbuilt until its first card.
- **The catalog's half** — `effects.Spec.BlockRules`, its line in `effects.buildDef`, `CardDef.BlockRules`, and the first cards (Prowler's Helm, Hungering Hydra, Gingerbrute). The hook ships ahead of them for the reason `land_drops.go` gives for `CatalogAdditionalLandPlays` ("No catalog card declares AdditionalLandPlays yet… so that when Exploration and Azusa are written they are a one-line Spec field and not an engine change").
- **Stale comment left behind:** `BlockerCountValid` in `keywords.go` still says it is called from the lock-in. Nothing calls it now; the addendum's PR 2 deletes it. Flagged in `block_rules.go`'s doc rather than churning `keywords.go` while other work is in flight.

## Checks

`gofmt -l` clean on all four files. The suite was not runnable here; CI is the backstop. Each pushed blob's SHA was verified against a local `git hash-object` of the intended content.

https://claude.ai/code/session_0132cK54Ko2vjtXR3ba1GxFz


krakenhavoc:
Holding the merge: CI fails at Server — lint: `internal/game/block_rules.go:175:67: (*Game).blockerBoundsLocked - result src is never used (unparam)`. Either drop the result or use it at a call site (or `//nolint:unparam` with a reason if it is deliberately there for a follow-up). I'll merge once CI is green.

krakenhavoc:
Thanks for the fix and the merge — your `92b01255` (drop the result) is the better of the two options I was choosing between, and it's what shipped.

One loose end: our two fixes crossed. I pushed `55193ea` (keep `src`, `//nolint:unparam`) after you had already merged `92b01255`, so that commit is sitting on `feat/750-conditional-block-restrictions` **unmerged** and it re-adds the result you just removed. Please delete the branch rather than re-merging it — nothing in it is wanted on top of `develop`. I don't have branch-delete access to do it myself.

Follow-ups this PR deliberately left, for whoever picks up the next slice of #750:

- Decision 13's refusal of an illegal COUNT at declaration (needs `DeclareBlockers`, i.e. `mutations.go`).
- `Spec.BlockRules` / `buildDef` / `CardDef.BlockRules` and the first cards, which is what makes `CatalogBlockRules` non-nil in production.
- `BlockerCountValid`'s doc comment in `keywords.go` still claims it is called from the lock-in; nothing calls it now.

PR #1088, opened 2026-09-19
feat(client): opponent boards as read-outs — derivations, panel, settings, ADR 0077

krakenhavoc:
## Summary

The opponent-board summary from ADR 0077, in reviewable layers. Nothing here is wired into the table yet — that waits for a follow-up on top of #1072, which has merged.

- `client/src/lib/seatSummary.ts` — the pure derivations: untapped mana by colour, creature counts and power split tapped/untapped, the structural-permanent filter, the eliminated-seat case.
- `client/src/lib/seatSummary.test.ts` — 34 tests over those four functions.
- `client/src/lib/components/board/SeatSummary.svelte` — the panel that draws them.
- `client/src/lib/settings.ts` — `display.opponentDetail`, `expandActivePlayer`, `expandStyle`; `SETTINGS_VERSION` 10 → 11.
- `client/src/lib/settings.test.ts` — the v10 → v11 migration and the new defaults.
- `client/src/lib/components/Settings.svelte` — the three Display-tab rows that drive them, at the top of the tab.
- `docs/decisions/0077-opponent-board-summary.md` — the ADR.

## Sprint

S32 — Playtest stabilisation, round 1

**Not the active sprint.** S44 is, going by the ADRs merged 2026-09-18. This work belongs to S32 because #956 is an in-app playtest report and "triage the pain points from the real game" is S32's stated scope. Flagged here per AGENTS.md §4 rather than left for a reviewer to notice.

## Issues

Relates to #956. Does **not** close it — the layout fix in #1072 did.

## Two contracts worth reviewing as contracts

**The summary never ranks.** No scoring, no "the two or three permanents that matter". A heuristic that ranks threats is wrong at the worst possible moment and gets blamed for the loss — fairly, having chosen what to hide. The rule is structural: planeswalkers, commanders, anything wearing an attachment. Wrong only in ways a player can predict.

**The mana read under-reports on purpose.** A dual land counts once, not once per printed ability, because it can only be tapped once — summing would report a Triome as three mana. Anything whose colour can't be pinned down goes to `flexible` and is never folded into a colour. Several tests assert that explicitly; they are the contract, not a gap in it. A confidently wrong mana read is worse than none, because the player acts on it.

## What reviewers should push back on

**`opponentDetail` defaults to `"summary"`, so existing players' tables change appearance on upgrade** without them touching anything. Every earlier migration in that chain changed behaviour; this is the first that changes what the table looks like. Deliberate — the full-card rendering clips at small panel sizes and has nowhere left to shrink — and `"full"` restores the old look exactly, from the first row of the Display tab. But it's the decision most worth disagreeing with.

**`expandStyle` ships two mechanisms knowing one gets deleted.** Reflow keeps combat arrows in one plane; overlay animates better but crosses a z-index boundary. It's a feel judgement a prototype can't settle, so both ship behind a setting flagged `experimental` in the UI. If you'd rather not ship a doomed option at all, now is the time to say so.

## The ADR is 0077, not 0076

It was written as 0076. The tutorial ADR merged to `develop` the same afternoon and took that number, so both files carried an "ADR 0076" heading and `TestADRNumbersAreUniqueAndMatchTheirHeading` failed. Renumbered, the old file deleted, and the three code comments that pointed at 0076 moved with it. The ADR's own numbering note now records what happened rather than the claim that 0076 was free.

## Test plan

- [x] `npm run test` — `seatSummary.test.ts` (34) and `settings.test.ts` (22) both pass; reproduced locally against the pinned toolchain
- [x] `tsc` with this repo's `tsconfig.json` options over the changed `.ts` files and their imports — clean
- [x] `prettier --check` on every changed file, against the pinned prettier 3.8.2 / prettier-plugin-svelte 3.5.1
- [x] `eslint` on every changed file, against the pinned eslint 9.39.4 / eslint-plugin-svelte 3.23.0
- [ ] `npm run check` (`svelte-check`) — **not run**, see Provenance
- [ ] `TestADRNumbersAreUniqueAndMatchTheirHeading` passes for 0077
- [ ] Existing e2e unaffected: nothing mounts `SeatSummary` yet
- [ ] Manual: open Settings → Display, flip **Opponent boards** — the three rows render and persist, though nothing reads them until the Board wiring lands

## Provenance

This session cannot clone the repo (the API is 403 for git and for CI status), so everything reached GitHub through the MCP connector and CI is the only thing that has built it.

What *was* run: the client's dependency graph for the changed files was reconstructed locally from raw blobs and checked against the exact versions in `package-lock.json`. That caught the two things CI failed on — a prettier reflow in two help paragraphs, and `svelte/no-at-html-tags` on the keyword pip — plus the ADR collision above. `svelte-check` is the remaining gap: the component graph behind `SeatSummary.svelte` reaches too far to rebuild that way, so the `.svelte` files are typechecked only by CI.

Every file was verified byte-identical against its GitHub blob SHA after pushing.

## Still to come

The Board wiring and the expansion model, on their own branch now that #1072 has merged. Both touch `Board.svelte`, which is why they are not stacked here.

ADR 0077 stays **Proposed** until that wiring lands and §6 — which expansion mechanism survives — has an answer from a real game.


krakenhavoc:
Holding the merge — two things:

1. **Client — lint fails:** `client/src/lib/components/board/SeatSummary.svelte:262:51` — `{@html}` can lead to XSS (`svelte/no-at-html-tags`). Please render it without `{@html}`, or if the content is provably safe, justify a scoped disable. I haven't touched this since it's a security call.
2. **ADR number collision:** this PR adds `docs/decisions/0076-opponent-board-summary.md`, but `0076-tutorial.md` is already on develop (#1075). `TestADRNumbersAreUniqueAndMatchTheirHeading` will fail once lint passes. Per AGENTS.md §4, renumber this one (filename, H1, inbound links, PR title) to the first number free on every branch.

I'll merge once CI is green.

Issue #474, opened 2026-09-13
CI: actions/setup-go@v5 targets Node.js 20, forced onto Node 24 by the runner


krakenhavoc:
## Summary

Every `cmdctrl-ci` run now emits this annotation:

> Node.js 20 is deprecated. The following actions target Node.js 20 but are being forced to run on Node.js 24: `actions/setup-go@v5`.
> https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

Not breaking today — the runner is silently running it on Node 24 — but it is a warning on every single run, and "forced to run on" is a compatibility bet the runner is making on our behalf. It stops being free when the shim is removed.

## Where

`actions/setup-go@v5` is used in **three places**:

- `.github/workflows/ci-cd.yml:72`
- `.github/workflows/e2e-nightly.yml:65` (the `bot-games` job)
- `.github/workflows/e2e-nightly.yml:94` (the `playwright` job)

Every other action in the repo is already on a current major:

| action | version |
|---|---|
| `actions/checkout` | v6 |
| `actions/setup-node` | v6 |
| `actions/upload-artifact` | v5 |
| `actions/download-artifact` | v6 |
| `actions/cache` | v4 |
| **`actions/setup-go`** | **v5** ← the only laggard |

`scryfall-refresh.yml` uses no actions at all, so it is unaffected.

## Fix

Bump to the current `actions/setup-go` major. Two things to preserve when doing it:

1. **`go-version: "1.22"`** must stay — `server/go.mod` pins that minimum deliberately (method-aware `ServeMux`, 1.22+).
2. **`cache: false` must stay.** There is a long comment above each usage explaining why: these are self-hosted runners, `$HOME` survives between jobs so `~/go/pkg/mod` is already warm, and Go marks the module cache read-only (files 0444, dirs 0555) so `actions/cache` untarring over it fails with `/usr/bin/tar: ... Cannot mkdir` on every run and then silently re-downloads the whole module set. Check the new major has not changed the input's name or semantics before assuming the flag carries over.

Note `setup-node`'s `cache` input is an **enum** (npm/yarn/pnpm), not a boolean — `cache: false` is a hard error there and the input is omitted instead. Do not "fix" the two to match; they are different by necessity.

## Why file rather than fix now

[...]

PR #1109, opened 2026-09-19
ci: bump actions/setup-go to v6, off the deprecated Node 20 runtime (#474)

krakenhavoc:
## What

`actions/setup-go@v5` targets Node.js 20, which the runner now force-runs on Node 24 with a deprecation annotation on every `cmdctrl-ci` and `e2e-nightly` run. `actions/setup-go` v6.0.0 is exactly this upstream fix ("Upgrade Nodejs runtime from node20 to node24"), matching `actions/checkout@v6` and `actions/setup-node@v6` already in use right next to it.

## Preserved exactly

- **`go-version: "1.22"`** — untouched at every site; `server/go.mod`'s minimum is unaffected.
- **`cache: false`** — untouched at every site. Confirmed in `actions/setup-go`'s `v6` `action.yml` before assuming this carries over: the `cache` input is still a plain boolean (default `true`), same name and semantics as on `v5`. (`setup-node`'s `cache` stays a separate, enum-typed input, deliberately not touched — the two were never meant to match.)

## Scope: four sites, not two

The issue named two sites in `e2e-nightly.yml` (`bot-games` and `playwright`), plus `ci-cd.yml:72` — three total. The repo has moved since #474 was filed: a `bot-soak` job was added to `e2e-nightly.yml` with its own `actions/setup-go@v5`. All four sites are bumped here so the fix actually clears the annotation everywhere, not just where the issue could see:

- `.github/workflows/ci-cd.yml:72`
- `.github/workflows/e2e-nightly.yml:99` (`bot-games`)
- `.github/workflows/e2e-nightly.yml:266` (`bot-soak`, new since #474)
- `.github/workflows/e2e-nightly.yml:481` (`playwright`)

## Verification

- Both workflow files parse as valid YAML (`yaml.safe_load`).
- `git diff` shows only the `uses:` line changing at each site — `go-version` and `cache` lines are byte-identical before and after.
- CI is the real check here; watching this PR's `cmdctrl-ci` run for the deprecation annotation to be gone, per the issue's own verification checklist (annotation gone, `bot-games` nightly still runs `AISEAT_GAME_TESTS=1`, Playwright nightly still green, no `tar` cache warning).

Closes #474.

https://claude.ai/code/session_0132cK54Ko2vjtXR3ba1GxFz


krakenhavoc:
CI failed on this branch, and the cause is a real behaviour change in the bump rather than anything wrong with the four sites.

**`actions/setup-go@v6` exports `GOTOOLCHAIN=local` for the rest of the job.** v5 left Go's own `auto` default in place. Every other step wants `local` — the server builds and tests on the pinned `1.22`, which is `go.mod`'s minimum. The lint step does not: `make lint` is `go run golangci-lint@v1.64.8`, which requires go >= 1.23, so under `local` it refuses outright instead of fetching what it needs:

```
go: github.com/golangci/golangci-lint@v1.64.8 requires go >= 1.23.0
    (running go 1.22.12; GOTOOLCHAIN=local)
```

Side by side, same commit range, only the action version differing:

| Run | setup-go | `GOTOOLCHAIN` | lint |
|---|---|---|---|
| [35475495763](https://github.com/krakenhavoc/cmd_and_ctrl/actions/runs/35475495763) | v5 | `auto` | `switching to go1.26.8` → pass |
| [35475662846](https://github.com/krakenhavoc/cmd_and_ctrl/actions/runs/35475662846) | v6 | `local` | refused → fail |

Pushed e0982257: `GOTOOLCHAIN: auto` as step-scoped `env:` on **Server — lint** only, which is what v5 was silently giving it. Scoped to the one step deliberately — a job-level `auto` would let gofmt/vet/test/build drift off 1.22 without anyone noticing.

`e2e-nightly.yml`'s three sites need no equivalent, checked rather than assumed: every Go command in that workflow is a plain `go test` against the same 1.22 module, with no `go run` of a tool that outranks it.

The `cache: false` and `go-version: "1.22"` analysis in the PR body still holds — that part of the bump was fine.

Issue #1107, opened 2026-09-19
Cards: the Vivi Ornitier equipment/control deck — 30 buildable cards + 12 seam-blocked


krakenhavoc:
## Summary

A real imported deck (Vivi Ornitier, Izzet equipment + counterspells) was reviewed against the catalog. Of **87 distinct cards, 43 are already catalogued**; 2 are basics (synthetic shape, no entry needed); **42 are missing**.

Triaged: **30 are buildable on today's engine**, 12 are blocked on seams (most already tracked).

## Buildable — batched by mechanic

- **Commander + payoff:** Vivi Ornitier, Ojer Axonil, Deepest Might // Temple of Power
- **Equipment (cheap):** Bone Saw, Cathar's Shield, Hero's Blade, Dragonfire Blade, Thran Power Suit, Bilbo's Ring, Leyline Axe
- **Equipment (big / living weapon):** Aettir and Priwen, Excalibur Sword of Eden, Kaldra Compleat, Ultima Weapon, Mithril Coat, Batterskull, Batterbone
- **Attach enablers + creatures:** Brass Squire, Magnetic Theft, Enduring Curiosity (see #321)
- **Instants:** Counterflux, Devious Cover-Up, Discombobulate, Blue Sun's Zenith, Finale of Revelation
- **Artifacts / misc:** Chrome Mox, Mana Vault, Spellbook, The One Ring, Roadside Reliquary, Ugin Eye of the Storms

Vivi's `{0}: Add X mana in any combination of {U} and/or {R}, where X is Vivi's power` is expressible today: `ManaAbilityCost{}` (no tap — Runaway Steam-Kin precedent), `ProducedFunc` returning a repeated `{U|R}` pipe slot, and `Condition` for "only during your turn / only once each turn".

## Seam-blocked (do NOT ship these weaker-than-printed)

| Card | Seam | Tracked |
|---|---|---|
| Ancient Silver Dragon, Diviner's Portent, Invert Polarity | Effect-reachable RNG / die roll | yes |
| Narset's Reversal | Counter-to-hand / counter-to-zone | #299 |
| Mox Diamond | Discard cost on an entry replacement | #295 |
| Slip Out the Back, March of Swirling Mist | Phasing — no support at all | **untracked** |
| Roaring Furnace // Steaming Sauna | Rooms / unlockable doors | **untracked** |
| Three Steps Ahead, Return the Favor | Spree (modal + per-mode additional costs) | **untracked** |
[...]

Issue #343, opened 2026-09-11
[in-app] Anng swift savior transform missing


krakenhavoc:
I do not have an option on anng to transform him for 8 mana and I do not see him. Also Anng shows as a 0/0 when he blocks his 2/3 stats work but it is strange.

---

<details>
<summary>Client log — last 200 entries before the report</summary>

```text
16:07:21.656  received  snapshot seq=231 turn=5 seat=1 step=upkeep
16:07:21.726  received  snapshot seq=232 turn=5 seat=1 step=draw
16:07:21.733  sent      action pass_priority id=09d0251c
16:07:21.794  received  snapshot seq=233 turn=5 seat=1 step=draw
16:07:21.868  received  snapshot seq=234 turn=5 seat=1 step=precombat_main
16:07:39.951  sent      action cast_spell id=97e05cb4
16:07:39.987  error     server error code=insufficient_mana message=insufficient mana to cast
16:07:40.902  sent      action cast_spell id=73647fb6
16:07:40.940  error     server error code=insufficient_mana message=insufficient mana to cast
16:07:45.778  sent      action cast_spell id=03226173
16:07:45.827  received  snapshot seq=235 turn=5 seat=1 step=precombat_main
16:07:49.128  sent      action pass_priority id=ff18ecd7
16:07:49.172  received  snapshot seq=236 turn=5 seat=1 step=precombat_main
16:07:49.252  received  snapshot seq=237 turn=5 seat=1 step=precombat_main
16:09:25.430  sent      action pass_priority id=91fc8dc0
16:09:25.478  received  snapshot seq=238 turn=5 seat=1 step=precombat_main
16:09:25.558  received  snapshot seq=239 turn=5 seat=1 step=begin_combat
16:09:25.562  sent      action pass_priority id=d89d55bf
16:09:25.609  received  snapshot seq=240 turn=5 seat=1 step=begin_combat
16:09:25.681  received  snapshot seq=241 turn=5 seat=1 step=declare_attackers
16:09:27.400  sent      action pass_priority id=3bec6b7d
16:09:27.451  received  snapshot seq=242 turn=5 seat=1 step=declare_attackers
16:09:27.624  received  snapshot seq=243 turn=5 seat=1 step=declare_blockers
16:09:27.627  sent      action pass_priority id=7695e9eb
16:09:27.672  received  snapshot seq=244 turn=5 seat=1 step=declare_blockers
[...]

Issue #259, opened 2026-09-10
Waterbender's Restoration ships uncosted — waterbend {X} is free


krakenhavoc:
## Debt accepted knowingly in #255

`Waterbender's Restoration` landed in PR #255 (`feat(s22): flicker — return-from-exile + CR 603.7 delayed triggers`) with its **waterbend {X} cost not charged**.

Printed, waterbend {X} is a cost paid by tapping untapped permanents you control, and X — the number of creatures the spell can exile and return — is *defined by that cost*. The engine has no tap-permanents-as-cost component, and X here does not come from `{X}` in the mana cost, so there was nothing to charge.

The clause therefore shipped as **"any number of target creatures you control"**, for free.

## Why this one matters more than the usual simplification

Every other declared simplification in this catalog makes a card **strictly weaker** than printed (Y'shtola blinking once instead of twice, Cosmic Intervention without foretell, verges without their conditional mana). This one makes a card **strictly stronger**. An uncosted mass-blink distorts any playtesting done while it is live — it is not a smaller version of the card, it is a better one.

Merged deliberately rather than held, because the delayed-trigger engine in #255 unblocks work across both tracked decklists and should not wait on an unrelated mechanic.

## Fix

Blocked on **tap-permanents-as-cost**, which is the same cost component convoke needs. Per the Aang decklist triage that group is 7 cards — Clever Concealment and The Wandering Rescuer (convoke); Aang Swift Savior, Katara Water Tribe's Hope, The Legend of Kuruk, The Unagi of Kyoshi Island, and this card (waterbend {X}, which is convoke with a different name and no colour clause). One cost component serves both keywords.

When that lands:
- [ ] Charge waterbend {X} on `Waterbender's Restoration` and derive X from the number of permanents tapped
- [ ] Remove the "strictly stronger than printed" simplification note from the card file
- [ ] Remove it from the simplifications section of `docs/decklists/aang-is-so-flashy.md`

## Interim

If it distorts a playtest before then, the card can be dropped from the catalog — deleting its one file un-registers it and it reverts to manual sandbox behaviour.

PR #1118, opened 2026-09-19
feat(server): Vivi Ornitier and Ojer Axonil enter the catalog

krakenhavoc:
## Summary

- **Vivi Ornitier** ships `CompletenessFull` — both printed halves. The `{0}: Add X mana in any combination of {U} and/or {R}, where X is Vivi's power` ability is a no-cost `ManaAbilityCost{}` (Runaway Steam-Kin precedent) whose `ProducedFunc` returns one `{U|R}` pipe slot per point of **effective** power, so counters, Equipment and anthems all scale it and the controller picks a colour per slot.
- Both printed restrictions are really enforced: `AllConditions(DuringYourTurn(), ManaAbilityNotUsedThisTurn())`, the latter counting `EventManaAbilityActivated` in `g.EventsThisTurn()`. An unenforced once-per-turn would have shipped Vivi **stronger than printed**, so this was the gate on shipping the ability at all.
- **Ojer Axonil** ships `CompletenessCaveats`. Its damage clause is a `RepEventDamage` replacement with all three printed conditions — red source you control, noncombat only, opponent **player** — that raises to Ojer's power only when the amount is strictly less, and never lowers.

## Sprint

## Issues

Relates to #1107

## Notes

Ojer's dies trigger is **omitted, not faked**: transforming a permanent is an open seam (#343), and returning it untransformed would be a recurring 4/4 God instead of a land — strictly stronger than printed (#259). `TestOjerAxonilShipsWithoutItsDiesTrigger` pins the gap and flips when transform lands.

`AllConditions` was appended to `activation_conditions.go`; no existing function changed.

## Test plan

- [x] `go test ./internal/cards/...` — exit 0 after rebase onto develop
- [x] `gofmt -l internal/cards/effects/` — clean
- [x] 9 new cases: counter + per-opponent ping on a noncreature spell (and neither on a creature spell); three mana from power 3 with the second same-turn activation refused and no mana minted; refused on another player's turn; Ojer's yours-only / red-only / opponents-only / floor-not-bonus matrix; combat damage left alone
- [x] `TestNonCatalogSpellStaysSandbox`, `TestNoNewExactClonesInTheCatalog` pass


krakenhavoc:

Issue #762, opened 2026-09-17
ADR then implementation: token creation as a replaceable event (CR 701.7b, 614) — Parallel Lives, Anointed Procession, Academy Manufactor; clears 9 catalog caveats


krakenhavoc:
## Why

Creating a token never reaches the CR 614 window. Nothing can double it, add to it or swap it, and a created token also skips the entry replacements every other permanent gets. Checked on develop (`ad9413f`):

- `CreateTokensForEffect` (`server/internal/game/effect_api.go:1615`) builds each token from the template and calls `g.Battlefield.PushTop(tok)` (`:1654`), then emits `EventTokenCreated` and `EventETB`. It never calls `applyReplacementsLocked`, and it never calls `fireETBHookLocked` (`game/effect_hooks.go:462`). `CreateTokensAttackingForEffect` (`:1690`, push at `:1709`) is the same. `CreateTokenForEffect` (`:1568`) and the card-side `CreateTokenCopy` (`cards/effects/token_copy.go`) go through them.
- The `TokenEntryOptions.Counters` comment states the gap (`effect_api.go:1594-1599`): token counters skip the counter replacement pipeline, and `Tapped` is a field because tokens bypass `RepEventMove`.
- There are six replacement event kinds: draw, move, counter, life, damage, step (`game/replacements.go:52-57`). None is for token creation.
- ADR 0013 §7 promised one `enterBattlefieldLocked` helper covering token creation (`docs/decisions/0013-replacement-effects.md:181-190`). It was scoped out in favour of per-site calls (`:443-446`), and token creation never got its call.
- 218 card files in `cards/effects` call `CreateToken*` and use the returned IDs straight away. That is why pausing is the hard part.

Parts that already exist and can be reused: the CR 616.1 order prompt (`queueReplacementOrderPromptLocked`, `game/pending_choice.go:983`) and its resume dispatcher `applyResolvedReplacementEventLocked` (`:1196`), which needs a token case; `Optional` replacements (`replacements.go:343`) gated by `optionalReplacementResumableLocked` (`:663`); `RegisterTurnScopedReplacement`; `TurnTally` (#586); the Treasure / Food / Clue templates.

Catalog cards shipping with a caveat this seam clears (9):
- Token doubling: Doubling Season, Primal Vigor
- Opponents' tokens don't enter tapped: Urabrask the Hidden, Manglehorn, Kinjalli's Sunwing
[...]

PR #1120, opened 2026-09-19
feat(server): living weapon, and seven Equipment cards enter the catalog

krakenhavoc:
## Summary

- **Aettir and Priwen, Excalibur, Ultima Weapon, Mithril Coat, Batterskull, Batterbone** — `CompletenessFull`. **Kaldra Compleat** — `CompletenessCaveats`.
- `LivingWeapon()` lands as **one shared helper** in `living_weapon.go`, feeding all three living-weapon cards, with a "0/0 black Phyrexian Germ" row in `tokens_table.go`.
- **Excalibur needed no caveat** — `Spec.SelfCostModifiers` already took a per-cast amount, so the historic-permanent reduction is real: generic only, floors at zero, excludes itself.

## Sprint

## Issues

Relates to #1107

## Notes — two additive engine changes, both to avoid shipping a card wrong

1. **`Card.PrintedPTKnown`** (+ `copy.go`, snapshot capture/restore, and a `snapshot_drift_test.go` entry). `ToughnessIsKnown` deliberately skips CR 704.5f for a 0/0 **token template** (#683), so a Germ whose Equipment left **survived forever as a 0/0 blocker**. Found by a test, not by reading. Opt-in per token template, set only on the Germ row; `VariableToughness` still wins.
2. **`StaticAbility.DependsOnLifeTotal`** + a conditional layer-cache bump on `EventChangeLife`, mirroring the existing `DependsOnHandSize`/Psychosis Crawler precedent. Without it Aettir and Priwen would freeze at whatever size it had when something unrelated last invalidated the cache.

The living-weapon attach rides `CreateTokensThenForEffect`'s **continuation**, not the next statement: since #762 token creation goes through the CR 701.7b replacement window, and a Doubling Season board pauses it for a CR 616 ordering prompt, so the ID slice comes back empty exactly on the boards where it matters.

Kaldra's caveat: the equipped creature does not exile creatures it damages — that clause grants a **triggered ability to another permanent**, the already-tracked #754 seam, to whose card list Kaldra was appended.

## Test plan

- [x] `go test ./internal/cards/...` — exit 0 after rebase onto develop; full `go test ./...` and `go vet ./...` exit 0
- [x] `gofmt -l internal/` — clean
- [x] `living_weapon_test.go` (4): Germ created, attached, live 4/4 with vigilance+lifelink; **the {3} bounce leaves the Germ to die as a 0/0** — the test that found the `PrintedPTKnown` gap
- [x] `equipment_legends_test.go` (5): Aettir re-reads base P/T through a real `ChangePlayerLife`, +1/+1 counters land on top; Excalibur's reduction sums mana value, ignores non-historic and opponents' permanents, floors at zero; Mithril Coat's trigger dropped with no prompt when there is no legendary creature (CR 603.3d)


krakenhavoc:

PR #1121, opened 2026-09-19
feat(server): Brass Squire, Magnetic Theft and Enduring Curiosity

krakenhavoc:
## Summary

- All three ship `CompletenessFull`.
- **Heterogeneous two-target clauses are supported** — ADR 0065's flat clause list (`Clauses(...)` / `ctx.ClauseTarget(i)`) already shipped, with Bite Down using it live, and it works for `ActivatedAbility.Targets` as well as `Spec.Targets`. So Brass Squire's "target Equipment you control **and** target creature you control" is a genuine two-slot clause, not an auto-pick. Shared machinery (`TwoSlotAttachTargets`, `AttachClauseTargets`) appended to `attachments.go`.
- **Magnetic Theft** gets "(Control of the Equipment doesn't change.)" for free: `AttachForEffect` only ever writes the `AttachedTo` link and never touches `Controller`, so leaving both clauses unfiltered is already the printed behaviour.

## Sprint

## Issues

Closes #321, relates to #1107

## Notes

**Enduring Curiosity** is the interesting one. "If it was a creature" is checked once against the CR 603.10 LKI characteristic the harvester already threads through — a fact about the past, not a live condition. "It's an enchantment. (It's not a creature.)" is a CR 613.3 layer-4 type change with a while-it-remains duration (`StaticForDuration` + `DurationWhileSourceRemains`, ADR 0063), pinned to the returned object's own instance ID so it cannot leak onto another copy at the table or a fresh hard-cast. Only "Creature" is stripped; Cat and Glimmer stay, since the printed text doesn't say "loses all other types". Dying again as an enchantment correctly does not re-trigger.

On **#321**: that bug's own triage comment confirmed the card was never in the catalog at all — no file, zero grep hits — so there was never a trigger to misfire. Both things it was blocked on shipped separately; this is the card finally being written.

## Test plan

- [x] `go test ./internal/cards/...` — exit 0 after rebase onto develop; `go vet` exit 0
- [x] `gofmt -l internal/cards/effects/` — clean
- [x] 5 new tests, each asserting observable state: the Equipment's `AttachedTo` really points at the chosen creature and its static rides along; an Equipment you don't control is rejected; an opponent's Equipment moves without changing controller; hand size increments once per connecting creature; the returned permanent is `IsCreature()==false` / `IsEnchantment()==true` under its owner's control, and a second death does not re-trigger


krakenhavoc:

Issue #1115, opened 2026-09-19
Tokens have no art: resolve a Scryfall token printing at runtime and render it


krakenhavoc:
## Why

A token renders as its name in text on a card-shaped div (`client/src/lib/components/board/Card.svelte:345`). Treasures, Clues, Food and every creature token are grey rectangles with words on them, while every real card has art.

## What the survey found (2026-09-19)

- A token is a `game.Card` with an empty `ScryfallID` and `OracleID` (`server/internal/game/effect_api.go:1545`). The client asks for no image because there is no id (`client/src/lib/cardImage.ts:41`).
- The comment at `server/internal/cards/effects/tokens.go:9-13` says tokens "aren't a Scryfall printing". **That is not true of the data we already have.** The `default-cards` dump at `data/scryfall/` holds 2,958 `"layout":"token"` records, each with the full six-size `image_uris`.
- `Index.Load` keeps every record (`server/internal/cards/index.go:288`); token layouts are only *demoted* when resolving a deck's names (`isPlayablePrint`, `index.go:491`).
- **So `GET /cards/<token-printing-id>/image` already works.** The missing piece is choosing an id and stamping it on the card.
- The service worker's image-cache pattern already covers that URL shape (`client/src/sw/service-worker.js:67`), so reusing it needs no client cache change.
- 96 tokens live in `tokens_table.go`, plus ~15 one-off literals.
- The client cannot tell a token from a real card: there is no `is_token` on `CardView`.

## Decisions (owner, 2026-09-19)

1. **The printing is chosen at runtime by a fixed rule**, not pinned per token in the repo. A later feature will let players pick a token's art, so the resolver is a seam that an override can sit in front of.
2. **The chosen id goes on the existing `scryfall_id`**, so art renders with no client image changes, **and `CardView` gains `is_token`** so the client can style a token as a token.
3. **This ships before #521** (synthetic token catalog keys, S33 sub-PR 5). #521 is about a live Treasure blocking restore points; when it lands it carries the art id as registry data.

## Scope

- [ ] ADR 0078.
- [ ] A resolver: token name, type line, P/T and colors to a printing id, deterministic and cached, with the override seam.
- [ ] Stamping at token creation, covering the table rows, the behaviour tokens and the one-off literals.
- [ ] `is_token` on `CardView`, `docs/protocol.md`, and the client using it.
- [ ] Tests, including one that every token in the table resolves to a printing.

## Open, to settle in the ADR

[...]

Issue #734, opened 2026-09-16
S35 — Playtest stabilisation, round 2


krakenhavoc:
**Phase:** 6 · **Goal:** turn the in-app bug reports from the games since S32 into closed issues or named decisions. S32 ([#277](https://github.com/krakenhavoc/cmd_and_ctrl/issues/277)) exit criterion 6 said these reports would become the next sprint. That number went to S33 ([#515](https://github.com/krakenhavoc/cmd_and_ctrl/issues/515), surviving a deploy) and S34 ([#607](https://github.com/krakenhavoc/cmd_and_ctrl/issues/607), user database), so they become S35.

Built 2026-09-16 from the [#534](https://github.com/krakenhavoc/cmd_and_ctrl/issues/534) triage (verified on `main` @ `f1bdc37`, 2026-09-14) and the 18 open `[in-app]` issues. Every report was re-checked against `origin/develop` on 2026-09-16, after the fixes merged since #534: #537 (#530), #539 (#529), #606 (#599), #618 (#499), #646, #648, #649, #701 (#605), and today's card batches. **No reported card has landed in the catalog since #534, and none of the 18 reports is confirmed fixed on `develop`.** This is a tracker for review, not a design.

**Lanes are split so parallel agents don't collide.** *Engine rules* is `internal/game`. *Cards* is `internal/cards/effects`, one file per card. *Autopass and stops* is `client/src/routes/Game.svelte` and `priorityStops.ts`. *Client UX* is the rest of `client/`. None of these reports needs the *bots* lane: #596 happened in a bot game, but the fault is in the engine.

## The reports

| # | Report (one line) | #534 verdict | Re-check 2026-09-16 | Lane | Duplicate? |
|---|---|---|---|---|---|
| [#540](https://github.com/krakenhavoc/cmd_and_ctrl/issues/540) | Delighted Halfling tapped for mana the turn it entered | not triaged (filed after) | **Likely still reproducible.** `ActivateManaAbility` refuses a summoning-sick creature's `{T}` (`game/mutations.go` ~:3395). The auto-tapper checks neither in planning (`gatherTapSources`, `game/autotap.go` ~:157) nor in its executor, which sets `Tapped = true` directly (`mutations.go` ~:1170-1200). Paying by auto-tap can tap a sick mana creature. Pin it with a test first | engine rules | No. #530 was the opposite: sickness wrongly reported on non-creatures |
[...]

Issue #596, opened 2026-09-16
[in-app] Bouncing a token


krakenhavoc:
Bouncing a token back to hand returns an instance of the card back to hand when that is not a legal game action

---

<details>
<summary>Client log — last 200 entries before the report</summary>

```text
11:55:41.987  received  snapshot seq=28 turn=1 seat=0 step=end
11:55:41.988  sent      action pass_priority id=ce7f5085
11:55:42.024  received  snapshot seq=29 turn=1 seat=0 step=end
11:55:42.733  received  snapshot seq=30 turn=1 seat=1 step=upkeep
11:55:43.445  received  snapshot seq=31 turn=1 seat=1 step=upkeep
11:55:43.446  sent      action pass_priority id=464c9dc7
11:55:43.479  received  snapshot seq=32 turn=1 seat=1 step=draw
11:55:44.191  received  snapshot seq=33 turn=1 seat=1 step=draw
11:55:44.192  sent      action pass_priority id=cb3504c2
11:55:44.225  received  snapshot seq=34 turn=1 seat=1 step=precombat_main
11:55:44.930  received  chat from=Bot 1 text="Play Steam Vents"
11:55:44.933  received  snapshot seq=35 turn=1 seat=1 step=precombat_main
11:55:45.643  received  chat from=Bot 1 text="Steam Vents — pay 2 life so it enters untapped?: pay 2 life"
11:55:45.645  received  snapshot seq=36 turn=1 seat=1 step=precombat_main
11:55:46.359  received  chat from=Bot 1 text="Cast Lightning Bolt targeting krakenhavoc"
11:55:46.363  received  snapshot seq=37 turn=1 seat=1 step=precombat_main
11:55:47.076  received  snapshot seq=38 turn=1 seat=1 step=precombat_main
11:55:47.077  sent      action pass_priority id=1d7fbb83
11:55:47.110  received  snapshot seq=39 turn=1 seat=1 step=precombat_main
11:55:47.821  received  snapshot seq=40 turn=1 seat=1 step=precombat_main
11:55:47.822  sent      action pass_priority id=e0b474c9
11:55:47.857  received  snapshot seq=41 turn=1 seat=1 step=begin_combat
11:55:48.568  received  snapshot seq=42 turn=1 seat=1 step=begin_combat
11:55:48.569  sent      action pass_priority id=3aa181ea
11:55:48.601  received  snapshot seq=43 turn=1 seat=1 step=declare_attackers
[...]

Issue #526, opened 2026-09-14
[in-app] Smart autopass not honoring requested stops


krakenhavoc:
When autopass is turned on, and a manual stop is placed on a step, the game continues to autopass right through without ever stopping

---

<details>
<summary>Client log — last 200 entries before the report</summary>

```text
13:02:38.567  received  snapshot seq=17 turn=1 seat=0 step=begin_combat
13:02:39.279  received  snapshot seq=18 turn=1 seat=0 step=declare_attackers
13:02:39.280  sent      action pass_priority id=be722378
13:02:39.361  received  snapshot seq=19 turn=1 seat=0 step=declare_attackers
13:02:40.080  received  snapshot seq=20 turn=1 seat=0 step=declare_blockers
13:02:40.082  sent      action pass_priority id=a0d8f323
13:02:40.163  received  snapshot seq=21 turn=1 seat=0 step=declare_blockers
13:02:40.875  received  snapshot seq=22 turn=1 seat=0 step=combat_damage
13:02:40.876  sent      action pass_priority id=90a0ac94
13:02:41.008  received  snapshot seq=23 turn=1 seat=0 step=combat_damage
13:02:41.668  received  snapshot seq=24 turn=1 seat=0 step=end_combat
13:02:41.669  sent      action pass_priority id=72e750d3
13:02:41.747  received  snapshot seq=25 turn=1 seat=0 step=end_combat
13:02:42.459  received  snapshot seq=26 turn=1 seat=0 step=postcombat_main
13:02:42.460  sent      action pass_priority id=60a29d1b
13:02:42.540  received  snapshot seq=27 turn=1 seat=0 step=postcombat_main
13:02:43.254  received  snapshot seq=28 turn=1 seat=0 step=end
13:02:43.254  sent      action pass_priority id=454b6a34
13:02:43.335  received  snapshot seq=29 turn=1 seat=0 step=end
13:02:44.051  received  snapshot seq=30 turn=1 seat=1 step=upkeep
13:02:44.764  received  snapshot seq=31 turn=1 seat=1 step=upkeep
13:02:44.764  sent      action pass_priority id=21ac9233
13:02:44.847  received  snapshot seq=32 turn=1 seat=1 step=draw
13:02:45.559  received  snapshot seq=33 turn=1 seat=1 step=draw
13:02:45.560  sent      action pass_priority id=7f1e643f
[...]

PR #1124, opened 2026-09-19
docs(docs): ADR 0078 — tokens get real artwork, resolved at runtime (#1115)

krakenhavoc:
## Summary

- **ADR 0078 — tokens get real artwork.** Design only: no resolver, no server change, no client change. **Status is Proposed, and it is awaiting the owner's approval before any implementation.**
- Records the owner's decisions — a runtime rule rather than ids pinned per token (so a later user-selectable-token-art feature has a seam to sit in front of), the id on the existing `CardView.scryfall_id` plus a new `is_token`, and shipping before #521 — and decides the design content behind them: the matching rule and its tie-break, where resolution happens, cost, the failure case, the public catalog route, attribution, and double-faced tokens.
- **Adds the `docs/sprints.md` S35 section**, which had an index row (#734) and no body, and moves the S35 index row `planned` → `partial`.
- Also updates AGENTS.md §3's ADR range line, which was stale at 0075.

## Sprint

S35 — Playtest stabilisation, round 2 — **answering the open question the first revision of this PR raised.** #1115 and #1127 both carry the S35 milestone now.

## Issues

Relates to #1115 (tracking issue — **not** closed: this is sub-PR 0 of the ADR's own scope). Relates to #1127, split out of this work (below), and to #521, which the ADR explains is separable and deliberately later.

## Updated after owner review

**The 21 colourless templates are #1127, not sub-PR 4.** Matching every template against the dump found 23 of 123 that resolve to no printed token, 21 of them because the template declares `Colors` empty where every printed version is coloured. Correcting those changes what the tokens **are** — colour is read by removal, lords, protection, cost reduction and devotion regardless of art — so it is a rules-correctness bug with its own per-card review, filed as **#1127**, and this ADR now **edits no template at all**.

The consequences are recorded in decision 3d: until #1127 lands those 23 render as text, which is **expected rather than a failure**; when it lands their art resolves under the same rule with **no further work in this ADR's scope**. The board is mixed for a while, accepted deliberately over holding every token's art behind a per-card rules review.

**The resolver test follows suit.** `TestEveryTokenTemplateResolvesToAPrinting` compares against a checked-in `knownUnresolvedTokens` list rather than against zero misses, and fails on a **set difference in either direction**: a template not on the list that stops resolving is a failure (the guard), and a listed template that starts resolving is a failure saying "delete its entry" (so #1127 cannot leave a stale exemption). **The suite is green the whole time #1127 is open**, and the list is deleted with its last row.

**The 0/0 requirement is unchanged**, and the ADR now says explicitly that #1127 does not cover it: `&& !c.IsToken()` on `ToughnessIsKnown` (`card.go:911`) and `fromScryfallPrinting` (`mutations.go:5659`) lands **in the same commit as the stamping**, because it is needed the moment any token carries a `ScryfallID`, whatever #1127 has done by then.

## The S35 section

Written from #734's own lanes and the state of the issues it names — not invented. Sources, in order:

| Source | What it gave the section |
|---|---|
| #734 | the goal, the S32-criterion-6 lineage, the lane split, the six exit criteria, the "explicitly not in this sprint" list |
| Issue state + milestone on each report | which lines are `[x]`; that #343 left for **S46**, and that #492 → S37, #489 / #478 / #360 → S39, #482 closed unmilestoned |
| PRs #896, #898, #899, #908, #900, #904 | the engine-rules and autopass lines, with the commit each closed on. #500's owner decision was to **enforce** the land drop with a raisable limit, not to close it |
| PR #1018 | the five trigger verdicts: four are one fact (no registered `Spec`, so no trigger to fail to fire), each now pinned in `NeedsCatalogEffect`; the fifth, #370, was real and already fixed by #797 — the discard answer returned without running state checks. Also the Client UX line |
| `git log --grep='Sprint: S35'` | "Also shipped under the S35 name": **ADR 0075** table settings and host controls (#1042, #1043, #1110, #1104) and the `actions/setup-go` v6 bump (#1109) |
| `docs/decklists/card-coverage-roadmap.md` | only Monument to Endurance is on the roadmap, so exit criterion 1 is honestly *partly* met |
| PR #1103 | the house style for a sprint section, and the S32 section for the lane/exit-criteria/status shape a triage sprint uses |

**Index row `planned` → `partial`**, per the status legend in the same file: load-bearing work has shipped under the S35 name (#596, #526, ADR 0075 and its four PRs) while criteria 1, 4, 5 and 6 are unmet — and a stale `planned` row is the one the legend says does real damage, because agents read it as "this mechanic is unbuilt".

## What the ADR decides

- **The matching rule.** Eligible pool = `layout == "token"`, `lang == "en"`, `border_color == "black"`, `set_type ∉ {memorabilia, minigame, funny}`, non-empty `image_uris` → **2,722 records in 714 identity buckets**. Identity is `name` + *(card types, subtypes)* + `power`/`toughness` + `colors`, all exact and all required; `keywords` is a soft preference that narrows, never rejects. Type lines are parsed rather than string-compared, because the Halfling token prints `"Tolkien Creature — Halfling"`.
- **Tie-break: the lowest Scryfall UUID** among the best candidates. Deterministic, needs no new `released_at` field, and unlike "newest printing" does not churn the art on exactly the tokens reprinted most. Stability: permanent per token instance, total per process, best-effort across processes.
- **Resolution at token creation**, stamped on `Card.ScryfallID` — already classified `carried` in `snapshot_drift_test.go`, so art survives capture, restore, undo and replay with **no new `Card` field and no new drift-test row**.
- **Double-faced tokens out of scope**: **0 of the 119** `double_faced_token` records carry a top-level `image_uris`, and `viewOfFaces` would start emitting `?face=N` URLs by accident the moment a token carried both `Faces` and a `ScryfallID`. A token *copy* of a DFC already works correctly per CR 707.2.
- **`/catalog/image/{id}` keeps refusing token printings** — what makes that unauthenticated route safe is its bounded scope, not its content.
- **Attribution** recorded as an unaddressed housekeeping gap, not fixed here and not overstated.

## Test plan

- [x] `docs/decisions/0078-token-art.md` added; H1 number matches the filename.
- [x] Docs guard green in Docker (`golang:1.22`, whole repo mounted, `-v $HOME/go-mod-cache:/go/pkg/mod`): `go test ./internal/docsguard/...` → `ok …/docsguard`, exit 0. Re-run after each rebase, including the last one.
- [x] ADR number checked per AGENTS.md §4 — `git fetch origin` then `git ls-tree docs/decisions/` over all **385** remote heads. Highest anywhere is `0077-opponent-board-summary.md`; **0078 appears on no branch.**
- [x] AGENTS.md §3 range line updated (`0001 WS library … 0078 token art`). No other ADR index exists in the repo.
- [x] S35 section appended after S34 with the same `---` separator; index row updated in place. `docs/sprints.md` is not touched by any commit on `develop` since #1103, so no conflict.
- [x] Rebased onto `origin/develop` (`c35b5fee`); `develop` moved three times during this work and the branch was rebased each time.
- [x] Milestones set: `gh issue edit 1115 --milestone "S35 — …"` and the same for #1127, both verified back.
- [x] No code changes — nothing to build or run.

## Not in this PR

No template edits (#1127), no resolver, no `is_token`, no client change. **Do not merge before the owner accepts the ADR.**


krakenhavoc:
Heads up: #1128 (`feat/343-transform`, "transforming a permanent is not a zone change") had also taken **ADR 0078**, pushed four minutes after this branch.

**No action needed here** — this PR was opened first, so per AGENTS.md §4 it keeps 0078, and I renumbered the other one to 0079 (its filename, H1 and numbering note). Flagging it only because the collision is invisible from inside either PR: the filenames differ, so git merges both without a conflict and `TestADRNumbersAreUniqueAndMatchTheirHeading` would not have fired until both were on `develop`.

Issue #1112, opened 2026-09-19
Cards: the Women of the Sea pirate deck — 13 buildable, 9 caveated, 10 seam-blocked (+ transform #343)


krakenhavoc:
## Summary

The Moxfield deck **[Women of the Sea — Pirates and Ships](https://moxfield.com/decks/Yjlzz2_Zs0CruZU_vw8m-g)** (Mary Read and Anne Bonny, Izzet pirate tribal) re-triaged against `origin/develop`.

**85 distinct cards. 51 are already catalogued**, 2 are basics (synthetic shape, no entry needed) → **53/85 automated today**. **32 are missing.**

The existing triage doc [`docs/decklists/pirates-mary-read-anne-bonny.md`](../blob/develop/docs/decklists/pirates-mary-read-anne-bonny.md) is **stale**: it lists attack triggers, crew/Vehicles, end-step triggers, until-EOT pumps, per-mode target slots and becomes-target triggers as blockers, and every one of those has shipped since (S22, S25, S27, #764). Refreshing it is part of this issue.

Triaged: **13 buildable in full, 9 buildable with caveats, 10 seam-blocked** (one of which, Ojer Axonil, is already claimed by #1107).

## Buildable in full — batched by mechanic

- **B1 — Vehicles (crew, S27):** Jackdaw, Magmatic Galleon, RMS Titanic, The Indomitable
- **B2 — Pirates and artificers:** Captain Lannery Storm, Captain Howler Sea Scourge, Coercive Recruiter, Quicksmith Genius
- **B3 — Spells:** Mishra's Command, Prismari Command, Change of Fortune
- **B4 — Harder creatures:** Breeches the Blastmaker
- **B5 — Class:** Cool but Rude

Three of these are only buildable because of work that landed after the last triage: **Quicksmith Genius** needs reflexive triggers (#636, "when you do") for its rummage-not-loot ordering; **Mishra's Command** and **Prismari Command** need per-mode target slots (#764); **Cool but Rude** needs Classes (#757, ADR 0071); the four Vehicles need `CrewCost` (S27).

## Buildable with honest caveats

Shipping these with `CompletenessCaveats` rather than holding them — never stronger than printed, seam appended to `docs/engine-seams.md`.

| Card | What is deferred | Seam |
|---|---|---|
| Bag of Holding | the exile-and-return half | Post-departure LKI / "exiled with this" record |
[...]

PR #1125, opened 2026-09-19
chore(server): audit Completeness for ten Women-of-the-Sea catalog cards

krakenhavoc:
## Summary
- Audits ten unreviewed cards from the Mary Read and Anne Bonny ("Women of the Sea") pirate deck against their real Scryfall oracle text and stamps `Spec.Completeness`.
- Eight cards ship `CompletenessFull`: Captain Storm, Cosmium Raider; Corsair Captain; Mary Read and Anne Bonny (the deck's commander); Pull from Tomorrow; Reckless Fireweaver; Thrill of Possibility; Unexpected Windfall; Windfall.
- Two ship `CompletenessCaveats`: Decaying Time Loop (printed Retrace isn't modelled) and Ingenious Artillerist (the "one or more artifacts... deals that much damage" batching fires one trigger per artifact instead of one trigger for the whole batch, which is stronger than printed against a per-hit damage booster like Torbran).

## Sprint

## Issues
Relates to #1112

## Findings worth flagging

- **Ingenious Artillerist — real (documented, not fixed) stronger-than-printed edge case.** The card prints "Whenever one or more artifacts you control enter, this creature deals that much damage to each opponent." The engine has no way to read a batch count (`artifactEnteredUnderYourControl`'s own doc comment already flags this as a shared gap), so it fires one trigger per artifact instead of one trigger for the whole batch. The running total matches printed for an ordinary game, but a per-instance damage booster in the same catalog — Torbran, Thane of Red Fell (`+2` per red damage *event* a source you control causes) — would apply its bonus once per artifact instead of once per batch: two Treasures entering at once is 4 total damage printed, 6 total damage as currently implemented. This is the #259 "stronger than printed" direction, so I did not stamp it Full; I stamped `CompletenessCaveats` with a player-facing caveat instead of attempting an engine fix (reading a batch count for a scaling trigger is a real primitive gap, not a one-line fix, and it blocks exactly this one card today, so per `docs/engine-seams.md`'s own stated threshold — "a seam that blocks exactly one card ... is noted on that card's own batch issue rather than duplicated here" — I did not add a new seam-registry row; it's documented in the card file and this PR instead).
- **Decaying Time Loop's Retrace** needs both a cast-from-graveyard path (the same shape Flashback/Escape already use) *and* a typed "discard a land" cost component that `DiscardCost` doesn't have (it only takes an unfiltered count). Also a single-card seam today, documented the same way.
- No other bugs found. Mary Read and Anne Bonny (the commander) was audited clause-by-clause: haste via `PrintedKeywords`, the tap-loot via the S21 cost model, and the discard-into-Treasure trigger verified to fire off `EventDiscardCard`, which every discard path in the engine (loot, additional-cost discards, discard effects, cleanup-step hand-size discard) emits uniformly — nothing routes around it.

## Test plan
- [x] `docker run --rm -v "$WT":/src -v /tmp/claude-1000/gomodcache:/go/pkg/mod -w /src/server golang:1.22 go test ./internal/cards/effects/...` → `ok github.com/krakenhavoc/cmd_and_ctrl/server/internal/cards/effects 1.079s`
- [x] `docker run --rm -v "$WT":/src -v /tmp/claude-1000/gomodcache:/go/pkg/mod -w /src/server golang:1.22 go test ./internal/cards/coverage/ -update` (regenerated the census block in `docs/decklists/card-coverage-roadmap.md`: full 1604→1612, caveats 400→402, unreviewed 71→61 — exactly the 8 Full / 2 Caveats / 10 fewer Unreviewed this PR declares)
- [x] `docker run --rm -v "$WT":/src -v /tmp/claude-1000/gomodcache:/go/pkg/mod -w /src/server golang:1.22 go test ./internal/cards/coverage/...` → `ok github.com/krakenhavoc/cmd_and_ctrl/server/internal/cards/coverage 0.917s`
- [x] `docker run --rm -v "$WT":/src -w /src/server golang:1.22 gofmt -l internal/cards/effects/ internal/cards/coverage/` → clean
- [x] `docker run --rm -v "$WT":/src -v /tmp/claude-1000/gomodcache:/go/pkg/mod -w /src/server golang:1.22 go vet ./internal/cards/...` → clean


krakenhavoc:

PR #1126, opened 2026-09-19
feat(server): The Aetherspark enters the catalog

krakenhavoc:
## Summary

- **The Aetherspark** — `Legendary Artifact Planeswalker — Equipment`, the first card in the catalog that is a planeswalker and an Equipment at once. Ships `CompletenessCaveats` with exactly one caveat; the +1, the granted combat-damage loyalty trigger, the −5 and the −10 all ship in full.
- The +1 matters most here: a free repeatable attach that also puts a +1/+1 counter on the creature, which in an Equipment deck whose commander's mana equals his power is a permanent power-and-mana ratchet.

## Sprint

## Issues

Relates to #1107

## Notes

**Three questions settled while writing it:**

1. **A loyalty ability can carry a target clause** — unchanged from any other CR 602 activation (precedent: Teferi, Time Raveler's −3). `.WithCount(0, 1)` is the printed "up to one": the activation is legal with no target and still ticks up. The +1/+1 counter is deliberately **not** conditional on the attach succeeding — an Aetherspark that left in response still grows the creature, which is the printed behaviour.
2. **The granted trigger does not hit the #754 seam.** The ability is granted to The Aetherspark *itself*, not to another permanent, so it is a plain `Spec.Triggered` entry with an intervening condition. "That many" is captured as `ev.Amount` in `Build` (Cold-Eyed Selkie's pattern) and applied as loyalty counters at resolution.
3. **The −10 follows the Ugin precedent** — an activated ability with `LoyaltyCost(-10)` that uses the stack, not a `ManaAbility` (CR 605.1a). It uses the #742 one-pick-worth-ten grammar so it is **ten mana of one colour**, not ten independent colour picks, which would have been stronger than printed.

**The one caveat:** "Nothing can attack a planeswalker in this game yet, so 'The Aetherspark can't be attacked' never comes up." `DeclareAttacker` rejects any attack target that is not a seated player and the client only ever offers seats, so ADR 0032 §7 ("Absent") is still accurate. Adding a `game.Restriction` bit with no declare-attackers site to read it would be the badge-with-no-rule failure AGENTS.md forbids. A **new engine-seams row** — "Attacking a planeswalker, and 'can't be attacked'" — records it, noting CR 120.3c damage-removes-loyalty is missing alongside.

`Spec.StartingLoyalty` deliberately unset per ADR 0032 §1. Worth a follow-up: most existing catalogued walkers *do* set it as a fixture fallback, so the tree is inconsistent on this point.

No shared file modified — both new helpers live in the card file.

## Test plan

- [x] `go test ./internal/cards/...` — exit 0 after rebase onto develop
- [x] `gofmt -l internal/cards/effects/` — clean; `go build ./...`, `go vet`, docsguard/legal/catalog all green
- [x] 7 new tests: 3 combat damage → +3 loyalty; nothing while unattached; nothing on another player's turn; counters do not move before the trigger resolves; the +1 attaches and adds the +1/+1 counter; a no-target activation is legal and still ticks up
- [x] Clone gate and caveat-tone guards pass


krakenhavoc:

Issue #984, opened 2026-09-19
The public log has no case for EventColorChosen or EventCreatureTypeChosen, so a chosen colour or type is never narrated


irobinson010:
Noted by the #781/#740 agent (PR #983). `server/internal/protocol/log.go:~365`'s public-log switch has no case for `game.EventColorChosen` (`game/color_choice.go:51`) or `EventCreatureTypeChosen` (`game/creature_type_choice.go:47`), so the table never sees "Ian chose green for Coldsteel Heart" in the log even though the choice is public (CR 105.4) and #983 now shows it on the card. One `LogKind` each (or one shared "chose a value" kind with the value and the card), a line in `docs/protocol.md`, and the client's log renderer. #929's choose-a-player answer should get the same treatment in the same PR.

irobinson010:
Claiming this together with #986 on branch `fix/984-986-chosen-colour-surfaces` (base `develop`). Engine mechanics first, cards as proof.

irobinson010:
PR #1023 (base `develop`, head `5f1a6fae`, CI green first run).

**What landed.** Three `LogKind`s at the one public-log switch —
`choose_color`, `choose_type`, `choose_player` — sharing one new wire
field (`choice`: the colour LETTER, or the canonical creature type) and
one sentence shape, "P1 chose green for Coldsteel Heart". #929/#1007's
choose-a-player answer had the same hole and is in the same commit; its
answer rides `target_seat` rather than `choice`, like every other player
reference in the log, and deliberately never falls through onto
`target`, where an unseatable seat would have put a raw player UUID on
the wire.

Three kinds rather than one shared "chose a value" kind, because the
discriminator would be the kind: the client tones and filters a line by
`kind`, and `LOG_TONE` is a `Record<LogKind, …>`, so the union will not
compile without an entry.

**`choice` is redacted with the card's name.** CR 105.4 makes the
answer public at the table, but the answer also identifies the card —
which is exactly why #781 strips `CardView.chosen_color` /
`named_tribe` from a non-knower. A viewer who may not identify the card
gets "P1 chose a color for a card" and no `choice` at all. Nothing in
the catalog can produce that today (a face-down permanent has no
abilities, so it asks nothing), so it is the log failing closed ahead of
the engine, as `revealEntry` does. A card the view simply cannot account
for is a different case and keeps the colour: "P1 chose green for a
card".

**The mechanism.**
[...]

PR #1128, opened 2026-09-19
docs(docs): ADR 0078 — transforming a permanent is not a zone change

krakenhavoc:
## Summary

- **ADR 0078** settles the design for the largest open engine seam — turning a permanent over to its other face (CR 701.27, CR 712). The audit puts 47 cards behind it as their *only* blocker and 72 behind it in total.
- **Two verbs, deliberately kept apart.** An in-place transform (CR 701.27a, CR 712.18) keeps the object: `InstanceID`, counters, marked damage, attachments, the CR 613.7 timestamp, summoning sickness, controller, and every `Game`-side per-object registry. It changes only the printed characteristics, the layer cache and the catalog key. **"Exile it, then return it transformed"** (CR 712.14a) is two zone changes and a *new* object. Conflating them gives Fable of the Mirror-Breaker's chapter III a Saga that kept its lore counters, and Storm the Vault a permanent that re-triggers its own ETBs — wrong in opposite directions.
- **Design only, no code.** PR 2 (the verb + engine tests + the log line) and PR 3 (two proof cards from #1112) follow on branches stacked on this one.

## Sprint

S46 — Permanents that change what they are

## Issues

Relates to #343 (tracker), relates to #1112

## What the ADR settles

1. Transform is not a zone change — an explicit preserved / not-preserved list, including why a planeswalker back face does **not** get its CR 306.5b loyalty stamp.
2. The verb: `Game.TransformPermanentForEffect` in its own `internal/game/transform.go`, plus `effects.Transform{Target}` / `effects.TransformThis{}`.
3. `EventTransform`, and the `layerVersion` bump it owes — a face change is a *printed-value* change and `Effective()` serves a warm cache, so without invalidation a transformed Storm the Vault keeps reporting "Legendary Enchantment". Plus `LogTransform`, because `TestEveryEventKindIsNarratedOrDeliberatelySilent` (#984) requires a new kind to be narrated or listed as a deliberate silence, and a card physically turning over is the clearest "announced out loud" fact there is.
4. CR 701.27c/d, CR 712.4c, CR 712.15a: a permanent that can't transform does **nothing** — `nil` and no event, not an error and not `EventEffectError` (the catalog soak fails on that, and "transformed nothing" is a legal resolution). The layout allowlist is the load-bearing line: an `adventure` card also has two `Faces`.
5. `ExileAndReturnTransformedForEffect`, built from `ExileCardThenForEffect` + `ReturnFromExileToBattlefieldForEffect`, with the "set the face in the source zone before the CR 614 pipeline runs" step recorded as a deliberate choice rather than an accident.
6. **Transforming Sagas need no engine change**, and the ADR shows why: `IsSaga` reads the *effective* subtypes and `SagaFinalChapter` derives from the back face's (absent) chapter triggers, so CR 714.4 finds nothing to sacrifice either way.
7. The wire, the snapshot and clone all need nothing — ADR 0034 already carries `layout` / `faces[]` / `active_face` through `viewOfCard`, `cardSnapshot` and `cloneCard`. **No new `game.Card` field**, which is the deliberate part: "this permanent is transformed" is `ActiveFace != 0` (CR 701.27g reads exactly that way).
8. CR 712.8e: a nonmodal back face's mana value comes from its **front** face's cost. `ParsedManaValue` reads `c.ManaCost`, which for a `transform` back face is empty, so every transformed permanent in the game would be mana value 0 today.

Out of scope and stated explicitly: Incubator tokens, meld, day/night, the Siege grant's existing cast path, CR 701.27f's anti-loop clause, "as this transforms", a named `WhenThisTransforms` constructor, and Brass's Tunnel-Grinder (which also needs discover and descend, so the verb unblocks a third of it).

## A rule-number correction

Transform is **CR 701.27** in the pinned August 7, 2026 edition, not 701.28 — 701.28 is *Convert*. The seam registry and #1112's write-up both say 701.28, which was right in a pre-2025 edition and moved in the re-sort AGENTS.md §6 warns about. Verified against the downloaded `MagicCompRules 20260819.txt`, where 701.26 is Tap and Untap and 701.29 is Fateseal. Saga cards are CR 714 here (715 is Adventurer cards), matching the rest of the tree.

## Numbering

Checked on 2026-09-19 after `git fetch --all --prune`: all **385 remote branch heads** and all **259 local branch heads** listed with `git ls-tree --name-only <ref> docs/decisions/`, every commit reachable from any ref swept with `git log --all --name-only -- docs/decisions/`, and the ~90 sibling worktrees on this machine listed for uncommitted ADR files. Highest number present anywhere: **0077** (`0077-opponent-board-summary.md`). Nothing holds a `0078-*`.

## Test plan

- [x] `go test ./internal/docsguard/...` — the ADR guard (unique number, H1 matches filename):
  ```
  ok  	github.com/krakenhavoc/cmd_and_ctrl/server/internal/docsguard	0.002s
  ```
- [x] Docs-only change; no Go source touched, so nothing else can regress.
- [ ] PR 2 lands the verb and its tests; PR 3 lands Storm the Vault // Vault of Catlacan and Fable of the Mirror-Breaker // Reflection of Kiki-Jiki.


krakenhavoc:
**ADR number collision — renumbered this one to 0079** (pushed as `6afb0828`).

#1124 (`docs/adr-0078-token-art`, "tokens get real artwork") was pushed four minutes before this branch and claims the same 0078. This is the case AGENTS.md §4 is written about: the filenames differ, so git merges both cleanly, each PR is green on its own, and `TestADRNumbersAreUniqueAndMatchTheirHeading` only fires once both are in one tree — which would have been on `develop`, after the fact, with the number already in two PR bodies.

§4's tie-break is "the one that merges first keeps the number", and #1124 was opened first, so this file moved rather than that one. Changed: the filename, the H1, and the ADR's own **Numbering** paragraph — it asserted a sweep that found nothing holding 0078, which was true when written and is not any more, and leaving that standing would mislead the next person who reads it as precedent. Nothing links to this ADR yet, so there were no inbound references to fix.

I checked 0079 is free across **every** open PR branch rather than just this one — listed `docs/decisions/` on each of the 18 open heads; only these two claim anything above 0077.

`go test ./internal/docsguard/...` green.

Issue #1027, opened 2026-09-19
A discard PROMPT has no continuation, so Archon of Cruelty runs its last three clauses out of order


irobinson010:
Found while migrating the prompted-sacrifice payouts for #1019 on branch `feat/1019-1013-player-sacrifice-continuation-and-bot-cost-choice`.

`b09ArchonOfCrueltyTrigger` (`server/internal/cards/effects/batch09_helpers.go:128`) is the one caller of `g.PlayerSacrificesForEffect` with a clause after it that #1019 did **not** migrate, and the reason is a second missing continuation rather than a judgement about this card.

Printed: *"target opponent sacrifices a creature or planeswalker of their choice, discards a card, and loses 3 life. You draw a card and gain 3 life."*

Today the body queues the sacrifice prompt, queues the DISCARD prompt beside it, and then runs the life loss, the draw and the life gain — all three on the line after two open questions. Every one of those clauses is about a PLAYER and happens whichever way the prompts are answered (ADR 0013 §5m item 5), so nothing is wrong except the ORDER, and the order is observable.

**Why #1019 could not finish it.** The sacrifice half now has a continuation (`g.PlayersSacrificeThenForEffect`, ADR 0013 §5x), but the discard does not. `g.QueueDiscardChoiceForEffect` returns before the cards move, and the payout lint's own `discardVerb` row says so:

> there is no continuation form for a discard yet — a discard is an exit (ADR 0013 §5g) and DiscardRandomForEffect returns before a paused leg lands, so write the wrapper beside g.ExileCardThenForEffect rather than reading back on the next line

So migrating half of it — sacrifice-then-(discard + everything else) — would trade one wrong order for another: the life loss would land before the discard instead of before the sacrifice, which is no closer to printed and is a change nobody asked for. The card is best moved once, after the discard prompt grows the same shape.

**Shape.** A `Then` on the discard prompt, in the same family as `SacrificePermanent.Then` (#993) and the prompted-sacrifice run (#1019): `g.QueueDiscardChoiceForEffect` with a continuation run once the picked cards have reached a graveyard, so a discarded commander's CR 903.9 prompt holds it too, plus the `dropDefault` row so a withdrawn prompt settles rather than stranding the card. Then Archon of Cruelty is one nested chain in printed order, and the `discardVerb` row of `exit_payout_guard_test.go` gets a fix to name.

Related: #1019, #993, ADR 0013 §5g / §5x.

[...]

PR #1130, opened 2026-09-19
feat(server): Cool but Rude, Malcolm, Alluring Scoundrel, Read the Runes

krakenhavoc:
## Summary

- **Cool but Rude** (`full`) — the first three-line Class in the catalog beyond the two ADR 0071 shipped with. Level 1's rummage is **one trigger per attack declaration** (`OncePerBatch`), not one per attacker; the "if you do" draw is the discard run's count, so declining draws nothing. Level 2's ping is gated at level 2, so the discards level 1 was already making at level 1 hit nobody. Level 3 is `BecomesLevel(3, …)` over Gamble's tutor body, with the random discard chained behind the search's prompt (discarding before the tutored card is in hand would draw from the wrong hand).
- **Malcolm, Alluring Scoundrel** (`full`) — counter, draw, discard, then the "four or more" read off the live board once the discard prompt has settled (#1027's run gives the card that really left the hand). The free cast is a per-instance `game.CastPermission` (ADR 0066) over the **one** discarded card object, priced `{0}`, `CastOnly`, flash-timed (CR 608.2g) and expiring at end of turn — the same grant-instead-of-inline-cast shape cascade, madness and suspend use, declared as that shared simplification in the card comment.
- **Read the Runes** (`full`) — recorded as blocked in two earlier triages under "variable-count sacrifice cost" and "resolution-time choose N of your own permanents". Neither applies: nothing here is a cast cost, and the choice is one permanent at a time. It is Torment of Hailfire's `PickOption` + `SacrificeChoice` chain (#568) pointed at its own controller, X repetitions asked strictly one at a time.

Nothing was skipped, so `docs/engine-seams.md` needs no new row. The two decklist triage entries that recorded these cards as blocked are corrected in place, and the coverage census is regenerated with `-update`.

One shared helper added: `PermanentsControlledBy` in `resolution_choice.go`, beside the nonland one it copies the shape of — a bare "sacrifice a permanent" includes lands.

## Sprint

## Issues

Relates to #1112

## Test plan

- [x] `go test ./internal/cards/effects/...` — `ok … 1.141s`
- [x] `go test ./internal/cards/coverage/...` — `ok … 0.334s` (clone gate and census both green after `-update`)
- [x] `gofmt -l internal/cards/effects/` — no output
- [x] `golangci-lint run ./internal/cards/...` — exit 0
- [x] `go test ./... -timeout 20m` (whole server tree) — exit 0
- [x] Eleven new cases: Cool but Rude's three-attacker one-trigger, the declined rummage, the level-2 gate (a level-1 discard pings nobody; a level-2 discard deals 2 to each opponent), the level-3 tutor-then-random-discard; Malcolm's keywords, the sub-threshold hit, and the four-counter free cast asserted **all the way through a resolved cast out of the graveyard with an empty mana pool**; Read the Runes answering repetition one with a discard and repetition two with a sacrifice, the one-option list with no permanents, and X = 0.


krakenhavoc:

PR #1131, opened 2026-09-19
feat(server): four Vehicles for the Women of the Sea pirate deck

krakenhavoc:
## Summary
- Jackdaw — "you may discard your hand. If you do, draw a card for each artifact you control" as a single same-resolution `Effect` (CR 608.2c) — **not** a CR 603.12 reflexive trigger; see the fixup note below.
- Magmatic Galleon — ETB 5 damage to target opponent creature, plus a "whenever one or more creatures your opponents control are dealt excess noncombat damage, create a Treasure" watch reconstructed from `DamageMarked` before/after the event, batched with `OncePerBatch`.
- RMS Titanic — flying/trample, sacrifices itself and creates Treasures equal to the combat damage it dealt a player.
- The Indomitable — trample, draws a card whenever any creature you control deals combat damage to a player, and is castable from the graveyard when you control three or more tapped Pirates and/or Vehicles (a `CastCondition`/`CastConditionLabel` pair, scoped to the graveyard path only).

All four ship `CompletenessFull` — every printed clause turned out to have a shape already in the engine, so none needed a `Caveats` entry, and nothing was added to `docs/engine-seams.md`.

**Fixup (after initial review):** Jackdaw's "If you do" was originally built on `ReflexiveTrigger`, over-generalizing the Undead Butler / Generous Plunderer precedent (which both print "**When** you do" — a real CR 603.12 reflexive trigger) without checking that Jackdaw prints "**If** you do" instead. CR 603.12's own example ("you may sacrifice another creature. WHEN YOU DO, …") confirms the rule is keyed on "when you do"; "if you do" is ordinary same-resolution sequencing under CR 608.2c ("later text on the card may modify the meaning of earlier text" — the same mechanism "Counter target spell. If that spell is countered this way, …" uses). Two sibling batches in this run (#1130, #1135) independently reached the same conclusion for the identical "you may discard a card. If you do, draw a card" wording. Rebuilt as one `Effect` closure with no continuation machinery (the discard offers no choice, so nothing can pause between the two halves), and added a regression test that walks `StackMeta`'s keys from the "yes" answer to the stack emptying and fails if any key other than the parent trigger's own ever appears — confirmed it fails against the reverted reflexive version and passes against the fix. Checked Undead Butler and Generous Plunderer against this same question: both are correct as shipped (they print "When you do"), not a related bug.

## Sprint

## Issues
Relates to #1112

## Test plan
- [x] `go test ./internal/cards/effects/...` — all package tests pass, including 11 new tests across `jackdaw_test.go`, `magmatic_galleon_test.go`, `rms_titanic_test.go`, `the_indomitable_test.go`
- [x] `gofmt -l internal/cards/effects/` and `gofmt -l internal/` — both clean
- [x] `go vet ./internal/cards/effects/...` — clean
- [x] `go build ./...` — clean
- [x] `go test ./internal/cards/coverage/...` (with `CMDCTRL_SCRYFALL_DUMP`) — `TestNoNewExactClonesInTheCatalog` passes (no new duplicate helper bodies); `TestRoadmapCensusIsCurrent` regenerated via `go test ./internal/cards/coverage/ -update` after rebasing onto current `develop`
- [x] Regression check: reverted `jackdaw.go` to the original `ReflexiveTrigger` version and confirmed the new `TestJackdawIfYouDoIsSameResolutionNotAReflexiveTrigger` test fails against it, then restored the fix and confirmed it passes


krakenhavoc:

Issue #764, opened 2026-09-17
ADR then implementation: modal and multi-target clauses (CR 700.2, 601.2c) — per-slot target clauses, then per-mode targets, repeated modes, and modal triggers and activated abilities


krakenhavoc:
## Why

Modes exist for spells only. A target clause has a single predicate, and nothing can carry two differently-constrained targets or a mode choice on a trigger or an activated ability. Checked on develop (`ad9413f`):

- `ModeSpec` / `ModeOption` (`server/internal/game/modes.go:26-47`) give each option its own `*TargetSpec`. `validateModes` (`:66`) rejects a repeated index, so "you may choose the same mode more than once" (CR 700.2d) is impossible.
- `castTargetSpec` (`modes.go:89`) returns `ErrInvalidParam` when two chosen options both target. `effects.Register` panics at registration on "targeted modes with Max > 1" (`cards/effects/registry.go:45`), so Kolaghan's Command can't even be declared.
- `TriggeredAbility` (`game/triggers.go:73`) has `Targets` and `OptionalPrompt` but no modes. Its comment says modal triggers "need a separate ModePrompt slot" (`:119-122`). `effects.ActivatedAbility` (`cards/effects/spec.go:467-472`) is Label / Cost / Targets / SorcerySpeed / Effect, also with no modes.
- `TargetSpec` (`game/targets.go:28`) is one clause: one `CardOK` / `PlayerOK`, one `Zones`, one `Min`/`Max`. Multi-slot cards read `StackItem.Targets` by position and check each slot's real constraint only at resolution.
- ADR 0019 defers per-mode target slots on purpose (`docs/decisions/0019-structured-targeting.md:170-174`: "needs `StackItem.Targets` grouped by mode and a two-step picker").

Catalog cards already shipping with caveats because of this:
- Bite Down, Soul's Fire and Resourceful Defense: per-slot constraint checked only at resolution
- Aetheric Amplifier: only one mode of its modal activated ability (`cards/effects/aetheric_amplifier.go:21-31`)
- Gala Greeters: the modal trigger takes the first unused mode in printed order (`cards/effects/gala_greeters.go:23-24`)

## Cards

The audit's two clusters (`modal-multi-target`, `modal-multi-target-clauses`) are merged here. **54 cards where this is the only core blocker, 58 with it as any core blocker; the verifier estimated 30–40% of the sampled cards are also blocked by something else, or are misattributed (≈33–38 real).**

[...]

Issue #705, opened 2026-09-16
Landwalk (islandwalk etc.): enforce the block restriction; make Lord of Atlantis, Elvish Champion and Goblin King live


krakenhavoc:
## Why

Landwalk (CR 702.14: islandwalk, forestwalk, mountainwalk, swampwalk, plainswalk, and nonbasic landwalk) is **granted but not enforced**. Three shipped catalog cards are weaker than printed because of it:
- **Lord of Atlantis:** grants islandwalk, which is inert; the card declares a caveat.
- **Elvish Champion** and **Goblin King:** forestwalk and mountainwalk are deliberately omitted (`server/internal/cards/effects/tribal_lords.go` ~:22-28, :59, :65).

Batch PRs also skipped cards for it, e.g. Trailblazer's Boots and Nim Deathmantle's nonbasic landwalk/intimidate in #555.

Split out of the #176 umbrella (closed 2026-09-16), where landwalk was the keyword with the strongest forcing function.

## Structural blocker

A landwalk creature can't be blocked while the **defending player controls a land of that type**. That is a block restriction checked against game state. Today `CanBlock` can't see the defending player's battlefield (`server/internal/game/restrictions.go` ~:76-83 says landwalk's home is `CanBlock` "once CanBlock can see…").

## Scope

- [ ] Give the block-legality check access to game state (the defending player's lands), without breaking the legal-move enumerator's read-only contract.
- [ ] Canonical keyword tokens (`islandwalk` etc., CR 702.14c "nonbasic landwalk" variants), and enforcement in block legality: server validation, the `internal/legal` blocker enumeration, and the client's blocker highlighting (which reads server data, it doesn't derive rules).
- [ ] Layer interaction: granted landwalk (Lord of Atlantis is layer 6) and land types changed by effects (layer 4, e.g. Urborg or Blood Moon) are both read from **effective** characteristics.
- [ ] Cards: make Lord of Atlantis's islandwalk live and drop its caveat; add Elvish Champion's forestwalk and Goblin King's mountainwalk; re-check their Completeness.
- [ ] Tests: blocked/unblockable against lands of each type, a land type changed by an effect, granted vs printed, and the bot enumerator not offering an illegal block.
- [ ] Check the pinned Aug 2026 CR numbering (#684) before citing.

Related: #176, #555, ADR 0014, ADR 0045 (combat restrictions).

krakenhavoc:
[...]

PR #1133, opened 2026-09-19
feat(server): three pirate creatures — Breeches, Departed Deckhand, Kitesail Larcenist

krakenhavoc:
## Summary

- **Breeches, the Blastmaker** — `CompletenessFull`. The second-spell trigger, the optional artifact sacrifice, the keyed coin flip (ADR 0054) and both payoffs as CR 603.12 reflexive triggers, so each goes on the stack above the spell that triggered it.
- **Departed Deckhand** — `CompletenessCaveats`. Only the sacrifice-on-being-targeted trigger ships; both "except by Spirits" clauses are skipped rather than approximated with the flat `CantBeBlocked` bit, which would make it unconditionally unblockable (#259).
- **Kitesail Larcenist** — `CompletenessCaveats`. Flying and ward {1} ship; the enters trigger is deferred whole, because its chosen permanents *gain* a mana ability (#754) and shipping the type change without it is strictly better removal than printed.

### Two seams, honestly

**Conditional blocking restrictions (#750).** The ENGINE half has landed — `game.BlockRule` in `game/block_rules.go` is slot 4 of `BlockPairRefusalLocked`, and `blockerBoundsLocked` carries a minimum and a maximum. The CATALOG half has not: there is no `effects.Spec.BlockRules`, no `CardDef` slot and nothing sets `game.CatalogBlockRules`, which the file itself names as #750's PR 4 (first cards Prowler's Helm and Hungering Hydra, plus the enumerator-agreement tests). The Deckhand's `{3}{U}` ability waits on a second piece even after that: `Game.TurnScopedBlockRules`, named in the same file as deliberately unbuilt. The seam row said the slot was "reserved and empty", which stopped being true when #705 landed; this PR corrects it.

**Ability granted to another permanent by a static (#754).** Kitesail's three sub-questions came out one blocker, not three. The duration ("for as long as this creature remains on the battlefield") has a shape since #755 / ADR 0063 — `DurationWhileSourceRemains` + `StaticForDuration` + `SnapshotAffected`, what Sower of Temptation uses. Layer-4 type overwrite and `LoseAllAbilities()` have one since S24. What has none is granting `"{T}, Sacrifice this artifact: Add one mana of any color"` to another permanent: mana abilities are read off the permanent's OWN oracle ID through `CatalogAbilityKey`, and `Characteristic` carries granted keywords as strings with nowhere to put a cost and a production. The per-player target clause ("for each player, choose up to one other target…") is a second, smaller gap — #764's multi-clause statements are flat and fixed at catalog-build time, so a clause count derived from the seat count has no shape. It blocks one card today, so it rides the #754 row's entry rather than earning a row of its own.

## Sprint

## Issues

Relates to #1112. Records skips against #750 and #754.

## Test plan

- [x] `go test ./internal/cards/effects/... ./internal/cards/coverage/... -count=1` — `ok …/effects 5.208s`, `ok …/coverage 1.271s`
- [x] `gofmt -l internal/cards/effects/` — no output
- [x] `go test ./... -count=1 -timeout 25m` — exit 0, whole module
- [x] `go test ./internal/cards/coverage/ -update` for the roadmap census (2075 → 2078 registry keys, `full` 1604 → 1605, `caveats` 400 → 402)
- [x] Breeches: `TestBreechesCopiesOnAWinAndBlastsOnALoss` plays the whole card and reaches BOTH faces deterministically by burning N flips off Breeches' own keyed stream first. The Bolt is {R} for 3, so the outcomes are arithmetically distinct — a won flip copies it (6 life lost), a lost one deals the spell's mana value of 1 (4 life lost).
- [x] Breeches: declining the sacrifice makes no flip and no payoff; a controller with no artifact is asked nothing at all.
- [x] Deckhand: a spell that targets it kills it; an activated ability that targets it does not ("a spell", not "a spell or ability").
- [x] Deckhand / Kitesail: the deferred clauses are pinned by a test, so the day the seam closes the test asks to be updated.


krakenhavoc:

Issue #1117, opened 2026-09-19
Cards: the Betor, Ancestor's Voice life-swap deck — 28 buildable, 4 seam-blocked


krakenhavoc:
## Summary

A real imported deck — [*Life is just a resource we swap around*](https://moxfield.com/decks/gZKXEsf3OE6lQYShpWV1Cw), Betor, Ancestor's Voice (Abzan lifegain / life-swap) — was triaged against the catalog on `origin/develop` @ `183d8201` (2075 registry keys).

Of **95 distinct entries**: 37 declared `full`, 16 `caveats`, 6 `unreviewed`, 3 basics (synthetic shape), and **33 missing**. The deck imports today — nothing in it trips the Partner/Companion gate.

Triaged by four verifier passes: **28 are buildable** on today's engine (one of them behind a small engine fix), **4 are seam-blocked**, and 1 (The One Ring) is already being written under #1107.

## Buildable — batched by mechanic

- **Life swap (E1):** Axis of Mortality, Magus of the Mirror, Mister Negative, Tree of Perdition. No new engine primitive: `b31LifeBecomes` (`batch31_helpers.go:261`) is the "life total becomes N" idiom; an exchange is two of those with both totals captured first. Shared `exchangeLifeTotals` helper.
- **Commander + lifegain payoffs (E2):** Betor Ancestor's Voice, Rodolf Duskbringer, Enduring Tenacity, K'rrik Son of Yawgmoth.
- **Sacrifice / graveyard (E3):** Greater Good, Disciple of Bolas, Yawgmoth Thran Physician, Vaultborn Tyrant. Yawgmoth ships `full` — `protection from Humans` is expressible since ADR 0072.
- **Mill / goyf + Impending (E4):** Barrowgoyf, Six, Overlord of the Balemurk. Impending needs no engine primitive: `AlternativeCost.EntersWithCounterName/Count` plus a self Layer-4 static gated on time counters, written once as a reusable `Impending()` constructor.
- **Lands + Staff (E5):** Indatha Triome, Lush Portico, Underground Mortuary, Staff of Compleation. The two surveil lands are one missing row each in the `surveil_lands.go` table; `ManaAbilityCost.Life` already exists (Mana Confluence).
- **Spells + Offspring (E6):** Hatred, Ripples of Undeath, Darkstar Augur. Offspring is an **optional additional cost** (Kicker, ADR 0073), not an alternative cost.
[...]

PR #1136, opened 2026-09-19
feat(server): Betor's end step and three lifegain payoffs (#1117)

krakenhavoc:
## Summary
- **Betor, Ancestor's Voice** (the deck's commander), **Rodolf Duskbringer**, **Enduring Tenacity** and **K'rrik, Son of Yawgmoth** join the catalog. Two are `full`, two ship with a declared caveat.
- Betor's one printed end-step ability has two "up to one" target clauses pointing at different zones, which a single `*TargetSpec` cannot hold. It ships as two `AtYourEndStep` declarations — the deviation (two stack items instead of one) is the caveat.
- New shared file `life_swing.go`: `ManaValueAtMostLifeLostThisTurn` / `ManaValueAtMostLifeGainedThisTurn` (live predicates over `Game.TurnTally`) and `drainTargetedOpponent`.

## Sprint

## Issues
Relates to #1117

## Test plan
- [x] `go test ./internal/cards/effects/ -run TestE2 -v` — 7 new tests, all pass. Betor: counters equal to the life gained, the reanimation bound computed from life lost (with an over-budget card asserted absent from the picker), and a no-swing end step that does nothing and leaves no prompt open. Rodolf: the indestructible grant, the pay-then-reanimate line with the mana-value bound, and the decline path. Enduring Tenacity: the drain, and that it does not come back. K'rrik: a counter on a black spell, none on a white one.
- [x] `go test ./internal/cards/...` — `effects` and `cards` green; `coverage` green except `TestRoadmapCensusIsCurrent`, which is the census block every parallel #1117 batch moves. Per the batch brief that block is regenerated once on the rebased tip, so it is deliberately not touched here.
- [x] `go vet ./internal/cards/...` clean; `gofmt -l internal/cards/effects/ internal/game/` clean.
- [x] Clone gate (`TestNoNewExactClonesInTheCatalog`), caveat contradiction and adoptable-gap gates all pass.

### Dropped clauses, and why
- **Enduring Tenacity** — "When this dies, if it was a creature, return it to the battlefield. It's an enchantment." There is no per-instance "came back without its creature type" state. Returned as an ordinary creature it would die, trigger and return forever — *stronger* than printed, which is the one direction that is never allowed, so the clause is dropped instead of approximated.
- **K'rrik** — "For each {B} in a cost, you may pay 2 life rather than pay that mana." The S28 cost-modifier engine only changes the amount of generic mana; `ColorRequirement.Phyrexian` is set by parsing the printed cost with no hook to override it. K'rrik's **own** `{B/P}` pips need no declaration and already work on both the cast and activation paths.


krakenhavoc:

PR #1138, opened 2026-09-19
feat(server): four life-swap cards — Axis of Mortality, Magus of the Mirror, Mister Negative, Tree of Perdition (#1117)

krakenhavoc:
## Summary
- Four life-swap cards for the Betor deck, all `CompletenessFull`: **Axis of Mortality**, **Magus of the Mirror**, **Mister Negative**, **Tree of Perdition**.
- One shared `exchangeLifeTotals` / `exchangeLifeTotalsThen` (`effects_life_exchange.go`) — both totals captured before either write, each half through the CR 614 window. The `Then` form reports the delta *applied* to the first player, which is what Mister Negative's "if you lost life this way, draw that many cards" reads rather than arithmetic on the raw totals.
- Tree of Perdition's toughness half is a layer 7b **set with no stated duration** (CR 611.2a), registered at resolution with `IndefiniteDuration()` and pinned to `(InstanceID, EnteredBattlefieldAt)` so a flickered Tree comes back 0/13. Repeated activations register a new scoped static rather than replacing the old one — correct under CR 613.6 timestamp order, and said so in the card file.

Two append-only shared additions: `DuringYourUpkeep()` in `activation_conditions.go` (the step **and** the active player — Magus's clause is CR 602.1b, not a timing speed) and `firstLegalPlayerTarget` in `helpers.go`.

The exchange is two independent replacement windows, not one atomic event. Nothing in the catalog can observe the difference today, so it is documented in `effects_life_exchange.go` rather than declared as a `Caveat` — a caveat has to be a sentence a player can act on.

## Sprint

## Issues
Relates to #1117

## Test plan
- [x] `go test ./internal/cards/effects/... -race` — **ok** (18.9s). 11 new cases in `life_exchange_test.go`: both directions of each swap, the declined "you may", the sacrifice paid at announce, the condition refused in your own draw step *and* in an opponent's upkeep, Mister Negative's 5-card draw and its no-draw-on-gain, and Tree of Perdition's opponent life + `effectiveToughness` after one activation and after a second that swaps from the new values.
- [x] `go vet ./internal/cards/...` — clean.
- [x] `gofmt -l internal/cards/effects/ internal/game/` — clean.
- [x] `go test ./internal/game/` — **ok** (12.4s).
- [x] `go test ./internal/cards/coverage/` — the clone gate and the caveat-tone checks pass; **only** `TestRoadmapCensusIsCurrent` fails, on the census block every parallel #1117 batch conflicts on. Left for the final regeneration PR per the batch brief.
- [x] All four oracle IDs verified against the local Scryfall dump (name, type line, mana cost, P/T and oracle text all match).


krakenhavoc:
Superseded by #1152, which merged this branch's commits verbatim (`--no-ff`, authorship preserved) together with the other three, so they shared one CI run. The conflicts here were textual — two branches appending helpers at the same anchor in `activation_conditions.go` / `helpers.go`, plus the generated census block — and all of them were kept; nothing was dropped. The work is on `develop`.

Issue #652, opened 2026-09-16
ADR: granted cast permissions — Snapcaster Mage, Past in Flames, The Grim Captain's Locker, Underworld Breach


krakenhavoc:
## Why

S29 shipped casting from non-hand zones for cards that **print** the permission: flashback (#411), escape (#510), warp (#413), and per-card castable zones (#409). But the cast path can only ask the **catalog** whether a card may be cast from the graveyard, and under which alternative cost. Nothing lets an effect grant that permission to a *different* card:

- `game/cast_zones.go:51-82`: `CastableZonesFor` / `CardCastableFromZone` read only `CatalogCastableZones(oracleID)`.
- `game/alternative_cost.go:289-300`: `AlternativeCostByKey` reads only the catalog, so `validateAlternativeCost` (`:314-326`) rejects a `"flashback"` claim on a card that doesn't print flashback. `altCostExilesFromStack` (`:503`) is catalog-only too.
- `protocol/view.go:1417-1420`: `castable_here` for graveyard cards is catalog-only.
- The only per-instance grant is `ExilePlayPermission` (`game/exile_play.go:26-140`). It is exile-only, carries no alternative-cost key and no timing bypass: a granted cast still goes through the sorcery-speed gate (`game/mutations.go:740-744`; `cascade.go:36-43` says so).

Cards waiting (none in the catalog):

| card | shape |
|---|---|
| Snapcaster Mage | one target card gains flashback until end of turn; cost = its mana cost |
| Past in Flames | each instant and sorcery card in your graveyard gains flashback until end of turn; the set is locked on resolution (CR 611.2c) |
| The Grim Captain's Locker | `{T}`: each creature card in your graveyard gains "Escape—{3}{B}, Exile four other cards" until end of turn. Activated and locked, so the **same shape as Past in Flames**, not Breach's |
| Underworld Breach | **static**: each nonland card in your graveyard has escape while Breach is on the battlefield; cost = the card's mana cost + exile three other cards |

Later consumers of the same seam: foretell (#658, a grant keyed `"foretell"` on an exiled card), madness (#657, keyed `"madness"`, must ignore timing), suspend (#659, a free cast).

**Decision (2026-09-16): ADR first; implementation is follow-up PRs.**

## The ADR should decide

[...]

PR #1141, opened 2026-09-19
feat(server): Barrowgoyf, Six, Overlord of the Balemurk, and impending

krakenhavoc:
## Summary
- **Impending (CR 702.175)** as a reusable constructor: `Register(Impending(5, "{1}{B}", Spec{…}))` attaches the alternative cost (whose rider is escape's `EntersWithCounter*` entry counters), the self-only layer 4 "isn't a creature while time counters remain" static, and the end-step countdown trigger in one call. Each piece alone is a wrong card and two of the three are wrong in the *dangerous* direction, so the whole Spec goes through one function and a card file cannot ship two of the three. The other nine Overlords are now one line each.
- **Three cards**: Barrowgoyf (`full` CDA + combat-damage mill, one caveat for the known graveyard layer-staleness gap), Six (`reach` + attack mill; retrace dropped, seam #652), Overlord of the Balemurk (`full`).
- **Shared, in append-only mechanic files**: `notACreature` in `helpers.go` — batch31's "isn't a creature" static now calls it instead of keeping its own copy — and `mill_then_take.go`'s "from among them" / "from your graveyard" prompts, used by all three cards.

## Sprint

## Issues
Relates to #1117

## Test plan
- `go test ./internal/cards/effects/` — **pass** (new `mill_impending_test.go`: 8 tests covering the CDA across all graveyards, the combat mill and its decline, Six's attack mill and land filter, the impending wiring, impended vs hard-cast entry, and the full five-end-step countdown including "cannot attack while counters remain" and the attacks half firing once it wakes)
- `go test ./internal/game/... ./internal/legal/...` — **pass**
- `go test ./internal/cards/coverage/ -run 'TestNoNewExactClones|Caveat|Mechanic'` — **pass** (clone gate and caveat guard clean; the `notACreature` extraction removes a duplicate rather than adding one)
- `go vet ./internal/...` — clean
- `gofmt -l internal/cards/effects/ internal/game/` — clean
- `TestRoadmapCensusIsCurrent` **fails on the census block only** (2075→2078 keys). Per the batch brief that block is regenerated once, on the rebased tip, by a later PR.

## Notes for review
- Barrowgoyf's caveat describes the layer-cache invalidation gap it shares with Tarmogoyf, Lord of Extinction and Multani. **Delete it when `feat/1117-e8-layer-invalidation` merges** — that branch closes exactly this.
- Six's dropped retrace clause needs three things: the alternative cost, a cast permission over a *set* of cards (seam #652), and a duration scoped to its controller's turn. Left out rather than shipped as a line the engine never honours.


krakenhavoc:
Superseded by #1152, which merged this branch's commits verbatim (`--no-ff`, authorship preserved) together with the other three, so they shared one CI run. The conflicts here were textual — two branches appending helpers at the same anchor in `activation_conditions.go` / `helpers.go`, plus the generated census block — and all of them were kept; nothing was dropped. The work is on `develop`.

Issue #753, opened 2026-09-17
ADR then implementation: turn machinery — extra combat and main phases (CR 500.8) and extra turns (CR 500.7); fix Turn.IsNewTurn and define Turn.Number


krakenhavoc:
## Why

The engine can't add a phase, a step or a turn. Checked on develop (`ad9413f`):

- `turnSequence` is a fixed list of 12 steps (`server/internal/game/turn.go:41-54`). `Turn.advance` (`turn.go:169-197`) moves to the next index, and after cleanup it always rotates to `(ActiveSeat + 1) % numSeats` (`:185`). ADR 0026 records this as a known limit (`docs/decisions/0026-delayed-triggers.md:174-177`, "No extra steps").
- **`Turn.IsNewTurn` compares seats** (`turn.go:158-160`: `t.ActiveSeat != next.ActiveSeat`). `onTurnAdvanceLocked` (`game.go:718-753`) returns early unless it is true. If a seat takes two turns in a row, the second turn gets no `layerVersion` bump, and `LoyaltyActivatedThisTurn`, `SpellsCastThisTurn`, `LandsPlayedThisTurn`, `DrawnThisTurn` and `TurnTally` are never cleared.
- **`Turn.Number` counts rounds, not turns.** It goes up only when the cursor wraps to seat 0 (`turn.go:186-189`), but the doc comment on `Turn` (`:94-96`) says it is "the turn number". It is referenced about 35 times outside tests, and several already work around it: `turn_scoped_statics.go:93-99`, `batch19_helpers.go:103-121` (an "until your next turn" grant stamped `Number + 2` plus a re-stamp trigger), `batch25_helpers.go:78`, `starting_town.go:23-33`, `teferi_time_raveler.go:51`, `alternative_cost.go:623` (warp's `notBefore`), `exile_play.go`, `cascade.go`, `delayed.go:137`, `protocol/reveal_frame.go:213` and the aiseat heuristic and prompt. ADR 0022 §2 says a turn number is "correct under extra turns". That holds only if the number counts turns.
- There is a second hand-written rotation: `advancePastEliminatedLocked` (`mutations.go:4943-4986`). It doesn't go through `advanceCursorLocked`, so when the active player leaves, the next turn starts without the per-turn resets. Today that affects only concede and elimination. With extra turns it becomes a third way the "new turn" hook can be missed.
[...]

PR #1144, opened 2026-09-20
feat(server): five legendaries — Dromoka, Selvala, Trouble in Pairs, Gix, Valgavoth (#1117)

krakenhavoc:
## Summary
- Five catalog cards from the Betor life-swap deck: **Dragonlord Dromoka** and **Selvala, Heart of the Wilds** at `full`; **Trouble in Pairs**, **Gix, Yawgmoth Praetor** and **Valgavoth, Terror Eater** at `caveats`, one dropped clause each and every one of them weaker than printed.
- Four additive pieces of shared vocabulary: `OpponentsCantCastDuringYourTurn`, `AnyCombinationOfColors` / `ProducedAnyCombinationOfColors`, `WardSacrificeCost.Count` + `WardSacrificeN`, `GraveyardBecomesExile.NotControlledByYou`. All four have a zero value that leaves every existing card charging and behaving exactly as before.
- A new `opponent_politics.go` holds the predicates Trouble in Pairs and Gix share — "their second draw / second spell this turn", "attacked YOU with N or more creatures", "a creature dealt combat damage to one of your opponents" — plus the seat-addressed twin of `payLifeThenDraw`.

Two details worth a reviewer's eye:

- **Dromoka's turn clause reads `g.Turn` directly**, through the existing lock-free `isActivePlayer`, not through `g.ActivePlayer()`. The latter takes an RLock and the predicate runs inside `CastSpell`, which already holds `g.mu` for write.
- **Selvala's mana ability is repeated pipe slots, not `ProducedOneColor`.** "X mana in any combination of colors" is X independent picks; `{W3|U3|…}` is one pick minting three of one colour, which is Nykthos and a materially different card. The test asserts both the produced string and the number of prompts a real activation opens.

### What was dropped, and why

| Card | Dropped clause | Why |
| --- | --- | --- |
| Trouble in Pairs | "If an opponent would begin an extra turn, that player skips that turn instead." | Inert rather than unimplemented: there is no replacement event for beginning a turn and no card in the engine can take an extra turn (#753). |
| Gix | `{4}{B}{B}{B}, Discard X cards: …` | The COST is a variable count. `AbilityCost.DiscardCards` is fixed and `DemandsX()` derives "prompts for X" from the mana string, which here is fixed — nothing would ever ask. Same seam as Ruthless Technomancer's "Sacrifice X artifacts". |
| Valgavoth | "During your turn, you may play cards exiled with Valgavoth." | A `ReplacementEffect` has no post-move hook to register a permission from, and `PermissionFilter` has no "exiled with this source" member. |

Caveat sentences as published:

- Trouble in Pairs — "The clause that stops an opponent taking an extra turn does nothing yet, because no card in the game can take one."
- Gix — "The last ability — \"{4}{B}{B}{B}, Discard X cards: Exile the top X cards of target opponent's library\" — isn't implemented. The combat-damage draw works."
- Valgavoth — "Cards Valgavoth exiles stay exiled — you can't play them."

## Sprint

## Issues
Relates to #1117

## Test plan
- [x] `go test ./internal/cards/effects/ -run 'TestDragonlordDromoka|TestSelvala|TestTroubleInPairs|TestGix|TestValgavoth|TestWardSacrificeCount' -v` — 14 tests, all pass. Dromoka refused an opponent's cast on her controller's turn, allowed her own controller's and allowed an opponent's on somebody else's turn, and resolved through a Counterspell that itself resolved and did nothing; Selvala produced five independent colour picks for a 5/5 and offered the draw to the entering creature's controller rather than her own, and declined a tie; Trouble in Pairs drew on the second draw and second spell but not the first or third, on a three-creature attack at its controller but not a one-creature one and not an attack aimed elsewhere; Gix's prompt went to the damaging creature's controller with `LifeCost: 1` declared and charged exactly that; Valgavoth exiled an opponent's dying creature, left a creature its controller had stolen in the graveyard, demanded a 3..3 sacrifice pick that offered no land, and countered the spell with no prompt when the caster held only two nonland permanents.
- [x] `go test ./internal/cards/... ./internal/game/... ./internal/legal/... ./internal/protocol/...` — all green **except** `TestRoadmapCensusIsCurrent`, which fails only because the census block in `docs/decklists/card-coverage-roadmap.md` has not been regenerated. Per the batch brief the roadmap census is regenerated once, on the rebased tip, by a final PR; nothing else in `internal/cards/coverage` fails, including the clone gate and the caveat-tone checks.
- [x] `gofmt -l internal/` clean; `go vet ./internal/cards/... ./internal/game/...` clean.


krakenhavoc:

Issue #952, opened 2026-09-18
No library→hand move for a named subset of revealed cards — four batch-41 cards are misfiled under the closed #744


krakenhavoc:
Found while writing roadmap batch 41 (#448). This is a **triage correction plus a missing primitive**, not a bug.

## The misfiling

**Horn of the Mark, Ao the Dawn Sky, Chaos Wand** and **Sandstone Oracle** are all filed on their batch issues under **#744** — "put the rest on the bottom in a random order". That is not what blocks them, and #744 is closed: `PutRestOnBottomInRandomOrder` already exists and works.

## What actually blocks them

There is **no library→hand move for a named set of revealed cards**. The search API shuffles and prompts over the whole library, which is a different operation: these cards reveal a bounded number of cards off the top, the player takes a subset, and the remainder goes to the bottom.

Goblin Ringleader was writable in batch 41 only because `BounceToHand` happens to route from any zone — which is luck, not a designed path, and it is worth not relying on it again.

## Why it matters beyond four cards

"Look at the top N, take a subset, do something with the rest" is one of the most common templates in the format — every Impulse, every Ringleader, every Commune. The roadmap's own ranking has a row for it: **"Library-top placement and ordered look (Brainstorm / tutors / Ponder)"**, 45 / 31 / 37 sole unlocks across the three 2000-card passes and 98 / 91 / 108 cards touched. It has **no tracking issue**, which is probably why these four ended up filed against an unrelated closed one.

## Ask

1. A "look at N, take a subset to hand, route the remainder" primitive.
2. Re-file the four cards above off #744 and onto it.
3. Consider making this the tracker for the library-top row of the roadmap's ranking tables, which currently reads "—".

Filed from the batch-41 card-coverage work (#949).

irobinson010:
Claiming this together with #951 on branch `fix/951-952-pay-unless-halts-stack-and-named-subset-to-hand` (base `develop`). Engine mechanics first, cards as proof.

irobinson010:
Both halves done in #1000 (base `develop`) — the primitive, and the triage correction, which turned out to be the more interesting half.

## The primitive

[...]

Issue #893, opened 2026-09-18
MillToZoneForEffect counts a leg a replacement sent elsewhere, so Oona's "exiled this way" over-counts (the mill twin of #870)


irobinson010:
Noticed by the #870/#877 agent (PR #878). `server/internal/game/effect_api.go:895` `MillToZoneForEffect` counts every non-paused leg as having reached `dest`, even when a replacement sent it somewhere else (a commander to the command zone via CR 903.9, a "if a card would be put into your graveyard from anywhere, exile it instead" effect); Oona, Queen of the Fae reads that list as "exiled this way" to make Faeries.

**Fix shape.** The same treatment #869/#878 gave destroy/exile/bounce: route the mill through `routeAllThenLocked` with a `millRoute` template (or the library→dest route it already uses) and hand the caller the landed list from `landedInZoneLocked`; convert Oona and any other reader of the mill result (grep "milled this way", "put into your graveyard this way", "exiled this way" after a mill call). Tests: Oona milling a commander that goes to the command zone → one fewer Faerie; a graveyard-replacement sending a card to exile instead → not counted as milled to the graveyard.

irobinson010:
Claimed (2026-09-18, S39): an agent has this together with #894 on branch `fix/893-894-mill-and-living-death` off `develop` (one PR, separate commits). Please don't pick it up in parallel.

irobinson010:
PR: https://github.com/krakenhavoc/cmd_and_ctrl/pull/909 — the mill routes through the one batch body with a `millRoute` template, `MillToZoneThenForEffect` hands callers the landed list (CR 400.7), and Oona, Helm of Obedience, Sphinx's Tutelage, Dread Summons and Oblivion Sower are converted.

irobinson010:
Done in #909, merged to `develop`: `millRoute` is the fourth template on `routeAllThenLocked`; `MillToZoneThenForEffect` hands the landed list forward, the fire-and-forget forms return the landed slice on the same body; Oona, Helm of Obedience, Sphinx's Tutelage, Dread Summons and Oblivion Sower converted. ADR 0013 §5l.

irobinson010:
Follow-ups filed: #910 (a sacrifice batch with a continuation) and #911 (three graveyard-exile read-backs that pay out on a cancelled leg).

PR #1145, opened 2026-09-20
feat(server): Hatred, Ripples of Undeath, Darkstar Augur + the offspring keyword

krakenhavoc:
## Summary
- Three catalog cards for the Betor life-swap deck, all `CompletenessFull`, no caveats and nothing dropped: **Hatred**, **Ripples of Undeath**, **Darkstar Augur**.
- **Offspring is now a general mechanic**, not a card-local trick: `offspring.go` carries `Offspring(mana)` (the optional additional cost, ADR 0073, keyed `"offspring"`) and `OffspringToken(cardName)` (the entry trigger, whose CR 603.4 intervening if reads the paid record off the permanent's own `Provenance`). The next offspring creature is two lines.
- `graveyard_to_hand.go` extracts `ReturnPickedToHand`, the choose-cards continuation Skullwinder and Ripples of Undeath now share — the clone gate caught the duplicate and this is the fix it asks for, not a baseline entry.

### Card notes

**Hatred** ({3}{B}{B} Instant) — Toxic Deluge's single-target sibling. The same `PayXLifeCost()` additional cost read back with `ctx.X()`, aimed through `BoostUntilEOT{Power: x}`. The life leaves at CAST, so it is gone before the spell resolves and is not refunded when the spell is countered; CR 119.4 caps X at the caster's life total at announce.

**Ripples of Undeath** ({1}{B} Enchantment) — built whole, including the optional payoff. `MillToZone` with a continuation supplies "those cards" as what actually *reached* the graveyard (#893), `MayPay` carries the `{1}`, and the 3 life is paid inside `OnPay` through the cost path before the card moves. The engine's may-pay prompt carries a parsed MANA cost and nothing else — a `ParsedCost` has no life component — so the cost is split, and the split is kept honest by not making the offer below 3 life (CR 119.4: a payment you cannot afford is no payment). A pending choice holds the table, so nothing can move the life total between the check and the payment. The prompt's question names both halves.

**Darkstar Augur** ({2}{B} 2/3 Bat Warlock) — Dark Confidant's upkeep flip on a flying body, plus Offspring {B}. The flip goes through the named library-to-hand door with a continuation (#952) rather than a `BounceToHand`, and reads the mana value before the move because that is the last moment the card is guaranteed findable. Dark Confidant itself is left alone: its own body is pinned by a line-keyed entry in `exitPayoutAllowlist`.

An offspring token carries no `Provenance`, so the intervening if is false for it and it cannot make a token of its own. That is asserted by the test, not assumed from the code.

## Sprint

## Issues
Relates to #1117

## Test plan
- `go test ./internal/cards/effects/ -run 'TestHatred|TestRipplesOfUndeath|TestDarkstarAugur'` — **pass** (8 tests / 12 subtests):
  - Hatred at X=0 and X=5: life leaves at cast, effective power is 2+X, toughness untouched, back to 2 after the turn ends; and an unpayable X is a refused cast that leaves the card in hand.
  - Ripples: mills three at your own precombat main and zero at an opponent's; paying `{1}` + 3 life offers exactly the three cards milled this way (a card already in the graveyard is not offered) and puts the chosen one in hand; a player at 2 life still mills and is not offered the payment.
  - Darkstar Augur: offspring paid → a second, 1/1, token, flying, copy — and exactly two Augurs, which is the no-recursion assertion; unpaid → one. Printed flying on the card itself. Upkeep flip puts the revealed card in hand and loses its mana value, 0 for a land and 3 for a `{2}{B}`.
- `go test ./...` (whole server tree) — **pass** except `TestRoadmapCensusIsCurrent`, which is the census block this PR deliberately does not regenerate (parallel batches all conflict on it; one final PR rebuilds it on the rebased tip).
- `go test ./internal/cards/coverage/` — clone gate and caveat/coverage guards **pass**.
- `gofmt -l internal/cards/effects/ internal/game/` — clean.
- `golangci-lint v1.64.8 run ./internal/cards/...` — clean.


krakenhavoc:

PR #1137, opened 2026-09-19
feat(engine): transform a permanent — the verb, and why it is two verbs

krakenhavoc:
## Summary

- **Closes the largest open engine seam.** `Game.TransformPermanentForEffect` turns a battlefield permanent over **in place** (CR 701.27a) and never goes near `MoveCard`. CR 712.18 is the whole contract — the permanent doesn't become a new object — so `InstanceID`, `ObjectEpoch`, counters, marked damage, tap state, attachments, the CR 613.7 timestamp, summoning sickness, the controller and every `Game`-side per-object registry survive, and no `EventZoneMove` / `EventETB` / `EventLTB` is emitted. A back face's "when this enters" trigger does **not** fire off a transform, because nothing entered.
- **`Game.ExileAndReturnTransformedForEffect` is the other shape** — "exile it, then return it to the battlefield transformed" (CR 712.14a): two zone changes, CR 400.7 in full, a genuinely new object. Two functions rather than one with a flag, because conflating them gives Fable of the Mirror-Breaker a Saga that kept its lore counters and Storm the Vault a permanent that re-triggers its own ETBs — wrong in opposite directions.
- **#1128 (ADR 0079) is merged**, so this branch is rebased onto `develop` and the diff is the code alone. The two proof cards from #1112 follow in #1147.

## Sprint

S46 — Permanents that change what they are

## Issues

Relates to #343, relates to #1112. Design: #1128 (ADR 0079, merged).

## What is here

| | |
|---|---|
| `server/internal/game/transform.go` | `CanTransform`, `TransformPermanentForEffect`, `ExileAndReturnTransformedForEffect` |
| `server/internal/game/transform_test.go` | 13 tests, written so that an implementation-as-a-blink fails them |
| `server/internal/game/events.go` | `EventTransform` |
| `server/internal/game/layer_listener.go` | the invalidation arm |
| `server/internal/game/cost_modifier.go` | CR 712.8e — `manaCostForValue` |
| `server/internal/protocol/log.go` | `LogTransform` + its projection and rendering arms |
| `server/internal/cards/effects/transform.go` | `Transform{Target}`, `TransformThis{}`, `ExileAndReturnTransformed{…}`, `TransformedPermanent` |
| `docs/engine-seams.md` | the row moves Open → Closed |
| `docs/protocol.md` | the `transform` log kind |

[...]

Issue #521, opened 2026-09-14
S33 sub-PR 5: tokens get catalog keys — a Treasure must stop blocking every restore point


krakenhavoc:
Sub-PR 5 of [#515](https://github.com/krakenhavoc/cmd_and_ctrl/issues/515). **Depends on:** nothing — independent of the reconnect chain, can land in parallel. Engine.

## The problem

ADR 0041's restore-point rule is that a snapshot holding un-rebuildable Go closures is not written, so what sits on disk is always a state that can be rebuilt exactly. Correct. The ADR also names the cost honestly:

> The rewind distance is **not** always short. A live token on the battlefield — a Treasure — makes every subsequent state uncleanable until it is spent. That is the strongest argument for prioritising phase 3.

This is not a rare edge. Treasure, Food, Clue, Blood, Gold and Eldrazi Spawn are ordinary output of the current catalog, and they sit on the battlefield for turns. **While one is out, no restore point is written at all**, so a deploy rewinds the table to whatever state preceded it — potentially many turns, potentially the start of the game.

`server/internal/game/snapshot.go:721`:

```go
if (len(c.ManaAbilities) > 0 || len(c.ActivatedAbilities) > 0) && c.OracleID == "" {
    cen.IntrinsicAbilityCards++
    cen.note("intrinsic abilities with no oracle id: %s", labelOr(c.Name, c.InstanceID.String()))
}
```

The check is on **whether there is an oracle ID to look up**, not on whether anything unserialisable is actually present. And `TreasureToken()` (`cards/effects/tokens.go:143`) is:

```go
ManaAbilities: []game.ManaAbilityShape{{
    TapCost:       true,
    SacrificeCost: true,
    Produced:      "{W|U|B|R|G}",
    Label:         "{T}, Sacrifice: Add one mana of any color",
}},
```

Every field set is pure data. `ManaAbilityShape`'s closure fields (`Condition`, `ProducedFunc`, `RestrictionsFunc`, `Rider`) are all nil. A Treasure is blocking every restore point in the game on account of closures it does not have.

`ActivatedAbilityShape.Effect` (`activated.go:142`) *is* always a closure, so Food and Clue need the catalog route rather than carrying data — which is the same route printed cards and token *copies* already take.

## Scope

Give token templates a **stable catalog key** and register them, so `CatalogManaAbilities` / `CatalogActivatedAbilities` re-derive a token's abilities the same way they already do for a printed card (`snapshot.go:1181`).

[...]

Issue #758, opened 2026-09-17
ADR 0020 addendum then implementation: "tap N untapped [permanents] you control" as an ability cost — Springleaf Drum, Earthcraft, Opposition, Azami, Heritage Druid


krakenhavoc:
## Why

No ability cost can tap a permanent other than the source. Checked on develop (`ad9413f`):

- **`AbilityCost`** (`server/internal/game/activated.go:35`) has `Tap` (the source only, `:38`), `SacrificeSelf`, `SacrificeOther` (`:49`), `Mana`, `Life`, `Loyalty`, `Crew` (`:116`), `RemoveCounters` (`:146`) and `MinX` (`:163`). Nothing taps a counted set of matching permanents.
- **`ManaAbilityShape`** (`game/effect_hooks.go:129`) has `TapCost` (the source, `:130`), `SacrificeCost`, `SacrificeOther` (`:147`), `LifeCost`, `ManaCost` and `Condition` (`:197`). No tap-another field either.
- **Crew is the model to copy, but it is keyed on power.** `validateCrewCostLocked` (`activated.go:604`) takes `ActivateAbilityParams.CrewIDs` (`:285`), checks no summoning sickness (`:589`) and taps at payment. It is wired end to end: `crew_ids` in `actions/actions.go:921`, `crewPayment` in `internal/legal/abilities.go:305`, and `client/src/lib/crew.ts` with `CrewCostModal.svelte`.
- **`TapPermanentsCost`** (`game/tap_cost.go:55`, `CastSpellParams.TapIDs` at `mutations.go:309`) is for spells only (convoke, waterbend). It reduces a mana demand; it isn't a standalone cost.
- **Two catalog cards ship without the ability**, as declared simplifications: The Shire (the Food ability, `cards/effects/the_shire.go:19-25`) and Scene of the Crime (the any-colour mana ability, `scene_of_the_crime.go:21-23`). Survivors' Encampment is excluded from batch 24 for the same reason (`batch24_test.go:140`).
- **Precedent for the shape of the work:** #625 added `RemoveCounters` as an addendum to ADR 0020 (`docs/decisions/0020-activated-abilities.md:114`) and touched the engine, actions, legal enumerator, heuristic, protocol and a client picker (`CounterCostModal.svelte`).

[...]

PR #1147, opened 2026-09-20
feat(cards): Storm the Vault and Fable of the Mirror-Breaker

krakenhavoc:
## Summary

- The two #1112 cards that prove ADR 0079's **two** verbs — one card for each, chosen so that an implementation that used the wrong one fails a test rather than shipping.
- **Storm the Vault // Vault of Catlacan** — `full`. A Legendary Enchantment becomes a Legendary Land **without going anywhere**.
- **Fable of the Mirror-Breaker // Reflection of Kiki-Jiki** — `caveats`. Chapter III is an exile and a return, so what comes back is a new object.
- **Stacked on #1137** (the verb). #1128 (ADR 0079) is merged, so this branch is rebased onto `develop` and only the verb commit sits underneath the card ones.

## Sprint

S46 — Permanents that change what they are

## Issues

Relates to #343, relates to #1112. Design: #1128 (ADR 0079, merged). Engine: #1137.

## Storm the Vault // Vault of Catlacan — `full`

> Whenever one or more creatures you control deal combat damage to a player, create a Treasure token.
> At the beginning of your end step, if you control five or more artifacts, transform Storm the Vault.
> // (Transforms from Storm the Vault.) {T}: Add one mana of any color. {T}: Add {U} for each artifact you control.

Three things this card was picked for:

1. **The back face's mana abilities are live the instant the face flips**, and nobody wires them up. `game.CatalogKey` starts answering `"<oracle_id>#1"` and `vault_of_catlacan.go` is simply there. `TestVaultOfCatlacanManaAbilitiesAreLiveAfterTheTransform` asserts the front face has **zero** mana abilities and the back face has two, across one transform — which is the whole payoff of the `#N` key and the reason the verb was this small.
2. **The intervening if (CR 603.4) is checked twice.** One test says four artifacts never puts the ability on the stack; another says five artifacts that become four *in response* leaves it on the stack and does nothing. An implementation that gated only `AppliesTo` passes the first and fails the second.
3. **"One or more creatures"** rides `WheneverOneOrMoreCreaturesYouControlDealCombatDamageToAPlayer`, which is `OncePerBatchPerPlayer`. Written per-creature it would make one Treasure per attacker, which is *stronger* than printed.

"{T}: Add {U} for each artifact you control" is `ProducedPerPermanent("U", MatchArtifact)` — a count read at activation, the shape Cabal Coffers uses.

## Fable of the Mirror-Breaker // Reflection of Kiki-Jiki — `caveats`

> I — Create a 2/2 red Goblin Shaman creature token with "Whenever this creature attacks, create a Treasure token."
> II — You may discard up to two cards. If you do, draw that many cards.
> III — Exile this Saga, then return it to the battlefield transformed under your control.
> // {1}, {T}: Create a token that's a copy of another target nonlegendary creature you control, except it has haste. Sacrifice it at the beginning of the next end step.

**Chapter III is the second verb, not the first.** `TestFableChapterThreeExilesAndReturnsItTransformed` asserts a new `InstanceID`, zero lore counters, a fresh CR 613.7 timestamp and summoning sickness — every one of which an in-place transform would keep. It also asserts the card is **not** in the graveyard, because the obvious worry is CR 714.4 sacrificing a finished Saga out from under the return. It does not, and needs no special case: by the time the SBA next runs there is no Saga on the battlefield, because `IsSaga` reads the effective subtypes and `SagaFinalChapter` derives from the back face's (absent) chapter triggers.

Chapter II reuses `b39MayDiscardThenDraw(2, false, …)` — Cathartic Pyre's clause, byte for byte the same instruction.

The back face's ability is `CreateTokenCopy` with a haste `Except` plus a CR 603.7 `ScheduleDelayedTrigger`. **"Sacrifice it" is the token**, and the test pins that by checking the Reflection *and* the copied creature both survive the end step — a misreading that sacrificed the source would empty the board instead.

### The caveat, and why it is not fixed here

> "The Goblin Shaman token is a plain 2/2 — its \"whenever this creature attacks, create a Treasure token\" ability is not implemented."

A non-copy token has no oracle ID, so `game.CatalogKey` answers `""` and `TriggersForCard` returns nil. That is the open **#521** seam ("Triggered and static abilities on non-copy tokens"), appended to that row in `docs/engine-seams.md`. The `GrantedAbilities` composite key would technically carry a trigger onto a minted token, but nothing in the tree exercises that path and inventing its first use inside a transform PR is how a seam gets closed by accident and half-tested.

Weaker than printed, with an honest caveat — the only direction allowed (#259). `TestFableChapterOneMakesTheGoblinShaman` asserts the token has **no** triggered abilities and says in its failure message that #521 landing makes this caveat stale, so the day it closes the test says so.

## Why the tests import real cards

`castCatalogSpell` builds a flat fixture with no `Faces` at all. A transform card's whole point is `Faces[1]`, which lives on Scryfall's `card_faces[1]` and reaches `game.Card` only through `deck.ToGameCard` — so a test written on a fixture would exercise a `CanTransform` that always answers false and **pass by doing nothing**. Every card here goes down the production import road, the same reasoning `siege_transform_test.go` gives.

One thing that fell out of writing them: a Saga has to be **cast** rather than put onto the battlefield directly. With no spell on the stack the first priority pass advances the *step* instead of settling chapter I's trigger, which adds a second lore counter and leaves two chapters racing for one trigger-order prompt. `importAndCast` says so in its doc comment.

## Also here

- `tokens_table.go` gains one row, `"2/2 red Goblin Shaman"` (append-only, as the file requires).
- `docs/engine-seams.md`: the Fable token appended to the #521 row.
- `docs/decklists/card-coverage-roadmap.md`: regenerated with `go test ./internal/cards/coverage/ -update`, which is the only way that block may change. 2080 → 2084 registry keys, 2014 → 2016 whole cards, 66 → 68 back faces (regenerated again after the rebase, against develop's live catalog).
- `docs/decklists/pirates-mary-read-anne-bonny.md`: two stale notes cleared. The Saga row (done since S27, and now done for the transforming half — Brass's Tunnel-Grinder still waits, on **discover** and **descend**, not on Sagas), and "not in the local Scryfall snapshot", which was wrong when written: all three cards are in the dump, and what was missing was the `card_faces` path.

## Test plan

- [x] `go test ./internal/cards/effects/ -run 'StormTheVault|VaultOfCatlacan|Fable|Reflection' -count=1 -v`:
  ```
  --- PASS: TestStormTheVaultTransformsAtTheEndStep (0.00s)
  --- PASS: TestStormTheVaultWithFourArtifactsDoesNotTrigger (0.00s)
  --- PASS: TestStormTheVaultLosesTheArtifactsInResponse (0.00s)
  --- PASS: TestVaultOfCatlacanManaAbilitiesAreLiveAfterTheTransform (0.00s)
  --- PASS: TestFableChapterOneMakesTheGoblinShaman (0.00s)
  --- PASS: TestFableChapterThreeExilesAndReturnsItTransformed (0.00s)
  --- PASS: TestReflectionOfKikiJikiCopiesWithHaste (0.00s)
  ok  	github.com/krakenhavoc/cmd_and_ctrl/server/internal/cards/effects	0.020s
  ```
- [x] `go test ./internal/cards/... ./internal/catalog/... -count=1` — exit 0. In particular `TestNoNewExactClonesInTheCatalog` passes with **no** baseline edit (the one helper that would have been a clone, `advanceToEndStepOf`, was deleted and `flicker_test.go`'s called instead), and the coverage census passes against the regenerated roadmap.
- [x] `go test ./... -count=1` over the whole server module — exit 0, no failures.
- [x] `gofmt -l .` under `server/` — clean.
- [x] `golangci-lint run ./...` (v1.64.8 on `golang:1.24`, `GOTOOLCHAIN=auto`) — clean, over all three commits.
- [ ] Manual smoke-test on a dev stack. Not run: both cards need a real decklist import to get their faces, which the `CMDCTRL_DEV_SKIP_DECK_VALIDATION` small-deck loop does not set up cheaply. The import path itself is what the tests exercise end to end.


krakenhavoc:
Added a fourth commit (23ba5133) applying AGENTS.md §7's stale-caveat rule to the card this whole chain's issue was filed about.

`aang_swift_savior.go` said "Waterbend {8}: Transform Aang" was blocked on a transform verb the protocol did not have. After #1137 it has one, so that comment was three days from sending somebody off to build something that already exists.

The **caveat** is unchanged and still true — Aang still can't transform — because the blocker moved rather than lifting: waterbend on an **activated** ability has no cost component (`Spec.TapCost` carries convoke and waterbend for a *spell*; `AbilityCost` has no tap-other component, which is #758). Writing the ability with the cost omitted would make it a free transform, which is #259's mistake. And the back face's attack trigger is a consequence of that rather than its own gap: face 1 is reachable by the verb now, but nothing on *this* card can reach it, so a `"#1"` entry would be a trigger no game could ever fire.

I grepped the rest of the catalog for caveats the verb invalidates (`grep -ril transform server/internal/cards/effects/`). The only other family is the four Sieges, whose shared `SiegeTransformedCastCaveat` is about the *timing of the free cast* after a defeat — S32 machinery this chain does not touch — so it stays.

krakenhavoc:
**Rebased onto `develop`, and the ADR number moved.** Force-pushed; the diff is unchanged in substance.

`#1128` was written and merged as **ADR 0078**, and the token-art ADR (#1124) took that number on `develop`. The merge renumbered this one to **ADR 0079** (`docs/decisions/0079-transforming-a-permanent.md`), which is the collision AGENTS.md §4 exists to catch and `TestADRNumbersAreUniqueAndMatchTheirHeading` would have caught on the next CI run either way. Every inbound reference here — the engine comments, the card files, `docs/engine-seams.md`, `docs/protocol.md` and the commit messages — now says 0079.

Three merge resolutions worth naming, since `develop` moved under both branches:

- `server/internal/protocol/log.go` and `docs/protocol.md` conflicted with **`LogSettings`** (ADR 0075 sub-PR 3), which landed in the same two switches. Both sides kept; the `kind` list in `protocol.md` now reads `…, class_level, settings, transform`.
- `docs/decklists/card-coverage-roadmap.md` conflicted on the census numbers. Resolved by taking `develop`'s and **regenerating** with `go test ./internal/cards/coverage/ -update` — the only sanctioned way that block changes — so the counts are measured against the live catalog rather than against the ones this branch was cut from.
- `docs/engine-seams.md` merged cleanly.

Re-verified after the rebase: `go test ./... -count=1` over the whole server module, exit 0, no failures — including `docsguard`, the coverage census, the clone gate and the log event-kind gate.

Issue #1032, opened 2026-09-19
Table settings and host controls: undo rules, life, commander damage, bot pace, and announced spawning (ADR 0075)


krakenhavoc:
## Why

A table has no owner and no settings surface. `set_undo_limit` can be sent by any seated player and has no UI, `Game.Start` silently turns an undo limit of 0 into 1, starting life and commander damage are constants, and spawning (needed for tokens and cards the engine can't create yet) is dev-only and not announced.

Design: **ADR 0075 — Table settings and host controls** (`docs/decisions/0075-table-settings-and-host-controls.md`).

**Decisions (2026-09-19, owner):** host + server admin control the table; production spawning is a table setting, off by default, and every spawn is announced in the game log; v1 settings are undo rules, starting life, commander damage, bot pace (plus the spawn switch); settings can change any time, and every change is logged.

## Sub-PRs

- [x] 1. (#1042) Engine settings: `game.TableSettings`, `UpdateSettings`, `EventSettingsChanged`, starting-life / commander-damage readers, `UndoUnlimited`, snapshot schema + migration, drop the `<= 0` → default rewrite
- [x] 2. (#1043) Host: `GameMeta.HostPlayerID`, `host_discord_id` on create, first-human fallback, transfer + auto-pass, `CanManageTable`, `is_host` on the wire, `/cc-invite` passes the invoker
- [x] 3. (#1110) Settings surfaces: `PATCH /games/{id}/settings`, `set_table_settings`, `set_undo_limit` gated to host-or-admin; protocol.md + lobby.md; log narration
- [x] 4. (#1111) Production spawn: `POST /games/{id}/spawn` (host-or-admin + `AllowSpawn`), token templates, `EventSpawned` + log line, free undo entry
- [x] 5. (#1150) Client: settings panel (lobby + in game), host crown, spawner with Tokens tab, "spawning on" badge, ∞ undo
- [x] 6. (#1104) Bot pace presets in `aiseat`; docs/bot.md

1 and 2 are independent; 3 needs both; 4 needs 2 and 1's `AllowSpawn`; 5 and 6 last.

Relates to #614 (settles its "host or admin" design point), #607 (S34 `created_by` seeds the host later), #734.

Sprint: S35 — Playtest stabilisation, round 2

krakenhavoc:
Claimed (krakenhavoc). ADR 0075 is up for review in the PR linked above; implementation sub-PRs start once it's accepted.

[...]

PR #1150, opened 2026-09-20
feat(client): table settings, host crown, the spawner's Tokens tab and the spawning badge (#1032)

krakenhavoc:
## Summary

- **Table settings panel** (ADR 0075 §2.5), one component with two hosts: the lobby page (`PATCH /games/{id}/settings`) and an entry in the in-game ⋯ menu (`set_table_settings` over the socket). The host and the admin edit it; everyone else gets the same panel with the controls disabled, because the settings are public on purpose. Every field carries a sentence in plain words, starting life disables itself once the game is active, and the undo row reads `-1` out as "Unlimited take-backs" rather than as a count.
- **Host crown** on the host's seat at the table, read off `PlayerView.is_host` — a chip beside the name, deliberately not a second crown in the marker column where the glyph already means *monarch*.
- **Spawner** moves out of `components/dev/` and grows a `managed` mode: the production routes, a **Tokens** tab (the reason the feature exists — a token has no Scryfall printing), and the battlefield-only rule that comes with tokens. The dev path is byte-for-byte unchanged in behaviour.
- **"Spawning on" badge** in the table header for every viewer while the switch is on, plus a Spawn menu entry that needs *both* of the server's gates.

No engine, lobby or `aiseat` change: every wire shape this uses shipped in sub-PRs 1–4.

## Sprint

S35 — Playtest stabilisation, round 2

## Issues

Closes #1032 (this is the last of ADR 0075's six sub-PRs; 1, 2, 3, 4 and 6 are merged).

## Decisions the ADR left open

- **No Save button.** Every field is independent and the server takes a partial, so a change commits on the spot. Batching five changes into one submit would produce one log line that names none of them — and the log line is the feature.
- **The panel is shown to people who cannot change it**, greyed, rather than hidden. `apply()` refuses as well, because `disabled` is an affordance and not a guard.
- **The stop-gap "Undo limit" row** sub-PR 3 left in the Sandbox menu is removed — the panel owns that setting now. One setting, one control.
- **The Spawn entry is absent, not disabled,** when the table's switch is off; the switch itself is one entry above it. The *badge*, by contrast, is gated on nothing but the switch, because it exists for the opponents.

## Known gap (server work, not in this PR)

In the **lobby** the panel is offered only to the host and the admin. The lobby's HTTP surface has no `GET` for settings — the read here is an empty `PATCH`, which inherits the write's `CanManageTable` gate. At the table, where `GameView.settings` is public, everyone does see the panel read-only, so the ADR's promise is kept where it matters most. A `GET /games/{id}/settings` would close the gap. Related: a **spectator** has no ⋯ menu at all (pre-existing gating on `viewerID`), so they reach the badge but not the panel.

## Test plan

- [x] `npm run check` — 0 errors, 0 warnings, 536 files
- [x] `npm run lint` — eslint + prettier clean
- [x] `npx vitest run` — 95 files, 1396 tests, all passing (32 new unit + panel-render tests, 4 host-chip render tests, 8 spawner-mode render tests)
- [x] `npm run build` — clean production build
- [x] Unit-tested: unlimited undo is a value and not an error; `"ended"` locks starting life exactly as `"active"` does; a patch carries only the field that moved (a false or a 0 included); the spawn entry needs both gates while the badge needs neither; a token is offered the battlefield and nothing else
- [x] Render-tested: a non-manager sees every control disabled and emits no patch; starting life is disabled mid-game with its reason; ∞ renders for `-1` and "Undo is off" for `0`; the dev spawner has no Tokens tab and keeps all six zones
- [ ] Manual: host + non-host at a live table, toggle each setting, watch the log line and the badge; spawn a Treasure and undo it


krakenhavoc:
CI note: the red check on the first run is **not from this PR**, which touches no Go at all.

`TestRoadmapCensusIsCurrent` fails because `docs/decklists/card-coverage-roadmap.md` still says **2089** registry keys while the live catalog is at **2094** — the gap opened when [#1119](https://github.com/krakenhavoc/cmd_and_ctrl/pull/1119) (seven Equipment cards) merged to `develop` without regenerating the census block. `develop`'s own run for `be295c22` fails the same way.

The remedy is the one the failure prints, and it belongs to a card PR rather than to a client one, so it is deliberately not folded in here:

```
cd server && go test ./internal/cards/coverage/ -update
```

Client-side, on this branch: `npm run check` 0 errors / 0 warnings, `npm run lint` clean, `npx vitest run` 95 files / 1396 tests green, `npm run build` clean.

krakenhavoc:
Confirmed: `develop` at `be295c22` is **red on its own run** with the same `TestRoadmapCensusIsCurrent` failure — https://github.com/krakenhavoc/cmd_and_ctrl/actions/runs/35478737161. Every PR branched off it inherits the failure until the census block is regenerated.

krakenhavoc:
For anyone reading the two earlier comments on this PR: they describe a CI failure that no longer reproduces. `TestRoadmapCensusIsCurrent` was red on `develop` itself (a card batch landed with a stale generated census block); #1151 regenerated it, and since PR CI builds the branch merged with `develop`, the rerun went green without a change here. This PR was merged on that green run.

Issue #1221, opened 2026-09-22
Abilities and statics that function from the graveyard, hand or exile (CR 113.6, CR 702.82 unearth, CR 702.96 scavenge, CR 702.128 embalm, CR 702.49 ninjutsu)


irobinson010:
Two rows on `docs/engine-seams.md` are the same seam seen twice: **CR 113.6** — an ability of a card that is not on the battlefield functions only if it says it does. #660 / ADR 0062 Decision 1 closed the HAND half for activated abilities (`ActivatedAbilityShape.Zones`, cycling and typecycling); #922 gave `TriggeredAbility` the same field. What is still open is every other zone, and the statics half has no field at all.

### The rows

**"Ability activatable from a non-battlefield zone"** (4 cards waiting, tracked #655). The shape exists; nothing declares a non-hand zone and no consumer walks one. The keyword family behind it:

- unearth (CR 702.82) — Hellspark Elemental, Sedraxis Specter
- scavenge (CR 702.96) — exile the card as a cost, put counters
- embalm / eternalize (CR 702.128 / CR 702.129) — exile it, make a token copy with changes
- ninjutsu (CR 702.49) — from the HAND, returning an unblocked attacker

Cards on the row today: Simian Spirit Guide (#296); Elvish Spirit Guide (#299); Reassembling Skeleton (#299); Drownyard Temple (#311).

**"Statics read from a non-battlefield zone"** (3 cards waiting, untracked). The layer pass gathers statics from `g.Battlefield` only (`Game.activeStaticAbilitiesLocked`), so a static printed "as long as this card is in your graveyard" is never applied:

- Anger (#295), Wonder (#300), Brawn (#310) — the Judgment incarnations, plus Valor
- Yixlid Jailer-style "cards in graveyards lose all abilities"
- "as long as this card is in your hand" statics

### Scope

1. **Activated abilities with a zone.** `ActivatedAbilityShape.Zones` read by `ActivatedAbilitiesForCard` through `Designation.Active`'s gate; activated through `ActivateCatalogAbility` with the source in that zone; enumerated by `internal/legal/abilities.go` per zone (the sibling of #1014's five-zone cast walk); stamped by `protocol/view.go` for the zone's OWNER (the per-viewer split of #1167 / #1174, `cant_activate`); offered by the client's zone browser. Cycling's hand path unifies onto the one field rather than growing a second.
[...]

Issue #1228, opened 2026-09-22
A mana ability that functions from the hand (CR 113.6): ManaAbilityShape.Zones and an exile-this cost — the Spirit Guides


irobinson010:
Noticed while building #1221, which gave activated abilities their graveyard and exile zones and then could not take two of the four cards off the seam row.

`ActivatedAbilityShape.Zones []ZoneKind` (CR 113.6) has existed since #660 and `TriggeredAbility.Zones` since #922. **`ManaAbilityShape` has neither the field nor the cost component**, and the cards that want it are not exotic:

```
Simian Spirit Guide   Exile this card from your hand: Add {R}.
Elvish Spirit Guide   Exile this card from your hand: Add {G}.
```

Both are on the engine-seams row ["Ability activatable from a non-battlefield zone"](../blob/develop/docs/engine-seams.md), and both are still blocked after #1221, because a mana ability is a different entry point:

- `ActivateManaAbility` (not `ActivateCatalogAbility`) finds its source on the BATTLEFIELD, so there is nothing to give a zone to yet.
- `ManaAbilityShape` has no `Zones`. [ADR 0071](../blob/develop/docs/decisions/0071-designations-that-switch-abilities-on.md) Decision 1 note 3 says the same thing from the designation side: mana abilities do not get `ActiveWhen` either, "not a principle — the field plus its accessor is the same two lines on the day one does". This is that day for the zone half.
- `ManaAbilityCost` has no exile-this component. #1221 added `AbilityCost.ExileSelf` for the graveyard ("Exile this card from your graveyard" — scavenge, embalm); the Spirit Guides want the same clause reading the HAND. The same asymmetry already exists for discard: `AbilityCost.DiscardCards` shipped with #660 and `ManaAbilityCost` still has none (the seam doc's "Discard cost on a non-activated-ability surface" row, Mox Diamond / Skirge Familiar).

The work is the #1221 shape, one entry point over:

1. `ManaAbilityShape.Zones []ZoneKind`, read by `ManaAbilitiesForCard` through the same `AbilityFunctionsFromZone`-style predicate (or a mana sibling of it — the shapes differ).
2. `ManaAbilityCost.ExileSelf`, validated against the HAND and paid through the one exit primitive with `MustSettleNow`.
3. `ActivateManaAbility` finding its source in whatever zone holds it and taking the OWNER as "you" (CR 108.4), exactly as `ActivateCatalogAbility` has since #660.
[...]

Issue #1227, opened 2026-09-22
Ninjutsu (CR 702.49): a return-an-unblocked-attacker cost, and an entry that puts a card onto the battlefield attacking (CR 506.3c)


irobinson010:
Split out of #1221, which delivered the graveyard and exile half of CR 113.6 for activated abilities (unearth, scavenge, embalm / eternalize) and declared ninjutsu out of scope.

**CR 702.49a:**

> Ninjutsu [cost] means "[cost], Return an unblocked attacker you control to hand: Put this card onto the battlefield from your hand tapped and attacking."

The ZONE is not the problem. Ninjutsu functions from the HAND, and `ActivatedAbilityShape.Zones = {ZoneHand}` has reached it since #660. Its other two halves do not exist:

### 1. A cost component that returns an unblocked attacker you control

`AbilityCost` has no "return a permanent to its owner's hand" component at all — that is the seam doc's own ["Return-a-permanent-to-hand cost component"](../blob/develop/docs/engine-seams.md) row (Quirion Ranger, Master Transmuter, Wirewood Symbiote) — and ninjutsu wants the narrowest form of it: exactly one permanent, a creature you control, ATTACKING and NOT BLOCKED.

- "Is this attacker blocked" has no exported accessor today: `Game.blockedAttackers` (game.go:346) is read only by the unexported `attackerBlockedLocked` (`internal/game/blockers.go:306`). Nothing in `internal/cards/effects` can ask.
- The payment is a bounce, and a bounce can PAUSE (a commander's CR 903.9 "may"), so like every other cost it has to go through `routeCardToZoneLocked` with `MustSettleNow` rather than through `BounceToHandForEffect`.
- The returned creature's `AttackingTarget` has to be READ BEFORE the bounce: the exit clears it (`zone.go:194`) and LKI carries no combat state.

### 2. An entry that puts a card onto the battlefield ATTACKING

[...]

Issue #1210, opened 2026-09-22
Activation restrictions and watches: 'can't activate' statics, the named-source gate, and 'whenever an opponent activates an ability' (CR 602.5, CR 101.2)


irobinson010:
Three rows of `docs/engine-seams.md` have no tracker, and all three are about the
same moment: **the announcement of an activated ability** (CR 602.5, CR 101.2).
This issue is their tracker.

### 1. `"Can't be activated" activation gate` — 3 cards

> primitive: neither `ActivateCatalogAbility` nor `ActivateManaAbility` can be
> told an ability can't be activated (the Null Rod / Stony Silence shape)

Cards waiting: **Collector Ouphe** (#385), **Linvala, Keeper of Silence** (#386),
**Cursed Totem** (#386).

This is the exact twin of the row #760 closed on the *cast* side (ADR 0073 §7):
`Game.CastGateLocked` is one function with two callers and one view stamp, and
nothing on the activation side answers the same question. `restrictions.go`
carries the per-permanent `CantActivate` / `CantActivateMana` bits (Arrest,
Faith's Fetters) — a *restriction on one permanent*, put there by an Aura
attached to it. What is missing is the *board-wide* shape: a permanent saying
that some OTHER set of sources can't activate their abilities.

### 2. The named-source half of the same gate — 3 more cards

ADR 0073 closed the cast gate with one open item named in the ADR: *"a ban on a
chosen card NAME (Meddling Mage, Nevermore) wants a choose-a-card-name prompt the
engine does not have"*. The activation side has the same hole and three printed
cards in it: **Pithing Needle**, **Phyrexian Revoker**, **Sorcerous Spyglass** —
"Activated abilities of sources with the chosen name can't be activated unless
they're mana abilities", read off a name chosen as the permanent enters.

The engine has `Card.NamedTribe` (choose a creature type), `Card.ChosenColor` and
`Card.ChosenPlayer` (#1007, CR 614.12) and no `Card.ChosenName`. Building it on
#1007's pattern also makes **Meddling Mage** buildable, which ADR 0073 listed as
out of scope for exactly this reason.

### 3. `"Whenever an opponent activates an ability" watch` — 2 cards

> event kind: `triggerHarvester.OnEvent` deliberately skips `EventTrigger`, so
> nothing can watch an opponent's activation

Cards waiting: **Harsh Mentor** (#384), **Runic Armasaur** (#385).

[...]

PR #1232, opened 2026-09-22
feat(engine, protocol, client, cards): an activated ability that functions from the graveyard or exile — unearth, scavenge, embalm (CR 113.6) (#1221)

irobinson010:
Closes #1221 (the activated-ability half; the statics half of the same issue is the next PR).

CR 113.6 is one rule and the engine had answered a quarter of it. `ActivatedAbilityShape.Zones` has existed since #660 and nothing but the hand ever used it — because the shape was never the missing part. **The consumers were.** The enumerator walked the battlefield and the seat's own hand and stopped; the view stamped `hand_abilities` on a hand card and nothing anywhere else; and `AbilityCost` had no component for the clause three of the four graveyard keywords print. The seam row said the same thing from the other side: four cards "one `Spec` edit away", and had been for months, because nothing would have offered the edit to anybody.

## Shapes

| shape | where | what |
|---|---|---|
| `AbilityCost.ExileSelf bool` | `server/internal/game/activated.go:282` | scavenge's and embalm's "Exile this card from your graveyard" — `DiscardSelf`'s sibling one zone over |
| `(*Game).validateExileSelfCostLocked` | `server/internal/game/exile_cost.go:50` | the graveyard check, with the rest of the cost, before anything is paid |
| `(*Game).payAbilityExileSelfLocked` | `server/internal/game/exile_cost.go:71` | paid last, through `routeCardToZoneLocked` with `MustSettleNow` |
| `(*enumerator).abilityZones()` | `server/internal/legal/abilities.go:137` | the activation sibling of #1014's five-zone cast walk — hand, graveyard, command, exile |
| `stampZoneAbilities` / `stampZoneAbilitiesInPile` | `server/internal/protocol/view.go:3772`, `:3805` | the same four piles on the wire |
| `castStamps.ZoneAbilities` + `(*CardView).stampZoneAbilitiesFor` | `server/internal/protocol/view.go:2821`, `:3092` | the per-seat carrier, merged rather than replaced |
| `CardView.ZoneAbilities` (`zone_abilities`) | `server/internal/protocol/view.go` | was `hand_abilities`; renamed, not duplicated |
| `ActivatedAbilityView.ExileSelf` (`exile_self`) | `server/internal/protocol/view.go:2003` | `discard_self`'s advisory sibling |
| `effects.ExileThis()` | `server/internal/cards/effects/activated.go:162` | the cost constructor |
| `effects.Unearth / Scavenge / Embalm / Eternalize` | `unearth.go:42`, `scavenge.go:32`, `embalm.go:41`, `:58` | the four keywords, in the `Cycling`/`Typecycling` mould |
| `effects.ExileInsteadOfLeavingBattlefield` | `server/internal/cards/effects/exile_instead_of_leaving.go:49` | factored out of Whip of Erebos, shared with unearth |
| `ZoneBrowserModal.onActivateAbility` | `client/src/lib/components/board/ZoneBrowserModal.svelte:311` | the prop the browser never had |

## The three decisions worth arguing about

**`ExileSelf` is a second bool, not a zone on one "the source pays itself" component.** Discarding is a CR 701.8 keyword action with its own event (`EventDiscardCard`), its own cause (`DiscardCauseCost`) and, for cycling, `EventCycle` on top. Exiling as a cost is a plain CR 406 zone change with none of that. A card file that wanted "discard this from your graveyard" would be writing a card that does not exist. What they genuinely share — the payment IS the source, so nothing is announced, picked, or sent — they get for free by both being a bool.

**The effects read their own source back out of exile; no LKI snapshot on the stack item.** Scavenge wants "this card's power" and embalm wants the whole card, and by then their own COST has moved it. `LookupCardForEffect` finds a card in whatever zone holds it, and a card outside the battlefield has no layers applied to it (CR 613 runs on permanents) — so the printed power read out of exile at resolution is the number the graveyard held. A snapshot field would be a second LKI store beside the damage event's, which `targets.go:96-105` already declines to build for the same reason. The cost is declared: a card moved out of exile between announce and resolution does nothing (CR 608.2a, the weaker direction, #259).

**`zone_abilities` is scoped, not merely hidden — and that is the change most likely to have been got wrong.** Through #660 the rows were written to the exported `CardView` field and took their privacy from the HAND: `FilterViewFor` blanks another seat's hand wholesale. A graveyard is PUBLIC. The same code would have shipped one seat's `legal_targets` and `clauses` to the whole table, which is exactly the surface #1055 / #1167 took off the cast stamps, arriving again one field over — hexproof, shroud and "target opponent" all narrow a target set *by who is asking*. So the rows ride `castOffers`, the per-seat carrier that already exists for that question, and `publicIn` drops them. The hand's rows moved there with the graveyard's rather than leaving two lifecycles for one field.

One wrinkle, because it breaks silently: `stampLegalTargets` and `stampGrantedPermissions` file a seat's entry with `stampsFor`, which REPLACES. The ability stamp therefore MERGES and runs after both — a plain `stampsFor` here would blank a flashback card's announce surface the moment that card also printed a graveyard ability.

**The library is not walked, in either consumer.** CR 401.2 makes it hidden, no printed ability functions from one, and an enumerator that read it would touch cards the seat may not see to answer a question whose answer is always "nothing". The one line it would take is named in the comment (`legal/abilities.go:137`'s doc) so the day a card prints such an ability it is a line, not a rediscovery. Exile, the one shared pile, gets the CR 108.4 treatment in both walks: the "you" comes off `Card.Owner`, not off the loop.

## Files

**Engine**
- `server/internal/game/activated.go` — `AbilityCost.ExileSelf`; the validate and pay call sites, beside the discard component's
- `server/internal/game/exile_cost.go` (new) — the component, modelled line for line on `discard_cost.go`
- `server/internal/legal/abilities.go` — `abilityZones()`, the four-pile walk, the CR 108.4 owner filter
- `server/internal/cards/effects/registry.go:372` — the boot-time refusal of `ExileSelf` off the graveyard

**Wire**
- `server/internal/protocol/view.go` — the rename, `castStamps.ZoneAbilities`, `publicIn`'s clear, `applyTo`'s promotion, `stampZoneAbilities` over four piles, `exile_self`
- `docs/protocol.md` — `zone_abilities` replaces the `hand_abilities` bullet; the `activate_ability` row gains the exile-this paragraph

**Catalog**
- `unearth.go`, `scavenge.go`, `embalm.go`, `exile_instead_of_leaving.go` (all new); `activated.go` (`ExileThis`, `Plus`); `whip_of_erebos.go` (refactored onto the shared helper)
- `dregscape_zombie.go`, `deadbridge_goliath.go`, `sacred_cat.go` (new)

**Client**
- `protocol.ts`, `faces.ts`, `contextMenu.logic.ts`, `Card.svelte`, `Hand.svelte`, `Board.svelte` — the rename
- `ZoneBrowserModal.svelte`, `Board.svelte` — the browser's ability rows and the CR 307.1 window

**Docs**
- `docs/decisions/0020-activated-abilities.md` — a dated addendum, decisions 24–28
- `docs/decisions/0071-designations-that-switch-abilities-on.md` — a dated addendum: the zone stays out of `Designation` and out of the accessors' return value, because `Active(c Card)` reads no zone and because a zone-filtered slice would renumber the announced ability index
- `docs/engine-seams.md` — the row rewritten, a Closed-seams entry, the two follow-ups linked
- `docs/decklists/card-coverage-roadmap.md` — regenerated with `go test ./internal/cards/coverage/ -update`, never by hand

## Tests

**Engine** (`server/internal/game/other_zone_ability_test.go`): activation from a graveyard and from exile; the CR 113.6 refusal in both directions (a graveyard ability on the battlefield, with the mana untouched); the CR 108.4 non-owner refusal; `ExileSelf` exiling the source at announce with the ability still on the stack and the card still readable by instance ID with its printed power; `ExileSelf` refused outside a graveyard with nothing moved and nothing spent.

**Per keyword** (`server/internal/cards/effects/graveyard_keywords_test.go`): every keyword declares a graveyard ability, is sorcery-speed, is not offered from the battlefield or the hand, and needs no permanent; unearth returns with haste, is exiled at the next end step, and is exiled rather than binned when destroyed; scavenge exiles at announce and puts five counters off a 5/5; embalm makes a white Zombie **Cat** with no mana cost at the printed 1/1; eternalize's exception is a 4/4 black Zombie that keeps its printed types; and `addedSubtypeTypeLine` vs `retypedTypeLine` — CR 707.9b's add form against CR 707.9a's set form — is pinned directly, because picking the wrong one is silent.

**Enumerator coherence per zone** (`server/internal/legal/other_zone_ability_test.go`): unearth enumerated only when payable and the move dispatched against the live game; the zone filter in both directions; another seat's graveyard card and another seat's card in the shared exile pile are not this seat's moves; the walk covers exile and the command zone and **not** the library.

**View per viewer** (`server/internal/protocol/zone_ability_view_test.go`): the graveyard's own seat gets the row; another seat gets the card *and no row* (the card's name is asserted still readable, so the test is about scoping and not about a hidden pile); a spectator gets none; exile's rows go to the card's owner alone; and the zone filter again on the wire.

**Boot** (`registry_test.go`): `ExileSelf` without `ZoneGraveyard` panics at `Register`, and with it registers.

**Client** (`client/src/lib/zoneAbility.render.test.ts`): the pop-over opens on a browsed graveyard card, the row is labelled, the click hands the ability's own index up, and the browser closes first; nothing is offered for a card the server stamped no rows for, or when no parent wired the callback.

## Back-outs

Each fix reverted in place, the named test run, the fix restored. All nine failed as expected:

| back-out | test that caught it |
|---|---|
| `abilityZones()` returns the hand only | `TestUnearthFromGraveyardIsEnumeratedOnlyWhenPayable` |
| the enumerator stops checking CR 108.4 ownership | `TestAnotherSeatsZoneAbilitiesAreNotEnumerated` |
| the view writes the rows to the exported field again (#660's shape) | `TestGraveyardAbilitiesAreNotProjectedForOtherSeats` |
| `ExileSelf` skips its zone validation | `TestExileSelfCostIsRefusedOutsideAGraveyard` |
| `ExileSelf` is never paid | `TestExileSelfCostExilesTheSourceAtAnnounce` |
| `Register` stops refusing an exile-this cost off the graveyard | `TestRegisterExileSelfOffTheGraveyardPanics` |
| embalm uses CR 707.9a's SET form | `TestSacredCatEmbalmsIntoAWhiteZombieCat` |
| unearth drops its leaves-the-battlefield redirect | `TestUnearthedCreatureDestroyedGoesToExileNotTheGraveyard` |
| the zone browser stops wiring the activation callback | `zoneAbility.render.test.ts` |

## Out of scope, and why

- **Statics that function from a graveyard** — `StaticAbility.Zones`, the layer pass's own half of CR 113.6c (Anger, Wonder, Brawn). The other row of #1221, and the next PR. `docs/engine-seams.md`'s "Statics read from a non-battlefield zone" row is untouched.
- **Ninjutsu (CR 702.49)** — filed as **#1227**. The zone is not the problem: it is a HAND activation and #660 already reaches it. Its other two halves do not exist. There is no "return an unblocked attacker you control" cost component, and "is this attacker blocked" has no exported accessor at all (`server/internal/game/blockers.go:306` `attackerBlockedLocked` is the only reader of `Game.blockedAttackers`, `server/internal/game/game.go:346`). And nothing puts a CARD onto the battlefield attacking: `ZoneEntryOptions` (`server/internal/game/battlefield_put.go:37`) carries `Controller`, `Tapped` and `FaceDown` and no `Attacking`, and CR 506.3c's "never declared" marking happens only in the minted-token branch of `server/internal/game/entry_choice.go:334`.
- **`ManaAbilityShape` has no zone dimension** — filed as **#1228**. Found by trying to take the Spirit Guides off the seam row: "Exile this card from your hand: Add {R}" is a CR 605 mana ability, which takes the other entry point (`ActivateManaAbility`) and has neither the `Zones` field nor an exile-this cost. ADR 0071 Decision 1 note 3 says the same thing from the designation side. Both cards stay on the row, now with the reason.
- **A per-instance grant of an exile ability** (Greater Gargadon while suspended) — ADR 0062 open question 2, unchanged: `Zones` is per-DECLARATION, and a grant over one suspended card is a different object.
- **The bounce hole**, inherited whole from Whip of Erebos and now shared: `BounceToHandForEffect` moves a card without running the CR 614 pipeline, so an unearthed creature bounced in response goes to hand rather than exile (`server/internal/cards/effects/exile_instead_of_leaving.go:27-33`). That is an engine seam, not a card decision; Dregscape Zombie carries the caveat Whip already carried.
- **Cost modification for activated abilities**, unchanged from ADR 0020's #1181 addendum.

## Gate

`go build ./... && go vet ./... && go test ./...` from `server/`; `AISEAT_GAME_TESTS=1 go test ./internal/... -p 2`; `make -C server lint` (golangci-lint v1.64.8); `make -C server fmt-check`; `npm run check && npm run lint && npx vitest run` in `client/` (104 files, 1505 tests). All green, re-run after `git rebase origin/develop`.

The rebase took two conflicts, both resolved forward: `legal/abilities.go` against #1210's `restricted` parameter — the new zone walk passes it, so the board-wide "can't be activated" gate reaches a graveyard activation exactly as it reaches a battlefield one — and the census, resolved by taking develop's and re-running `-update`.


irobinson010:
PR #1374 merged (cmd_and_ctrl@2e99fd02aa03b37ac8edb2bb7534cb34e7f23b28): fix(server): hand-sourced ability cost options reach the controller alone (#1369)

Issue #1215, opened 2026-09-22
Auto-tapper refuses every sacrifice-self mana ability, so a board of Treasures reads as unpayable


irobinson010:
## What

The auto-tapper refuses every mana ability with a **sacrifice-self** cost, so a
board whose only mana is Treasures (or Lotus Petals, or Eldrazi Spawn) reports
that it cannot pay for anything.

`server/internal/game/autotap.go:577`, the first line of `autoTapAbilityFor`:

```go
for i := range abilities {
	a := abilities[i]
	if !a.TapCost || a.SacrificeCost {
		continue
	}
```

The doc comment above it gives the reason as *"a sacrifice cost needs a permanent
named (S15's original note)"*. That is true of `SacrificeOther` (Ashnod's Altar's
"Sacrifice a creature" — the activator names the victims in
`ManaAbilityParams.SacrificeIDs`), and it is **not** true of `SacrificeCost`,
which sacrifices the SOURCE and needs no decision at all: `ActivateManaAbility`
pays it with `payCostSacrificesLocked` off the source's own ID. The two were
folded into one exclusion, and the half that needs no player decision is
excluded with the half that does.

## Why it matters

This is the #273 failure shape — the engine saying "can't pay" about a board that
can — on the commonest token in the format:

- three Treasures and no untapped lands is a real Commander board state, and the
  cast preview says `ok: false` with a missing-symbols list;
- the strict-mode cast gate refuses the cast;
- the legal-move enumerator (`internal/legal/cast.go:1048`) never offers it, so
  no bot ever cracks a Treasure to cast anything.

It also makes #1212's auto-tapper ordering hint **inert for the family it was
built for**. `Spec.WantsManaFrom: game.ManaSourceTreasure` on Hired Hexblade
declares "this card reads Treasure mana, prefer a Treasure when planning", and
the planner cannot act on it because a Treasure is not a candidate at all. The
hint bites today only on sources without a sacrifice cost (a mana creature for
Inga and Esika, a snow land). Declared in ADR 0068's 2026-09-22 amendment §A5 and
in `hired_hexblade.go`'s header rather than left to be discovered.

## What it needs

1. Split the exclusion: `a.SacrificeOther != nil` stays excluded (a decision the
   planner may not make); `a.SacrificeCost` alone becomes plannable.
2. `materializePlanLocked` then has to pay the sacrifice before it taps, in the
   same order `ActivateManaAbility` does — and re-check it, the way it already
[...]

Issue #1213, opened 2026-09-22
Cost components: return a permanent to hand, sacrifice X / one or more, and discard as a mana-ability cost (CR 118.3, CR 601.2h)


irobinson010:
Three rows in `docs/engine-seams.md` have no tracker, and all three are the same kind of gap: a **cost component** the cost vocabulary cannot express. They are filed together because they are one vocabulary with three owners (`game.AbilityCost`, `effects.ManaAbilityCost`, `game.AdditionalCost`), and every one of them has to be written once and read by the payer, the validator, the protocol view's picker and the legal-move enumerator, or those four disagree (#544).

### 1. Return-a-permanent-to-hand cost component (3 cards)

> cost component: `AbilityCost` has no "return a permanent you control to its owner's hand" shape

Cards waiting: **Quirion Ranger** (#313), **Master Transmuter** (#384), **Wirewood Symbiote** (#396).

CR 118.3 / CR 602.2b. The clause is the sibling of `AbilityCost.TapOthers` (#758, `tap_others_cost.go`) one verb over: a fixed number of permanents you control, chosen at announce, matched without targeting (CR 601.2h), refused outright when the board cannot produce them.

### 2. Variable-count sacrifice cost ("sacrifice X", "one or more") (2 cards)

> cost component: #747's sacrifice clause is a fixed count, and `effects.Register` refuses a clause with `Min != Max` or `CountFromX`. A variable count needs an announced count, a record of what was paid on the stack item, and last-known information for effects that read the sacrificed permanents ("the sacrificed creature's power")

Cards waiting: **Radiant Lotus** (#462) — `{T}, Sacrifice one or more artifacts: Choose a color. Target player adds three mana of the chosen color for each artifact sacrificed this way.`; **Grim Hireling** (#300) — `{B}, Sacrifice X Treasures: Target creature gets -X/-X until end of turn.`

The two printed forms are an **open count** with a floor of one and a count **announced as X** (CR 601.2b / 602.2b), and the effect reads the number back after the permanents are gone, so it has to be a fact about the announcement rather than a re-derivation from the board.

### 3. Discard cost on a non-activated-ability surface (2 cards)

> cost component: a discard-a-card cost on a mana ability or inside an entry replacement. `AbilityCost` grew one with #660 (see Closed seams); `ManaAbilityCost` still has none, and neither does an entry replacement

Cards waiting: **Mox Diamond** (#295), **Skirge Familiar** (#389).

[...]

PR #1246, opened 2026-09-23
fix(engine): develop does not compile — the auto-tapper sacrifice call lost an argument between #1243 and #1220

irobinson010:
**`origin/develop` at `dcb7b1c1` does not compile.** Everything queued behind it is blocked until this lands.

```
$ git checkout -b probe origin/develop && cd server && go build ./...
# github.com/krakenhavoc/cmd_and_ctrl/server/internal/game
internal/game/mutations.go:1607:6: not enough arguments in call to g.validateSacrificeCostLocked
	have (uuid.UUID, uuid.UUID, AbilityCost, nil)
	want (uuid.UUID, uuid.UUID, AbilityCost, []uuid.UUID, int)
```

## Nobody wrote a bug

Two PRs edited different regions of two files. Neither conflicted textually, and each was green against the develop that existed when it was built:

| commit | PR | what it did |
|---|---|---|
| `0855b4bc` | #1243 (#1215) | added a **new** caller in the auto-tapper — "the sacrifice-self component, resolved to the concrete list BEFORE anything is paid and by the SAME validator all three cost sites share" — with the four-argument signature it was built against |
| `dcb7b1c1` | #1220 (#1213) | added the **fifth** parameter, the announced `x` for "Sacrifice X Treasures", and updated every call site that existed in ITS base |

`#1243`'s call site did not exist when `#1220` was written, and `#1220`'s parameter did not exist when `#1243` was written. Nothing built the combination until both had merged.

This is exactly the failure class [`ci-cd.yml`'s own header](../blob/develop/.github/workflows/ci-cd.yml) documents from 2026-09-11 — *"two PRs each added a GoldToken to different regions of tokens.go, so neither conflicted textually and both were green against the main that existed before the other landed"* — and the reason the `merge_group` trigger exists. Worth a look at why the queue did not catch this one.

## The fix, and why the argument is 0 rather than a guess

```go
sacrifices, serr := g.validateSacrificeCostLocked(p.ID, cardID, AbilityCost{
    SacrificeSelf: ab.SacrificeCost,
}, nil, 0)
```

Three independent reasons it is 0, any one of which would be enough:

- The parameter is read **only** by a clause whose count comes FROM X (`SacrificeCountFromX`). A fixed clause ignores it.
- The clause at this site is `nil` **by construction**, which the comment immediately above the call already says: *"SacrificeOther is nil by construction — autoTapAbilityFor refuses a source that carries one — which is why no SacrificeIDs are named and none can be demanded."*
- This is a **mana ability**, and a mana ability announces no X at all.

And `validateSacrificeCostLocked`'s own doc, written by #1213, spells out this exact case: *"A FIXED clause reads exactly as it did and ignores `x` entirely; **pass 0 from a site that has no X to announce**."*

So the change restores precisely what #1243 meant and alters no behaviour. A comment at the call site records the reasoning, so the next person to touch the signature does not have to re-derive it.

## Verification

On a clean checkout of `origin/develop`, before the change: `go build ./...` fails. After: `go build ./...`, `go vet ./...`, `go test ./...`, `make -C server fmt-check` and `make -C server lint` (golangci-lint v1.64.8) all pass.

No test is added. The failure is a compile error, which the build step already catches the moment the two changes coexist — a test asserting that a call site has the right arity is a test that the compiler is working.

## Provenance

Found while rebasing #1232 (the activated half of #1221) off the #1220 stack and onto develop. Raising it as its own PR rather than folding it into a feature branch, because it blocks every queued PR and should merge ahead of all of them.


irobinson010:
Landed inside #1232 as commit fef8f0f7

